PR #1, opened 2021-05-27
Fix issues when compiling against ROOT6

plexoos:
Unlike rootcint, rootcling in ROOT6 seems to be mindful of the access modifiers for nested types used to declare data members. Such types should have public access


StRoot/StTrsMaker/src/StTpcDbElectronics.cc
```diff
@@ -89,9 +89,9 @@ StTpcDbElectronics::instance()
 #ifndef ST_NO_EXCEPTIONS
 	throw invalid_argument("StTpcDbElectronics::getInstance(): Argument Missing!");
 #else
-	cerr << "StTpcDbElectronics::getInstance(): Argument Missing!" << endl;
-	cerr << "No arguments for instantiantion" << endl;
-	cerr << "Exiting..." << endl;
+	std::cerr << "StTpcDbElectronics::getInstance(): Argument Missing!" << endl;
```
veprbl:
```suggestion
	std::cerr << "StTpcDbElectronics::getInstance(): Argument Missing!" << std::endl;
```

PR #8, opened 2021-06-02
Miscellaneous fixes to cons scripts aiming future builds with ROOT6 and Spack environment

plexoos:


klendathu2k (approved):
These changes look to be mostly harmless.

genevb (approved):
I don't see anything in particular that would be problematic, but I'm unfamiliar with most of this.

PR #12, opened 2021-06-03
pr request test: adding dummy IDL file (compilable)

dmarkh:
Testing PRs

### DIFF
--- a/StDb/idl/dummy.idl
+++ b/StDb/idl/dummy.idl
@@ -1,0 +1,10 @@
+/*    dummy.idl
+*
+*  Table: dummy
+*
+*       description: //  dummy file
+*/
+
+struct dummy {.
+  unsigned long dummy; /* dummy parameter  */
+};


plexoos:
Hm... I wonder why github did not automatically add @plexoos or @nigmatkulov as reviewers for this PR.

@veprbl Do you know why?

plexoos:
@dmarkh  Should we add another maintainer for  StDb/  in the CODEOWNERS file? Maybe Gene or Yuri?

dmarkh:
> @dmarkh Should we add another maintainer for StDb/ in the CODEOWNERS file? Maybe Gene or Yuri?

Well, this section contains IDL files = data descriptors. I also create another copy of the same descriptor directly in MySQL when I commit an IDL file. While Yuri and Gene may infrequently commit or update an IDL file that does not have a corresponding MySQL descriptor (for development purposes), they can't technically verify my commits as they do not see the MySQL part of the new data structure. Moreover, since IDL files are not a "code" but rather data descriptors, I'm not sure what needs to be approved here. Finally, since I have already created MySQL descriptor before I commit an IDL file, this review process just for an IDL creates an "out of sync" situation until the review is finished and DEV library is rebuilt with the IDL included.

To summarize: Is it possible to drop reviews for the /StDb/ section? The same trick could be needed for XML geometries Jason is committing.

plexoos:
> Is it possible to drop reviews for the /StDb/ section? The same trick could be needed for XML geometries Jason is committing.

At this moment, I don't see how it can be controlled at the subdirectory level with the features provided by GitHub. Of course, all of this administration can be customized and arbitrary logic can be implemented in  a script using Github API. Unfortunately, I don't have one available now. If you find a solution let me know.

StDb/idl/dummy.idl
```diff
@@ -0,0 +1,10 @@
+/*    dummy.idl
+*
+*  Table: dummy
+*
+*       description: //  dummy file
+*/
+
+struct dummy {.
```
plexoos, line 8: Is there an extra dot at the end of line?

dmarkh:
The following Action could be useful:
https://github.com/marketplace/actions/code-owner-self-merge

-Dmitry

StDb/idl/dummy.idl
dmarkh: That dot does not matter, it is not a C++ file. But, it clearly demonstrates why reviews can delay the merge, right? As I mentioned in the title, submitted IDL file is compilable by cons and it produces valid .h and .inc files. Create an "StDb/idl" directory at rcas, place this file inside, run cons, see it producing a library. This should be done by CI script, really.

plexoos: Yes, a PR with CI scripts is very welcome. Let me know if you want to work on it.

dmarkh:
> Hm... I wonder why github did not automatically add @plexoos or @nigmatkulov as reviewers for this PR.
> 
> @veprbl Do you know why?

Because I started with "draft" PR, then updated it to "ready for review". I guess, the hook Codeowners processing is only invoked upon the initial "create" state and does not process the "edit" states.

Personally, I find the "codeowners" title confusing. It is not about code ownership at all, it is about notifications: "if a file matches a pattern, then notify user(s) X". 

Also, is it possible to create "teams" in the free (presumably) GitHub version we are using now? Codeowners file allows team names instead of personal names. Would be great to have "database team" (me and Victor - historically St_db_Maker is his code), "simulations team", "tracking team", "QA team", etc. I read that GitHub Teams allows special handling for code reviews.

StDb/idl/dummy.idl
dmarkh: Well, SPHENIX has CI governed by Jenkins: https://web.sdcc.bnl.gov/jenkins-sphenix/
STAR Infrastructure team may consult Chris and SDCC folks about CI setup details. I'd love to do that myself (CI is great), but I have plenty of tasks on my plate already, and this list will grow rapidly when RHIC Run 21 will be over.

plexoos: If you don't have time like most of us then we'll just use what we have.
Now, you see how useful this PR was. Now I know that idl files are not in C++ and taht having a dot at the end of line is OK. You shared the knowledge. Thank you

veprbl:
> @veprbl Do you know why?

You are not a codeowner for /StDb:
https://github.com/star-bnl/star-sw/blob/9709bd2797ec3ebc4121be85e2eb299667d046fb/.github/CODEOWNERS#L3
only @dmarkh is, but since he is also an author of PR there were no reviews requested.
If you want to be a reviewer, on all requests you need to actually put yourself on every line of CODEOWNERS.

PR #23, opened 2021-06-09
consolidating chain options for pppAMode

BassamAboona:
I made some edits to reduce the number of chain options required to turn on pppAMode. I also added a log message to indicate that pppAMode is on. For some reason, the spacing in StVpdCalibMaker.cxx looks funny. This is not the case with the local copy that I committed and pushed.


fgeurts (dismissed):

fgeurts (changes requested):
I noticed that your PL includes adding two "swp" files. Please remove those. The proposed changes to the cxx files are fine.

BassamAboona:
> I noticed that your PL includes adding two "swp" files. Please remove those. The proposed changes to the cxx files are fine.

Sorry about that! They should both be deleted now.

fgeurts (approved):

plexoos (approved):

PR #22, opened 2021-06-09
Add CODEOWNERS entries for genevb

genevb:

PR #24, opened 2021-06-09
 Introducing setPPPAOutlierRej and pppAMode options in StBFChain

fgeurts:
These are changes on behalf of Bassam Aboona (TAMU) that will enable modes that are specific to application of pp/pA related TOF/VPD calibrations.

I note that my editor (atom) appears to have spuriously changed that do not have any impact.


plexoos:
We need to merge #22 first so @genevb could review the changes in StRoot/StBFChain

Note that the actual changes can be viewed under the "Files changed" tab. The changes in white space can be ignored by adding '?w=1' to the address, i.e.

    https://github.com/star-bnl/star-sw/pull/24/files?w=1

genevb (approved):

plexoos (approved):

genevb:
> These are changes on behalf of Bassam Aboona (TAMU) that will enable modes that are specific to application of pp/pA related TOF/VPD calibrations.

In hindsight, I have two thoughts...

(1) The name of the chain option doesn't really specify that it is for TOF/VPD. As we haven't used this in production, but we are about to do so, could we change it from pppAMode to either TOFpppAMode or pppAModeTof? Similarly, setPPPAOutlierRej could be TOFsetPPPAOutlierRej or setPPPAOutlierRejTof? It isn't critical to change it, but if there are no objections, it would be straightforward to do so now before its official use.

(2) I heard yesterday that it is of interest to use this in the low energy AA collisions of BES-II. It's possible someone told me this before, but I didn't remember that. Again, the name of the chain option implies something different, that its intended use is in pp and pA collisions. I won't propose any changes related to this point, but a more relevant name for the option may have helped us all understand when to include the chain option and when not to include it.

Thanks,
-Gene

PR #25, opened 2021-06-09
upgrade TPCCATracker and related changes

kehw:
Files changed are mostly in 
StRoot/StiCA   
StRoot/TPCCATracker  
StRoot/TPCCATrackerPerformance

other files changed are
StRoot/Sti/StiHit.cxx 
StRoot/Sti/StiKalmanTrackFinder.cxx
StRoot/StiMaker/StiStEventFiller.cxx

and 

mgr/ConsDefs.pm
mgr/Conscript-standard

@fisyak

### DIFF
--- a/StRoot/Sti/StiHit.cxx
+++ b/StRoot/Sti/StiHit.cxx
@@ -240,7 +240,6 @@ static int nCall =0; nCall++;
  void StiHit::addTimesUsed()
 {
     mTimesUsed++;
-    // assert(mTimesUsed <=mMaxTimes);
 }
 //_____________________________________________________________________________
  void StiHit::subTimesUsed()
--- a/StRoot/Sti/StiHit.cxx
+++ b/StRoot/Sti/StiHit.cxx
@@ -240,7 +240,7 @@ static int nCall =0; nCall++;
  void StiHit::addTimesUsed()
 {
     mTimesUsed++;
-    assert(mTimesUsed <=mMaxTimes);
+    // assert(mTimesUsed <=mMaxTimes);
 }
 //_____________________________________________________________________________
  void StiHit::subTimesUsed()
--- a/StRoot/StiCA/StiCATpcSeedFinder.cxx
+++ b/StRoot/StiCA/StiCATpcSeedFinder.cxx
@@ -19,8 +19,7 @@
 //#define KINK_REJECTION  
 #define OVERLAP_REJECTION
 //#define StiCATpcSeedFinderBLOG
-#include <unordered_map>
-#include <set>
+
 //________________________________________________________________________________
 Bool_t StiCATpcSeedFinder::SeedsCompareStatus(const Seed_t a, const Seed_t b)
 {
@@ -36,7 +35,6 @@ StiTrack *StiCATpcSeedFinder::findTrack(double rMin)
     // zero all banks before filling !!! 
     auto *map =  &StiToolkit::instance()->getHitContainer()->hits();
 
-
     // Run reconstruction by the CA Tracker
     caTrackerInt.SetHits(*map);
     caTrackerInt.Run();
@@ -49,24 +47,11 @@ StiTrack *StiCATpcSeedFinder::findTrack(double rMin)
     Seed_t aSeed = mSeeds->back();
     mSeeds->pop_back();
 
-    // std::unordered_map<const StiHit*, int> hitCount;
-    // for (auto h : aSeed.vhit) {
-    //   hitCount[h->hit]++;
-    // }
-    // for (auto h : hitCount) {
-    //   if (h.second > 1) {
-    //     cout << "seed: " << mSeeds->size() << " : duplicated hits: " << h.first << " : " << h.second << endl;
-    //   }
-    // }
-
     vector<StiHit*>        _seedHits;
-    // std::set<StiHit*> existingHits;
     int nHits = aSeed.vhit.size();
     Int_t nHitsUsed = 0;
     for (int iHit=0;iHit<nHits;iHit++)     {
       StiHit *hit = aSeed.vhit[iHit]->hit;if (!hit) continue;
-      // if (existingHits.count(hit)) continue;
-      // existingHits.insert(hit);
       if (hit->timesUsed()) nHitsUsed++;
       _seedHits.push_back(hit);
     }
--- a/StRoot/StiCA/StiCATpcSeedFinder.cxx
+++ b/StRoot/StiCA/StiCATpcSeedFinder.cxx
@@ -19,6 +19,8 @@
 //#define KINK_REJECTION  
 #define OVERLAP_REJECTION
 //#define StiCATpcSeedFinderBLOG
+#include <unordered_map>
+#include <set>
 //________________________________________________________________________________
 Bool_t StiCATpcSeedFinder::SeedsCompareStatus(const Seed_t a, const Seed_t b)
 {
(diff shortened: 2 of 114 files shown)


plexoos:
Ouch... Files changed 110  +20,472 −5,463

Can this be split into smaller commits with relevant log messages by any chance? A good start would be to separate the new files from the changes to the existing ones.

kehw:
There are three commits. The first two are small. The 3rd one is indeed large. It contents changes in the following two files

mgr/ConsDefs.pm
mgr/Conscript-standard

and changes in

StRoot/StiCA
StRoot/TPCCATracker
StRoot/TPCCATrackerPerformance (new directory)

I did not break the 3rd commit into smaller pieces, because they won't compile if they are not together. 
If there that is not a concern, I can break it.

mgr/ConsDefs.pm
```diff
@@ -1179,13 +1179,26 @@
...
-   if (-e "temp_gccflags.c") {`rm temp_gccflags.c`;}
- }
+
+    $TARGET = "";
+    if ($USE_64BITS){
+	$TARGET = "x86_64";
+    } else {
+	$TARGET = "x86";
+    }
+    my $VcDIR = `/cvmfs/star.sdcc.bnl.gov/star-spack/spack/bin/spack find --format '{prefix}' 'vc\@1.4.1' target=$TARGET`;
```
plexoos: This is a very fragile approach for passing dependencies to the build system. Cons should not know or assume anything about Spack. Hard coding the path to Spack is not going to help it either and that's why the CI fails.

A more robust way is to set everything cons needs outside of cons. See, for example, how I did it by using environment variables (a common approach) for the CI build:

https://github.com/star-bnl/star-sw/blob/6a0d42373a26c1ce21b23dc256c8938bf135cf5a/docker/Dockerfile.root5#L92-L93

kehw:
Hi @plexoo , the CI check failed because cons cannot find the Vc in star-spack. I use the absolute cvmfs path of the spack command, which is not available in the docker container.

mgr/ConsDefs.pm
kehw: Yes, I understand that. Your example assumes spack can be foudn in $PATH. I know that is true for CI build. Can I assume the production team also have spack in $PATH?

plexoos: To add spack to PATH one can `source star-spack/setup.sh`. `star-spack` is also installed in `/cvmfs/star.sdcc.bnl.gov`

plexoos:
Since this PR was submitted we had quite a few improvements in CI. You'd better merge 'main' or even better rebase.

StRoot/Sti/StiHit.cxx
```diff
@@ -240,7 +240,7 @@ static int nCall =0; nCall++;
  void StiHit::addTimesUsed()
 {
     mTimesUsed++;
-    assert(mTimesUsed <=mMaxTimes);
+    // assert(mTimesUsed <=mMaxTimes);
```
plexoos: Please do not comment code. If you remove the code, just remove it. It is better to explain in the log message what motivates this change. The commented line like this one is indicative that you are hesitant about the proposed change and there is a possibility that it will be reverted (i.e. "uncommented"). Well, if you are not sure that the hit usage should be constrained and it still needs to be studied or confirmed then introduce a run time switch.

mgr/Conscript-standard
```diff
@@ -908,26 +908,46 @@ if ( $pkg !~ /^sim$/ && $pkg !~ /^gen$/ ) {
...
+#	if ($STAR_SYS =~ 'gcc432$') {$CXXFLAGS .= " -DVC_IMPL=Scalar";}
+#	else                        {$CXXFLAGS .= " -DVC_IMPL=SSE";}
+	my @srcC = ();
+	foreach my $s (@src) {
+	    #      print "$s\n";
+	    #      next if $s eq 'benchmarks/benchmark.cpp';
+	    my $dirL = File::Basename::dirname($s);
+	    my $sdir = File::Basename::basename($dirL);
+	    next if $sdir eq 'examples' or $sdir eq 'benchmarks';
+	    push @srcC, $s;
```
plexoos: Why are you reintroducing the build for a local Vc package? This setup for Cons was remove in 5a87e821cb. We don't have StRoot/Vc and don't plan to have it in our repository as it is an external package.

mgr/Conscript-standard
```diff
@@ -908,26 +908,46 @@ if ( $pkg !~ /^sim$/ && $pkg !~ /^gen$/ ) {
...
-	$extraCXXFLAGS .= " -pedantic -Wno-long-long -Wno-variadic-macros -msse -msse2 -Wshadow -fno-threadsafe-statics -DTBB_DO_ASSERT=1";
-	if ($STAR_SYS eq 'sl64_gcc447' or $STAR_SYS eq 'sl64_x8664_gcc447') {$extraCXXFLAGS .= " --param large-function-growth=500";}
+#    } elsif ( $pkg =~ m/^TBB$/) {
+#	$CPPPATH = $INCLUDE
+#	    . $main::PATH_SEPARATOR . "#" . $Dir . "/src";
+#	#      "#" . $Dir . $main::PATH_SEPARATOR . "#" . $Dir . "/include" .
+#	#    print "CPPPATH => $CPPPATH\n";
+#	$CXXFLAGS .= " -MMD -DTBB_DO_ASSERT -DDO_ITT_NOTIFY -DUSE_PTHREAD -D__TBB_BUILD=1 -Wall -Wno-parentheses";
+#
+#
```
plexoos: Why are you adding these commented lines?

mgr/Conscript-standard
```diff
@@ -908,26 +908,46 @@ if ( $pkg !~ /^sim$/ && $pkg !~ /^gen$/ ) {
...
+	    my $dirL = File::Basename::dirname($s);
+	    my $sdir = File::Basename::basename($dirL);
+	    next if $sdir eq 'examples' or $sdir eq 'benchmarks';
+	    push @srcC, $s;
+	}
+	@src = @srcC;
 
-        if ( defined($ENV{Vc_DIR}) ) {
-          $CPPPATH = $ENV{Vc_DIR}."/include".$main::PATH_SEPARATOR.$CPPPATH;
-          $LIBPATH = $ENV{Vc_DIR}."/lib";
```
plexoos: Why are you proposing to remove these lines setting the build options for TPCCATracker and StiCA based on the `Vc_DIR` environment variable? This is related to the comment above about the hard coded `spack` command which we would like to avoid.

StRoot/StiCA/StiCATpcSeedFinder.cxx
```diff
@@ -42,153 +44,39 @@ StiTrack *StiCATpcSeedFinder::findTrack(double rMin)
...
+    // std::unordered_map<const StiHit*, int> hitCount;
+    // for (auto h : aSeed.vhit) {
+    //   hitCount[h->hit]++;
+    // }
+    // for (auto h : hitCount) {
+    //   if (h.second > 1) {
+    //     cout << "seed: " << mSeeds->size() << " : duplicated hits: " << h.first << " : " << h.second << endl;
+    //   }
+    // }
+
```
plexoos: Please remove commented lines.

StRoot/StiCA/StiCATpcSeedFinder.cxx
```diff
@@ -19,6 +19,8 @@
 //#define KINK_REJECTION  
 #define OVERLAP_REJECTION
 //#define StiCATpcSeedFinderBLOG
+#include <unordered_map>
+#include <set>
```
plexoos: These classes are not used in the code

StRoot/StiCA/StiCATpcTrackerInterface.cxx
```diff
@@ -11,387 +11,26 @@
...
-
-  // for sti perfo
-#ifdef DO_TPCCATRACKER_EFF_PERFORMANCE
-#include "TPCCATrackerPerformance/AliHLTTPCCAStiPerformance.h"
-#include "TPCCATrackerPerformance/AliHLTTPCCAMergerPerformance.h"
-#include "StDetectorDbMaker/St_tpcPadConfigC.h"
-#include "Sti/StiKalmanTrack.h"
-#include "Sti/StiKalmanTrackNode.h"
-#endif /* DO_TPCCATRACKER_EFF_PERFORMANCE */
+#include "TStopwatch.h"
```
plexoos: TStopwatch is not used in this file

StRoot/StiCA/StiCATpcTrackerInterface.cxx
```diff
@@ -404,12 +43,12 @@ void StiCATpcTrackerInterface::MakeHits()
 
         // get local coordinates. take into account distortion
       StiHit *hit = *cit;
-      if ( !(hit->detector()->isActive()) ) continue;
       if (! hit->stHit()) 	continue;
-      if (  hit->timesUsed()) 	continue;//VP
+      //yf      if (  hit->timesUsed()) 	continue;//VP
```
plexoos: Do not add the commented line

StRoot/StiCA/StiCATpcTrackerInterface.cxx
```diff
@@ -443,15 +82,16 @@ void StiCATpcTrackerInterface::MakeHits()
       caHit.SetErrZ( 0.16 );
       caHit.SetISlice( tpcHit->sector() - 1 );
       caHit.SetIRow( hitc.padrow );
-      caHit.SetID( fCaHits.size() );
+      //      caHit.SetID( fCaHits.size() );
```
plexoos: Do not add the commented like

StRoot/TPCCATracker/StTPCCAInterface.cxx
```diff
@@ -0,0 +1,397 @@
...
+// need for hits data
+#include "StMaker.h"
+#include "StTpcHit.h"                
+#include "StTpcDb/StTpcDb.h"
+#include "StDbUtilities/StTpcCoordinateTransform.hh"
+#include "StDbUtilities/StTpcLocalSectorCoordinate.hh"
+  // for MCdata
+#include "tables/St_g2t_track_Table.h" 
+#include "tables/St_g2t_tpc_hit_Table.h"
+#include "TDatabasePDG.h"
```
plexoos, line 15: Not used?

docker/Dockerfile.root6
```diff
@@ -93,6 +93,7 @@ ENV LIBXML2_DIR=/opt/software/linux-scientific7-x86_64/gcc-4.8.5/libxml2-2.9.10-
 ENV Vc_DIR=/opt/software/linux-scientific7-x86_64/gcc-4.8.5/vc-1.4.1-xdb5jhivtn2rejvvfs2whrf4nx4lxwbi
 ENV FASTJET_DIR=/opt/software/linux-scientific7-x86_64/gcc-4.8.5/fastjet-3.3.3-xj22tput5pd5atejy4igsrtgdtygi74u
 ENV GSL_DIR=/opt/software/linux-scientific7-x86_64/gcc-4.8.5/gsl-2.6-ao26bt7pm6xxywo2q7dcmhhg2lp2inx7
+ENV VC_DIR=/opt/software/linux-scientific7-x86_64/gcc-4.8.5/vc-1.4.1-xdb5jhivtn2rejvvfs2whrf4nx4lxwbi
```
plexoos: `Vc_DIR` is defined three lines above this line.

StRoot/TPCCATracker/StTPCCAInterface.cxx
kehw: Can we keep them for now?

mgr/Conscript-standard
kehw: I overlooked this part when I rebase my changes.

mgr/Conscript-standard
kehw: I overlooked this part when I rebase my changes.

docker/Dockerfile.root6
kehw: I did not notice you defined a Vc_DIR in dockerfile.root6, but it is not defined in dockerfile.root6.
I defined VC_DIR in both files and use all capital letters to maintain to other variables you defined.

mgr/Conscript-standard
kehw: Insead of reading environment variable, I defined a "package" defined
https://github.com/kehw/star-sw/blob/main/mgr/ConsDefs.pm#L1285

and use the package as
```perl
$CXXFLAGS .= $env->{Packages}->{Vc}->{CPP};
$LIBS .= $env->{Packages}->{Vc}->{LIBS};
```

docker/Dockerfile.root6
kehw: I guess the main branch is still using the Vc 0.7.4 from ROOT5 and you redefined the Vc_DIR in dockerfile.root6 to just pick up Vc 0.7.4 in spack. Removing Vc 0.7.4 from ROOT5 and just using spack + VC_DIR would give us a unified usages of Vc.

plexoos: `Vc_DIR` is defined by the spack package:

https://github.com/star-bnl/star-spack/blob/fba3a3ffbafc0c77fa46085d807b717ca807f345/packages/vc/package.py#L31-L34

That is the reason why it does not have to be set explicitly in Dockerfile.root5
The ENV definitons you are proposing to change are not relevant for the build stage because these values get overwritten by loading the environment modules before the actual build command. Also in case of ROOT6 the default value defined by Vc module is currently being overwritten manually:

https://github.com/star-bnl/star-sw/blob/948846a8dc8b9a891f42397889c2dbdfa1c644ec/docker/Dockerfile.root6#L124-L126

What you actually want to do in the context of this StiCA upgrade is to remove the line manually overwriting Vc_DIR in Dockerfile.root6 and instead add a similar one in Dockerfile.root5.

If you really want to rename the environment variable from `Vc_DIR` to `VC_DIR` then you better start with the definition in the spack package. I don't care about such nitpicking changes and would accept them. Although, this would require a rebuild of the base image for our CI tests.

mgr/Conscript-standard
plexoos: The VC package you define in ConsDefs.pm is not "instead of" the defined environment variable. It actually **uses** the same environment variable to define the package. By doing this you just introduce an extra layer between the CPPPATH and LIBPATH variables and the value set in the environment variable. While there is perhaps a deeper concept behind the Cons environments it is not clear if there is a real benefit from these extra lines of code...

mgr/ConsDefs.pm
```diff
@@ -1182,13 +1182,18 @@
...
+    my $VcLIB = $VC_DIR . "/lib/libVc.a";
+    #Vc check SSE support
+    # my $cmd = "touch temp_gccflags.c; $CXX -E -dM -x c++ -o - temp_gccflags.c | grep -q SSE";
+    my $cmd = "$CXX -E -dM -x c++ - < /dev/null | grep -q SSE";
+    if ($STAR_HOST_SYS =~ 'gcc432$' || system($cmd)) {# No SSE
+        $VcCPPFLAGS .= " -DVC_IMPL=Scalar";
+        # if (-e "temp_gccflags.c") {`rm temp_gccflags.c`;}
+    } else {
+        $VcCPPFLAGS .= " -DVc_IMPL=SSE";
+    }
```
plexoos: You are adding a definition `-DVC_IMPL=...` that is not used in your code.

mgr/ConsDefs.pm
```diff
@@ -1182,13 +1182,18 @@
...
- }
+
+    my $VcCPPFLAGS = " -I" . $VC_DIR . "/include" . " -Wabi -fabi-version=0";
+    my $VcLIB = $VC_DIR . "/lib/libVc.a";
+    #Vc check SSE support
+    # my $cmd = "touch temp_gccflags.c; $CXX -E -dM -x c++ -o - temp_gccflags.c | grep -q SSE";
+    my $cmd = "$CXX -E -dM -x c++ - < /dev/null | grep -q SSE";
+    if ($STAR_HOST_SYS =~ 'gcc432$' || system($cmd)) {# No SSE
+        $VcCPPFLAGS .= " -DVC_IMPL=Scalar";
+        # if (-e "temp_gccflags.c") {`rm temp_gccflags.c`;}
```
plexoos: Again, why add commented code? It is confusing

mgr/ConsDefs.pm
```diff
@@ -1182,13 +1182,18 @@
...
- my $VcCPPFLAGS = " -DVC_IMPL=SSE";
- if ($STAR_HOST_SYS =~ 'gcc432$' || system($cmd)) {# No SSE
-   $VcCPPFLAGS = " -DVC_IMPL=Scalar";
-   if (-e "temp_gccflags.c") {`rm temp_gccflags.c`;}
- }
+
+    my $VcCPPFLAGS = " -I" . $VC_DIR . "/include" . " -Wabi -fabi-version=0";
+    my $VcLIB = $VC_DIR . "/lib/libVc.a";
+    #Vc check SSE support
+    # my $cmd = "touch temp_gccflags.c; $CXX -E -dM -x c++ -o - temp_gccflags.c | grep -q SSE";
```
plexoos: Commented out code is just noise

mgr/ConsDefs.pm
```diff
@@ -1182,13 +1182,18 @@
...
- #Vc check SSE support
- my $cmd = "touch temp_gccflags.c; $CXX -E -dM -o - temp_gccflags.c | grep -q SSE";
- my $VcCPPFLAGS = " -DVC_IMPL=SSE";
- if ($STAR_HOST_SYS =~ 'gcc432$' || system($cmd)) {# No SSE
-   $VcCPPFLAGS = " -DVC_IMPL=Scalar";
-   if (-e "temp_gccflags.c") {`rm temp_gccflags.c`;}
- }
+
+    my $VcCPPFLAGS = " -I" . $VC_DIR . "/include" . " -Wabi -fabi-version=0";
+    my $VcLIB = $VC_DIR . "/lib/libVc.a";
```
plexoos: Why don't you define $LIBPATH for Vc? It is defined for other packages.
By hardcoding the library name as `libVc.a` you make the build configuration less portable.

mgr/Conscript-standard
```diff
@@ -908,26 +908,18 @@ if ( $pkg !~ /^sim$/ && $pkg !~ /^gen$/ ) {
...
-	$extraCXXFLAGS .= " -DNODEBUG";
-	$CXXFLAGS .= $extraCXXFLAGS;
-	$CXXFLAGS .= $env->{Packages}->{Vc}->{CPP};
-	if ( defined($ENV{Vc_LIB_DIR}) ) { die("Vc_LIB_DIR is not supported anymore. Please use Vc_DIR."); }
-	$LIBS = "-lVc";
-	if ($extraCXXFLAGS =~ /USE_TBB/) {$LIBS .= " -lTBB";}
+            $main::PATH_SEPARATOR . "#" . $Dir . "/../TPCCATrackerPerformance"
+            . $main::PATH_SEPARATOR . "#" . $Dir . "/../TPCCATracker";
+        $extraCXXFLAGS = " -DNODEBUG -DNVALGRIND ";
+        $extraCXXFLAGS .= " -Wno-parentheses -Wno-unused-function -Wno-deprecated-declarations -Wno-unused-but-set-variable -Wno-unused-variable";
```
plexoos, line 951: Why are you disabling warnings for unused but set variables? Wouldn't it be the right thing to not declare variables if they are unused?

StRoot/TPCCATracker/StTPCCAInterface.cxx
plexoos: No 😄  Unused includes are not free, they do slow down the build. Ideally, someone should run a utility to identify unused headers.

mgr/Conscript-standard
kehw: I am no cons expert. Read the file, usage like "$env->{Packages}" seems a consistent way to use third party packages. The packages was defined in ConsDefs.pm, but not used. I am trying to keep it consistent to rest of mgr/Conscript-standard.

docker/Dockerfile.root6
kehw: I do not know "Vc_DIR" is define by spack and export to the CI environment. Is this also true for other libraries, like  "GSL_DIR"? It would be better if such information is available more explicitly. I do not have string preferences between "Vc_DIR" and "VC_DIR", but I do prefer consistency. I assume redefine Vc_DIR exlicityly in Dokerfile.root5 would not be problem, right?

mgr/ConsDefs.pm
kehw: It is not for STAR code, it is used by Vc. This macro is used to control which instruction set should Vc use. Depend on CPU and compiler, If SSE2 and above are available, use SSE, otherwise  fallback to scalar version.

mgr/ConsDefs.pm
kehw: It can be defined. We both know the Vc only has this static library, it may not add much flexibility, but yes, it can be done.

mgr/Conscript-standard
kehw: That is for TPCCATracker. They keep some variables there for the convenience for tuning. To clear up, we need some help from the GSI group.

docker/Dockerfile.root6
plexoos: > Is this also true for other libraries, like "GSL_DIR"?

It is up to the package to set such variables. You'd need to check the Spack package file to know for sure.

> I assume redefine Vc_DIR exlicityly in Dokerfile.root5 would not be problem, right?

Again, it depends on where you set it... It might be slightly better to avoid setting anything by hand if we don't have to. For example, if the hash changes then you need to remember to update it, etc.

kehw: In this case,  I prefer to define every XX_DIR in both Dockerfiles explicitly. To check and remember every such variables in spack is not attractive to me.

mgr/ConsDefs.pm
plexoos: Isn't it set automatically if not defined?

kehw: If not defined, Vc will automatically choose the highest available instruction set. For example, if AVX is available, it will choose AVX. But we only want SSE.

mgr/Conscript-standard
plexoos: Since you like consistency... let's try to keep the same compiler flags as for all other packages unless we really have to add or modify them. The reasons are:

- Minimize the number of changes
- Less options to worry about in case we switch to a new version/compiler

mgr/ConsDefs.pm
plexoos: Well, in that case what you do here may be inconsistent with how Vc is built by Spack. I think this macro is set to 'auto' by default and that is how Vc lib is built

kehw: When you build Vc, it builds once for each supported instruction set and archive multiple version of binaries to a signal .a file. When one uses it, we can use the macro Vc_IMPL to pick the desired version.

mgr/Conscript-standard
kehw: It is not inconsistency here. These flags are recommended by the CA tracker developers. Once we have new version/compiler, they should review these flags.

docker/Dockerfile.root6
kehw: I do know how does the dockerfile interactive with spack, but redefine 

ENV Vc_DIR=/opt/software/linux-scientific7-x86_64/gcc-4.8.5/vc_-1.4.1-waaa6u2x4ofz4sv5hkrt5576vuioxl4q

in dockerfile.root5 does not work. It still points to 0.7.4.

kehw: I have tried what you suggested
```
What you actually want to do in the context of this StiCA upgrade is to remove the line manually overwriting Vc_DIR in Dockerfile.root6 and instead add a similar one in Dockerfile.root5.
```
But root5 CI jobs still failed. Can you take a look? @plexoos

plexoos: Try patching it like this

```diff
diff --git a/docker/Dockerfile.root5 b/docker/Dockerfile.root5
index 1d3735f570..99b5136a03 100644
--- a/docker/Dockerfile.root5
+++ b/docker/Dockerfile.root5
@@ -121,6 +121,8 @@ diff --git a/asps/rexe/Conscript b/asps/rexe/Conscript \n\
  && patch -p1 < root4star_cons.patch

 RUN source /etc/profile \
+ # Override default Vc version
+ && export Vc_DIR=/opt/software/linux-scientific7-x86_64/gcc-4.8.5/vc_-1.4.1-waaa6u2x4ofz4sv5hkrt5576vuioxl4q/ \
  && cd /star-sw \
  && cons +StarVMC/Geometry \
  && cons %OnlTools \

```

iraklic:
Hi @kehw there are some trivial changes/commits that I can easily approve, but some like hit reuse are not only TPCCA related and I think Yuri and Victor have to look at it. I actually remember that change many years ago and I remember Victor needing that assert. Victor can correct me if I remember right and if so I can go ahead and approve that too. Does partial approve do you any good for this PR?

kehw:
Hi @iraklic , the mTimesUsed is only used to prevent hit reuse. I discussed with Victor some time ago to remove this constrain. We discussed last Wednesday, that TPCCATracker and  TPCCATrackerPerformance do not need to be reviewed line by line. They will be evaluated by physics output. Therefore, it will be good if you can review the rest of the changes. Thanks!

iraklic (approved):
@kehw reports this being discussed and agreed with Victor, so I do not have any objections. I actually remember doing reconstruction with this change already in.

iraklic (approved):

iraklic (approved):

iraklic:
Wait, do I have to approve every commit one by one or is this PR approved all together now? My intention was to approve all of these, but I am not sure if that is what happened. @plexoos can you tell me if my approval went through or I need to do every one of these commits separately? thanks

plexoos:
>  or I need to do every one of these commits separately?

It never worked this way. Approval is for the entire PR

iraklic:
> > or I need to do every one of these commits separately?
> 
> It never worked this way. Approval is for the entire PR

That's what I thought but this PR is little strange so I thought I'd doublecheck. thanks.

PR #27, opened 2021-06-11
Move pams/gen/idl from star-mcgen repo to star-sw

plexoos:
This should allow to build star-sw independent of star-mcgen

Merge it with the corresponding change in https://github.com/star-bnl/star-mcgen/pull/1


klendathu2k (approved):
Agreed.  This directory should be in the main repo, rather than the generator repo.

veprbl (approved):
Looks innocent

PR #31, opened 2021-06-14
Update of the centrality definition for the isobar data

nigmatkulov:
Values for isobar centrality definition have been updated.


plexoos (approved):
Let's keep it unmerged for the time being. In case you come up with yet another set of numbers you can make changes here

nigmatkulov:
Those numbers will not change anymore. It is final for isobars.

PR #34, opened 2021-06-16
Adding dN/dx parameters for BES-II

nigmatkulov:


StRoot/StPicoEvent/StPicoTrack.h
```diff
@@ -368,7 +368,14 @@ class StPicoTrack : public TObject {
...
 
-  ClassDef(StPicoTrack, 8)
+#if !defined (__TFG__VERSION__)
+  /// Fitted dN/dx
+  Float_t  mDnDx;
+  /// Fitted dN/dx error
+  Float_t  mDnDxError;
+#endif
+
+  ClassDef(StPicoTrack, 9)
```
plexoos: I told you this preprocessor branching would get you sooner or later... Not today though :-)

This new code does not make sense as in the same file you already have this:

https://github.com/star-bnl/star-sw/blob/9ee24f6838d1829043b2a3ffe0f81b034b02148c/StRoot/StPicoEvent/StPicoTrack.h#L321-L326

nigmatkulov: Well, it does make sense. Otherwise the order of data memebers will change for the DEV version.

plexoos: You think the order is important? How?

StRoot/StPicoEvent/StPicoTrack.h
```diff
@@ -124,9 +124,11 @@ class StPicoTrack : public TObject {
   Float_t dEdxPullKaon()      const { return dEdxPull(0.493677,1); }
   Float_t dEdxPullProton()    const { return dEdxPull(0.93827231,1); }
   Float_t dEdxPullElectron()  const { return dEdxPull(0.51099907e-3,1); }
+#endif
+  /// Return dN/dx of a track
```
veprbl:
```suggestion
  /// Return dN/dx for the track
```

StRoot/StPicoEvent/StPicoTrack.h
```diff
@@ -124,9 +124,11 @@ class StPicoTrack : public TObject {
...
   Float_t dNdx() const              { return mDnDx; }
+  /// Return dN/dx error a track
```
veprbl:
```suggestion
  /// Return dN/dx error for the track
```

veprbl (commented):
Comment nitpicks

StRoot/StPicoEvent/StPicoTrack.h
nigmatkulov: In general, it is not very important, but I try not to modify classes in the way data members are layouted in memory in the same way and not brake the order. The main pain in the neck may appear in the case when one has to add special rules for changes in the LinkDef file. Do you suggest to get rid of the preprocessor directive for dN/dx here?

plexoos: I still don't see what problem this is supposed to fix or prevent... I'd say remove it.

StRoot/StPicoDstMaker/StPicoDstMaker.cxx
```diff
@@ -1137,10 +1137,15 @@ void StPicoDstMaker::fillTracks() {
...
-#if defined (__TFG__VERSION__)
     picoTrk->setDndx( gTrk->probPidTraits().dNdxFit() );
     picoTrk->setDndxError( gTrk->probPidTraits().dNdxErrorFit() );
-#endif /* __TFG__VERSION__ */
+    if ( gTrk->primaryTrack() ) {
+      picoTrk->setStatus(1);
+    }
+    else {
+      picoTrk->setStatus(0);
+    }
```
plexoos, line 1147: It looks as if this new `mStatus` member is added to distinguish between primary and global tracks. But we already have `StPicoTrack::IsPrimary()`. Wouldn't it be confusing to have two flags with the same meaning?

starsdong: Grigory, is this PR you mentioned today to resolve for the new library tagging?
Dmitri, my understanding is that mStatus will be set to 1 if this track is attached to any primary vertex (StMuTrack::primaryTrack()). While StPicoTrack::IsPrimary() is true only when the track is associate to the primary vertex stored in picoDst.

plexoos: Xin, if I interpret your statement correctly you are saying that the definition of primary tracks changes in PicoDst w.r.t. MuDst, i.e. some tracks defined as primary in MuDst may not be primary in PicoDst.

I agree with the first part of your statement about tracks defined primary if they associate with any vertex but I can't confirm the second half about the StPicoTrack::IsPrimary(). Looking at the [loop over global MuDst tracks](https://github.com/star-bnl/star-sw/blob/c71b0a035d7b3b31cad4191fe54c7acac1b7e8ac/StRoot/StPicoDstMaker/StPicoDstMaker.cxx#L1030-L1578) where the primary momentum of StPicoTrack is set it is not apparent to me how the definition of the primary tracks changes... It is possible that I missed the logic in this 500+line loop but a reference to the code would help in this case.

plexoos:
Grigory, should we expect more changes in this PR?

StRoot/StPicoDstMaker/StPicoDstMaker.cxx
starsdong: Dmitri, I think this is correct. In MuDst, we store multiple vertices, and tracks associated to any primary vertex are considered primary tracks, though some may not be the primary for the current MuVertex. In PicoDst, we only store one primary vertex (best one for analysis) and then only primary tracks associated with this vertex are flagged as primary in PicoDst. From the analysis point of view, often for HI data, the vertex selection is pretty settled while primary vertex selection in high luminosity / small system collisions (pp or pA), the primary vertex/track selection may need some special studies. I think this may be one reason why many spin analysis still uses MuDst data.

I remember Grigory was thinking also to add multiple vertices into PicoDst too. I think this may be some future development.

starsdong:
Rongrong, it seems that this PR needs your review/approval before getting merged. Please let us know whether you are OK.

marrbnl:
Sorry, I did not realize that that needed my approval. 

Two questions: 
- Do I understand it correctly that now DnDx values and errors are declared if "__TFG__VERSION__" is not defined? Shouldn't it be the other way around?
- What is the purpose of adding "picoTrk->setStatus(1)" since only one primary vertex is stored?

StRoot/StPicoDstMaker/StPicoDstMaker.cxx
plexoos: Hi Xin, I am fine with the logic you describe and the difference between the Pico and MuDst primary track.  My concern was that I don't quite see what you say implemented in the pico code... Anyway, I leave it to you, Grigory and Rongrong to sort out. No further comments from my side.

starsdong:
Dmitri and Rongrong, I should let Grigory comment more here about the purpose of adding mStatus. I remember he brought this up before suggested this change and the logic I see in the picoDst code does the implementation correctly.
gTrack->primaryTrack() is the MuDst function, it will return True for all primary tracks (associated with any primary vertices).

plexoos:
All I am saying is that to me it looks as if the following piece of code would fill primary momentum of Pico tracks for all MuDst primary tracks. I don't see where the redefinition of primary Pico track happens based on the saved pico vertex. Maybe it happens somewhere else in the code. Perhaps Grigory saw the need for the new flag but it is not very obvious without an explanation

https://github.com/star-bnl/star-sw/blob/c71b0a035d7b3b31cad4191fe54c7acac1b7e8ac/StRoot/StPicoDstMaker/StPicoDstMaker.cxx#L1227-L1233

nigmatkulov:
@marrbnl:
> Do I understand it correctly that now DnDx values and errors are declared if "TFG__VERSION" is not defined? Shouldn't it be the other way around?

Nope. Basically it activates dNdx fields (members) but in case it is not activated it will use 0 bytes.

> What is the purpose of adding "picoTrk->setStatus(1)" since only one primary vertex is stored?

Xin is correct. It is needed to address two items: 1) have a hint on the pile-up; 2) minimize number of iterations
when searching for the decay particles.

@plexoos 
>All I am saying is that to me it looks as if the following piece of code would fill primary momentum of Pico tracks for all MuDst primary tracks. I don't see where the redefinition of primary Pico track happens based on the saved pico vertex. Maybe it happens somewhere else in the code. Perhaps Grigory saw the need for the new flag but it is not very obvious without an explanation

I do not understand this question. The pMom keeps momentum of primary tracks from the ONLY ONE vertex. The best one (whatever it means). The status keeps a flag if the track was fitted to any vertex, which is not stored.

plexoos:
> The pMom keeps momentum of primary tracks from the ONLY ONE vertex. The best one (whatever it means).

Can you please point me to the code where this selection is done? It may be that I am just missing it...

There should be a place in the code where pMom for picodst tracks is set to 0 (or remains 0 as default value) for primary mudst tracks not pointing to the best vertex.

marrbnl (approved):

genevb:
We need this in SL21c if we're to produce BES-II data with SL21c, correct?

Thanks,
-Gene

genevb:
> On Jul 20, 2021, at 11:44 AM, Gene Van Buren ***@***.***> wrote:
> We need this in SL21c if we're to produce BES-II data with SL21c, correct?

Or maybe not since we're not including dN/dx in the production? I'd appreciate if someone could clarify this for me.
Thanks,
-Gene

starsdong:
Yes, Gene. This is not immediately for the upcoming 19.6 GeV production. However, I am wondering if the library is not built yet, it is probably better to include this commit in to avoid many branches. This way at least we have the functionalities there for R&D developers to look. I think in the production when these variables are all zeros, ROOT should be able to handle the suppression so the size increase should be negligible.

nigmatkulov:
@genevb we should include it because many strange particle (including hypernuclei) calculations will be affected.
It will also help for the future small system analyses where the pile-up presents.

@plexoos There should be a place in the code where pMom for picodst tracks is set to 0 (or remains 0 as default value).

https://github.com/star-bnl/star-sw/blob/c71b0a035d7b3b31cad4191fe54c7acac1b7e8ac/StRoot/StPicoDstMaker/StPicoDstMaker.cxx#L1111

And the default constructor sets it to 0. At the beginning of the function StPicoDstMaker::fillTracks() we have already defined
a map between global tracks and those which belong to the best primary vertex.

marrbnl:
I have approved the pull request. Please let me know if anything else is needed from my side. 

Best
Rongrong

nigmatkulov:
@marrbnl, to approve the PR we need at least two persons.

plexoos:
> At the beginning of the function StPicoDstMaker::fillTracks() we have already defined
> a map between global tracks and those which belong to the best primary vertex.

Ok, now I recall that StMuDst::primaryTracks(int) returns a track based on the currently "selected" vertex rather than all primary tracks... Fine, it makes then.

plexoos (approved):

genevb:
Dmitri, would you please clarify what hash we use to tag another version of SL21c that includes this update but not all the ROOT6 updates?

Thanks,
-Gene


On Jul 20, 2021, at 12:55 PM, nigmatkulov ***@***.******@***.***>> wrote:

Merged #34<https://github.com/star-bnl/star-sw/pull/34> into main.

plexoos:
Hi Gene,
We just branch from 0d53a1a98fb3cadf755744f8d1e08cab089837dc (which is currently tagged as SL21c) and add this commit to the branch. We can then force reset the SL21c tag to the tip of the new branch

plexoos:
Yesterday, I updated the SL21c tag to include this change in PicoDst but the libraries in /afs do not include it as I confirmed just now. Apparently, Amol built the release before the update so, I asked him to rebuild using the same tag.

PR #36, opened 2021-06-17
Introduce run time switch for dN/dx calculation

plexoos:
The functionality to calculate dN/dx is already in the code. In order to enable it one just need to define an environment variable `STAR_USE_DNDX` before starting processing jobs.

### DIFF
--- a/StRoot/StBFChain/BigFullChain.h
+++ b/StRoot/StBFChain/BigFullChain.h
@@ -1795,6 +1795,7 @@ Bfc_st BFC[] = { // standard chains
 #endif
   {"dEdxY2"       ,"dEdxY2","","tpcDb,StEvent","StdEdxY2Maker","libMinuit,StdEdxY2Maker"
    ,                                                                 "Bichsel method used for dEdx",kFALSE},
+  {"CalcdNdx", "", "", "dEdxY2", "", "",              "Option for StdEdxY2Maker to calculate dN/dx",kFALSE},
 
   // Options in need to be done after the tracker
   // second wave of BTOF options needed after Sti
--- a/StRoot/StBFChain/StBFChain.cxx
+++ b/StRoot/StBFChain/StBFChain.cxx
@@ -785,13 +785,12 @@ Int_t StBFChain::Instantiate()
       }
       mk->PrintAttr();
     }
-    if (maker == "StTpcHitMover" && GetOption("EbyET0")) {
-      mk->SetAttr("EbyET0", kTRUE);
+    if (maker == "StTpcHitMover") {
+      if (GetOption("EbyET0"))            mk->SetAttr("EbyET0", kTRUE);
+      if (GetOption("EmbeddingShortCut")) mk->SetAttr("EmbeddingShortCut", kTRUE);
     }
-    if ((maker == "StdEdxY2Maker"  || maker == "StTpcHitMover") &&
-	GetOption("EmbeddingShortCut"))  {
-      mk->SetAttr("EmbeddingShortCut", kTRUE);
-      mk->PrintAttr();
+    if (maker == "StdEdxY2Maker") {
+      if (GetOption("EmbeddingShortCut")) mk->SetAttr("EmbeddingShortCut", kTRUE);
     }
     if (maker == "StSvtDbMaker" || maker == "StSsdDbMaker"){
       mk->SetMode(0);
--- a/StRoot/StBFChain/StBFChain.cxx
+++ b/StRoot/StBFChain/StBFChain.cxx
@@ -791,6 +791,7 @@ Int_t StBFChain::Instantiate()
     }
     if (maker == "StdEdxY2Maker") {
       if (GetOption("EmbeddingShortCut")) mk->SetAttr("EmbeddingShortCut", kTRUE);
+      if (GetOption("CalcdNdx"))          mk->SetAttr("CalcdNdx", kTRUE);
     }
     if (maker == "StSvtDbMaker" || maker == "StSsdDbMaker"){
       mk->SetMode(0);
--- a/StRoot/StdEdxY2Maker/StdEdxY2Maker.cxx
+++ b/StRoot/StdEdxY2Maker/StdEdxY2Maker.cxx
@@ -169,8 +169,8 @@ Int_t StdEdxY2Maker::InitRun(Int_t RunNumber){
 
   if (! DoOnce) {
     DoOnce = 1;
-    if ((GetDate() > 20171201 && m_TpcdEdxCorrection->IsFixedTarget()) ||
-	(GetDate() > 20181201)) fUsedNdx = kTRUE; // use dN/dx for fixed target for Run XVIII and year >= XIX
+    if (IAttr("CalcdNdx") && ((GetDate() > 20171201 && m_TpcdEdxCorrection->IsFixedTarget()) ||
+	(GetDate() > 20181201))) fUsedNdx = kTRUE; // use dN/dx for fixed target for Run XVIII and year >= XIX
     if (TESTBIT(m_Mode, kCalibration)) {// calibration mode
       if (Debug()) LOG_WARN << "StdEdxY2Maker::InitRun Calibration Mode is On (make calibration histograms)" << endm;
       TFile *f = GetTFile();


plexoos:
@yfisyak Do you have any objections to this proposed change?

kehw:
Such switches are usually implemented as bfc chain options, right? I am afraid an environment variable based switch won't be as visible as the chain options in log files.

plexoos:
Perhaps you are right. To be honest I don't have much experience with bfc chain option implementation. If you know how to do it quickly please submit here

kehw:
I do not know the background of this PR. Is it for some quick fix?
There was a similar discussion in S&C management meeting serval weeks ago. If I understand the discussion right, people also suggested that some infrastructure code is needed around the dNdx calculation.

plexoos:
The background is very simple. The dNdx feature is enabled by default in the code (essentially hard coded). The dNdx feature is not going to be used in production jobs and has to be disabled BUT the main branch is requested to have the feature on. We either provide a switch or produce a branch with a patch turning the feature off for releases. The latter smells unprofessional to me. Also, the tagging for new release has to happen now so, yes, if you can help act now.
 
I don't understand the second part of your comment at all... What infrastructure code needed? Any pointers to the discussion?

kehw:
For an immediate action, I have no problem with code change here. Gene needs to comment if the production team is ok to work with a new environment variable. 
I cannot find the minutes of the meeting I mentioned, but it was about whether we need a chain option for dNdx.

StRoot/StdEdxY2Maker/StdEdxY2Maker.cxx
```diff
@@ -169,8 +169,8 @@ Int_t StdEdxY2Maker::InitRun(Int_t RunNumber){
 
   if (! DoOnce) {
     DoOnce = 1;
-    if ((GetDate() > 20171201 && m_TpcdEdxCorrection->IsFixedTarget()) ||
-	(GetDate() > 20181201)) fUsedNdx = kTRUE; // use dN/dx for fixed target for Run XVIII and year >= XIX
+    if (IAttr("CalcdNdx") && ((GetDate() > 20171201 && m_TpcdEdxCorrection->IsFixedTarget()) ||
```
veprbl, line 172: If it was on by default, this would read something like
```suggestion
    if ((!IAttr("DontCalcdNdx")) && ((GetDate() > 20171201 && m_TpcdEdxCorrection->IsFixedTarget()) ||
```

plexoos: A double negation... How about 

```
GetOption("CalcdNdx") ? mk->SetAttr("CalcdNdx", kTRUE) : mk->SetAttr("CalcdNdx", kFALSE);
```

in  StRoot/StBFChain/StBFChain.cxx instead? Would this work?

veprbl: Well, double negation is the whole point of "unless disabled do X" statements.
BFC should support turning option off with `-CalcdNdx`, but then you need to somehow enable it by default.

StRoot/StBFChain/StBFChain.cxx
```diff
@@ -785,13 +785,13 @@ Int_t StBFChain::Instantiate()
...
+      if (GetOption("EbyET0"))            mk->SetAttr("EbyET0", kTRUE);
+      if (GetOption("EmbeddingShortCut")) mk->SetAttr("EmbeddingShortCut", kTRUE);
     }
-    if ((maker == "StdEdxY2Maker"  || maker == "StTpcHitMover") &&
-	GetOption("EmbeddingShortCut"))  {
-      mk->SetAttr("EmbeddingShortCut", kTRUE);
-      mk->PrintAttr();
+    if (maker == "StdEdxY2Maker") {
+      if (GetOption("EmbeddingShortCut")) mk->SetAttr("EmbeddingShortCut", kTRUE);
+      if (GetOption("CalcdNdx"))          mk->SetAttr("CalcdNdx", kTRUE);
```
veprbl, line 794: The new option needs documentation in BigFullChain.h.

StRoot/StdEdxY2Maker/StdEdxY2Maker.cxx
veprbl: The new commit message reads "By default the calculation is off", so this is formally correct – good to mark as resolved.

Not sure if there is a policy that would prevent a sudden change of the default settings. IMHO it would be nice if there was one.

veprbl (approved):
Change looks fine apart from that it changes the default behaviour

plexoos:
> Change looks fine apart from that it changes the default behaviour

No worries. There was an email (see star-scmgt-l) that the default must be set to off. The "STAR Management" was mentioned four times in that email so, it must mean something 😄

plexoos:
@starsdong Xin, if you want to follow up on this change please do now. You can also approve it and merge. Thanks

genevb (commented):
I have no objections to the proposed changes under StRoot/StBFChain. The github interface seems to provide no way for me to approve only those specific changes in this larger pull request, and I abstain from review of the other proposed code changes to the main branch. Further, I see that the github interface is forcing me to be listed as a reviewer because I'm listed as a CODEOWNER for just a portion of the subdirectories involved in this pull request and it will no longer allow me to decline the role. So I will just have to leave this as a comment and leave it at that.

StRoot/StBFChain/BigFullChain.h
```diff
@@ -1795,6 +1795,7 @@ Bfc_st BFC[] = { // standard chains
 #endif
   {"dEdxY2"       ,"dEdxY2","","tpcDb,StEvent","StdEdxY2Maker","libMinuit,StdEdxY2Maker"
    ,                                                                 "Bichsel method used for dEdx",kFALSE},
+  {"CalcdNdx", "", "", "", "StdEdxY2Maker", "",       "Option for StdEdxY2Maker to calculate dN/dx",kFALSE},
```
genevb: Probably better not to have this option require the StdEdxY2Maker maker, but rather the dEdxY2 option:
"CalcdNdx", "", "", "dEdxY2", "","",...
This ensures that the dEdxY2 option handles the dependencies on other libs for the actual maker.

plexoos: Done

plexoos:
> Xin, if you want to follow up on this change please do now. You can also approve it and merge. Thanks

Xin, yesterday after my comment above you replied to me privately that you were still discussing this change with the STAR management. Is there any news on this? Can we merge already?

StRoot/StBFChain/BigFullChain.h
```diff
@@ -1795,6 +1795,7 @@ Bfc_st BFC[] = { // standard chains
 #endif
   {"dEdxY2"       ,"dEdxY2","","tpcDb,StEvent","StdEdxY2Maker","libMinuit,StdEdxY2Maker"
    ,                                                                 "Bichsel method used for dEdx",kFALSE},
+  {"CalcdNdx", "", "", "", "dEdxY2", "",              "Option for StdEdxY2Maker to calculate dN/dx",kFALSE},
```
genevb, line 1798: Almost.... Dependent options go in the 4th element of the array, not the 5th.

StRoot/StBFChain/BigFullChain.h
genevb: Almost.... Dependent options go in the 4th element of the array, not the 5th.

plexoos:
Ping @starsdong

starsdong:
Dmitri, would you please update the CODEOWNERS file to have Yuri F. as the owner for the dEdxY2Maker? It is preferred to have approvals from the code owners for all PRs. I hope we can conclude on this early next week.

plexoos:
Is there a problem for Yuri to do this himself? He is welcome to create a PR with changes to CODEOWNERS

starsdong:
Hi All,

After discussion with Yuri who is the code owner of dEdxY2Maker, Yuri proposed a different approach to use SkipdNdx option in the chain. See pull request
https://github.com/star-bnl/star-sw/pull/52

If OK with everyone, I would suggest reviewers to take a look at the above #52 request and proceed from there?

Thanks

/in

klendathu2k:
Hi Xin,

Both solutions (PR #36 and #52) allow the decoupling of the dE/dx and 
dN/dx algorithms.  PR #36 makes the decoupling the default, PR #52 
requires a specific chain option to be added to decouple the two.

With PR #36, the "dEdxY2" option reverts back to what it meant prior to 
the inclusion of dN/dx in the dEdx maker (2016-09-18).  Specifically, 
run the dEdxY2 algorithm.

With PR #52, the "dEdxY2" option maintains its meaning from the point 
where dNdx was added: run both algos.

Neither solution is ideal here, because they *both* change the behavior 
of production chains.

But there are a couple of reasons to prefer PR #36.

1) dN/dx has never been used in published results.  Dropping it from the 
production chains is less problematic.
2) dN/dx adds a CPU overhead.  PR #52 requires all users to add 
"skipDnDx" to their chain to avoid this.

Cheers,
Jason

plexoos:
I agree with Jason. #36 should be preferred over #52 because according to PWGC there is no immediate interest in using dNdx. Again, the foreseen production jobs will not include dNdx, therefore, enabling it by default just to use the "skipDnDx" option to turn it off looks strange.

plexoos:
Close in favor of #52. The difference between this and #52 is the default value of the switch turning the dNdx calculation on/off

[impl1](https://github.com/star-bnl/star-sw/pull/36/files#diff-515aa6cf1e18e49a23f1d93766effc076af994a2c1bb0fd7c99e4a801034286aR172)

[impl2](https://github.com/star-bnl/star-sw/pull/52/files#diff-515aa6cf1e18e49a23f1d93766effc076af994a2c1bb0fd7c99e4a801034286aR172)

PR #40, opened 2021-06-25
Added owners to more directories

starsdong:
Change to CODEOWNERS. Added owners to more directories.


genevb (changes requested):
Hi, Xin

I was added as an owner to a lot of directories in this commit, perhaps because at some point in the past I have committed codes in these. But I typically wasn't the maintainer of most of these codes, and I had already put my name beside directories and files I'm interested and willing to help moderate and watch moving forward. I can review each item for which you've added my name in case you've caught something I missed if that's what you would like me to do, but I will remove myself from most.

starsdong (commented):
Gene, understand. I think the proposed changes are fine to me. Thanks

starsdong (commented):
This looks OK to me.

.github/CODEOWNERS
```diff
@@ -1,14 +1,26 @@
 *                                                  @plexoos           @nigmatkulov
-/OnlTools                                          @kehw              @plexoos
+/OnlTools                                          @kehw              @plexoos          @jml985
 /StDb                                              @dmarkh            @plexoos
-/StRoot/RTS                                        @kehw              @akioogawa
+/StDb/idl                                          @akioogawa         @fgeurts          @iraklic        @yfisyak
```
veprbl:
```suggestion
/StDb/idl                                          @akioogawa @dmarkh         @fgeurts          @iraklic @plexoos        @yfisyak
```
To avoid removing those people from the owner list

starsdong: Actually,   a question that both @dmarkh and @plexoos are granted the permission to /StDb, does this also include all subdirectories or they have to be explicitly included when specifying the code owners for subdirectories?

dmarkh: Is it proposed to add the following people to /StDb/idl maintainers list?
 @akioogawa @dmarkh @fgeurts @iraklic @plexoos @yfisyak

Note that IDL file changes must be accompaned by database changes (by StDbLib design), therefore I need to be a reviewer for most (or all) StDb/idl requests, I guess. It is okay to commit an IDL file for the development purposes - in this case there is no corresponding database table (usually there is a test ROOT file), but for production IDLs I just have to be aware of those commits.

veprbl: > Actually, a question that both @dmarkh and @plexoos are granted the permission to /StDb, does this also include all subdirectories or they have to be explicitly included when specifying the code owners for subdirectories?

No, the names need to be specified for subdirectories as well.

plexoos: I appreciate all proposals involving my name but my final decision to take or not ownership of certain modules will heavily depend on how the situation with #36 evolves

starsdong (commented):
Gene, your modification is OK with me. Your commit should be in. But it seems still need your approval before this pull request can get picked up to merge in the main. This is just one step forward. We will need to follow up other experts and update this owner list.

genevb:
>  But it seems still need your approval before this pull request can get picked up to merge in the main. 

I don't think it needs *my* approval.

starsdong:
Dimitri or Grigory, could you please take a look at the updated one with Gene's modification? I have no problem with this. Would be good to get this merged. Then I will follow up with other subsystems.

plexoos (approved):

starsdong (commented):
Modifications accepted.

PR #42, opened 2021-06-29
Add example macros for converting embedding MuDst to PicoDst and analyzing embedding PicoDst

zhux97:


StRoot/macros/embedding/anapicodst/README
```diff
@@ -0,0 +1,7 @@
+how to run:
+1) convert embedding MuDst.root to picoDst.root
+adjust the pico chain setting in genDst.csh, then
+./genDst.csh a.MuDst.root
+Note that only the pico code (StPicoDstMaker/StPicoEvent) in recent STAR library (>= SL21b) includes MC information. One should use the same library (or pico code) to convert real data MuDst and embedding MuDst.
+2) run the example analysis code (SL19b is used here for testing only, StRoot/StPicoDstMaker and StRoot/StPicoEvent should be checked out from git and cons-ed in SL19b for this test)
```
genevb, line 6: Just a suggestion: I imagine this is a good place to offer an instruction to users on how to have local compilations of just these Pico codes, as people running jobs probably don't want to have the entire SL21b-or-later library compilation (gigabytes) copied into each job. I have a bigger concern with this compilation strategy, however, which I'm placing in a separate comment to this pull request.

genevb:
Xianglei, after looking at the README, I'm concerned about the procedure here. It sounds like you want to use a recent version of the PicoDst codes built within an older library, so that the PicoDst data structures can include MC information. We do commonly update (re-compile) simulation-related aspects of old libraries, sometimes creating separated "_embed" libraries as a result, but we are reliant on some level of independence of the simulation-related codes from the reconstruction-related codes. As far as I know, we are not striving to retain this kind of backward compilation compatibility in the PicoDst codes, but perhaps I'm not fully aware. I've added @nigmatkulov (Grigory) to this pull request because it would be good to hear his views on this.

Perhaps if this is desired, we should clearly identify where such compilation compatibility does and doesn't exist (e.g. in which old libraries, SLxx - SLyy, do the PicoDst codes from SLzz compile and work properly).

Alternatively, it may be worth branching the old libraries to have PicoDst codes modified for those specific libraries to include the MC data in simulations. This would likely be a considerable amount of work to ensure the PicoDst codes don't do anything differently for real data in those old libraries.

StRoot/macros/embedding/anapicodst/README
zhux97: Hi Gene, this script is only used for testing the new pico code. I somehow use ACLIC to compile the makePicoDstQA.C, which only works on old libraries like SL19b, the newer library (SL19e, for example) breaks ACLIC.

zhux97:
> Xianglei, after looking at the README, I'm concerned about the procedure here. It sounds like you want to use a recent version of the PicoDst codes built within an older library, so that the PicoDst data structures can include MC information. We do commonly update (re-compile) simulation-related aspects of old libraries, sometimes creating separated "_embed" libraries as a result, but we are reliant on some level of independence of the simulation-related codes from the reconstruction-related codes. As far as I know, we are not striving to retain this kind of backward compilation compatibility in the PicoDst codes, but perhaps I'm not fully aware. I've added @nigmatkulov (Grigory) to this pull request because it would be good to hear his views on this. Perhaps if this is desired, we should clearly identify where such compilation compatibility does and doesn't exist (e.g. in which old libraries, SLxx - SLyy, do the PicoDst codes from SLzz compile and work properly). Alternatively, it may be worth branching the old libraries to have PicoDst codes modified for those specific libraries to include the MC data in simulations.

zhux97:
Sorry, accidentally closed this pull request, now re-opened. Gene, the purpose of this makePicoDstQA.C is to show how MC information can be accessed in embedding PicoDst. I guess for future analysis (like BESII) people can easily adapt their own Pico analysis code accordingly. The SL19b (with the checked-out pico code) is only for testing purpose, so one may not expect the users do the same thing in their analysis.

genevb (approved):
OK, Xianglei. I have no other comments.

plexoos (approved):
This looks fine to me but I am not the code owner for these files.
Should we set a limit on the approval wait time? If there is no response from the designated code owners within, say, 7 days we merge.

genevb:
> Should we set a limit on the approval wait time? If there is no response from the designated code owners within, say, 7 days we merge.

Policy change proposals like that need a wider discussion.

plexoos:
Sure. What is the current policy by the way?

genevb:
You wrote, "Should we set a limit on the approval wait time?", implying that you understand there to be no existing limit with STAR's use of git, and I doubt I have more awareness of STAR's git policies than you. I think it's reasonable to say that creating something that didn't exist is a change of that thing, and in this case that thing would be a policy on "a limit on the approval wait time".

PR #49, opened 2021-07-01
Starsw issue 45

### LINKED ISSUE #45: StarVMC/StarGeometry doesn't compile on main branch
### Steps to reproduce

On RACF run:
```bash
git clone https://github.com/star-bnl/star-sw.git
cd star-sw
git checkout f7935bbbca1c9e9c69931bb5167ca9767be72807
env -i \
  PATH=$PWD/mgr:$PATH \
  QTDIR=/usr/lib64/qt-3.3 QTINC=/usr/lib64/qt-3.3/include \
  QTLIB=/usr/lib64/qt-3.3/lib CERN_LEVEL=pro \
  CERN=/cern CERN_ROOT=/cern/pro \
  ROOTSYS=/afs/rhic.bnl.gov/star/ROOT/5.34.38/.sl73_gcc485/rootdeb \
  ROOT=/afs/rhic.bnl.gov/star/ROOT \
  ROOT_LEVEL=5.34.38 \
  STAR_SYS=x8664_sl7 \
  STAR_HOST_SYS=sl73_gcc485 \
  OPTSTAR=/opt/star/sl73_gcc485 \
  XOPTSTAR=/afs/rhic.bnl.gov/opt/star/sl73_gcc485 \
  STAR_ROOT=/afs/rhic.bnl.gov/star \
  STAR=/nowhere \
  sh -c '. $ROOTSYS/bin/thisroot.sh; exec cons %StShadowMaker'
```

Observe:
```
Install StarVMC/StarGeometry/BtofConfig.cxx as .sl73_gcc485/obj/StarVMC/StarGeometry/BtofConfig.cxx
g++ -m32 -fPIC -pipe -Wall -Woverloaded-virtual -std=c++0x -Wno-long-long -pthread -g -Dsl73_gcc485 -D__ROOT__ -DNEW_DAQ_READER -I. -IStRoot -I.sl73_gcc485/include -I/afs/rhic.bnl.gov/star/ROOT/5.34.38/.sl73_gcc485/rootdeb/include -c .sl73_gcc485/obj/StarVMC/StarGeometry/BtofConfig.cxx -o .sl73_gcc485/obj/StarVMC/StarGeometry/BtofConfig.o
In file included from .sl73_gcc485/obj/StarVMC/StarGeometry/BtofConfig.cxx:1:0:
./StarVMC/StarGeometry/BtofConfig.h:3:45: fatal error: StarVMC/StarGeometry/BtofGeo1.h: No such file or directory # include "StarVMC/StarGeometry/BtofGeo1.h"                                              ^
compilation terminated.
cons: *** [.sl73_gcc485/obj/StarVMC/StarGeometry/BtofConfig.o] Error 1
cons: errors constructing .sl73_gcc485/obj/StarVMC/StarGeometry/BtofConfig.o
```

### Expected result

Everything compiles without errors

cc @klendathu2k

klendathu2k:
Change to mgr/Conscript-standard which may resolve issue #45.   

*IF* I understand the cons documentation correctly, removing the "#" means that the dependencies will be checked relative to the path from which cons is invoked.  With the "#" in place, the dependencies would be checked relative to the path which contains the top level directory containing the Construct file... (which may be under $STAR).   This seems ~ consistent with the behavior I have observed, that the codes under $STAR seem to convince cons that dependencies are satisfied, and no build is really necessary.

Needs review from cons expert(s).


plexoos (approved):
It did not break the CI build which is good enough for me

veprbl:
The removal of `#` doesn't resolve #45. If we were to do that, the `InstallAs` directives would need to be updated too.

veprbl:
The minimal fix seems to be
```diff
diff --git a/mgr/Conscript-standard b/mgr/Conscript-standard
index ec4317355f..b584b6b747 100644
--- a/mgr/Conscript-standard
+++ b/mgr/Conscript-standard
@@ -386,7 +386,7 @@ if ( $pkg !~ /^sim$/ && $pkg !~ /^gen$/ ) {
                            Depends $env ["#".$agX, "#".$agH, "#".$agE], ("#$xml", "#$agml_cmd");
                            # Depends $env --> some rule to build extra dict on the fly
 
-                           Command $env ["#".$agX], ("#$xml"), qq($agml_cmd --file=%1 --module=$agM --export=AgROOT --path=%>:d);
+                           Command $env ["#".$agX, "#".$agH], ("#$xml"), qq($agml_cmd --file=%1 --module=$agM --export=AgROOT --path=%>:d);
                            Command $env ["#".$agE], ("#$xml"), qq($agml_cmd --file=%1 --module=$agM --export=Mortran > %>);
 
                            # Install local file (#) as global
```

I think it would be also nice if it didn't put the generated files to `#StarVMC/StarGeometry`, those should better go straight to `#.sl73_gcc485/obj/StarVMC/StarGeometry` (and headers would need to be moved `#.sl73_gcc485/include/StarVMC/StarGeometry`).

klendathu2k:
Hi Dmitry,

Could you post the test you did to show this failed?

Thanks,
Jason

veprbl:
@klendathu2k See procedure in https://github.com/star-bnl/star-sw/issues/45

klendathu2k:
Hi Dmitry,

There are a few procedures mentioned in that PR, so I am not
sure which one you are referring to.

Using the standard STAR group environment, the following behavior
is expected.  Following a fresh checkout of the code

$ cons +StarVMC/Geometry

should result in the construction of the cxx,h and age files required to
compile the geometry libraries.  Following that, a full build using

$ cons     # (or a partial cons +StarVMC/StarGeometry 
+StarVMC/xgeometry)

should succeed.

I observe this with the proposed fix.

Cheers,
Jason

plexoos:
Shall we merge this one?

veprbl:
I don't see how this is a correct change.

plexoos:
The description of your issue is very terse (i.e. basically no description) making it hard to understand what your concern really is about. I read it as you expect the code to compile with a single call to cons. I think Jason already mentioned that you need to run cons twice in order to compile:

    cons +StarVMC/Geometry
    cons ...

That is also what we do in CI:

https://github.com/star-bnl/star-sw/blob/ecfe0a03767393e85db5fcb91f9d05077c9747fd/docker/Dockerfile.root5#L90-L96

Also setting $STAR to the location of checked out source works fine if you don't need to rely on a centrally installed release.

https://github.com/star-bnl/star-sw/blob/ecfe0a03767393e85db5fcb91f9d05077c9747fd/docker/Dockerfile.root5#L47

veprbl:
@plexoos What you show me appears to be a workaround for #45. The description is terse, but it contains an MRE. The fact that the header is missing during the normal order of compilation means that there is a problem with interdependencies. In this case, it seems like cons is not told that the header files can be produced from agml.

PR #47, opened 2021-06-30
Address differences in ROOT5 and ROOT6 interfaces

plexoos:
The main idea behind the proposed changes is to have the code that can seamlessly work with ROOT5 and ROOT6

PR #50, opened 2021-07-02
Add CI build based on ROOT6

plexoos:
In order to compile our code (`star-sw`) against ROOT6 we need the changes in #47 That is why this PR merges the `ds-pre-root6` branch and as the result includes all the commits on that branch.
(#47 is ready to be merged but if this PR #50 is merged first then #47 will be merged automatically)


docker/Dockerfile.root6
```diff
@@ -0,0 +1,98 @@
...
+ && patch -p1 < root4star_cons.patch
+
+RUN source star-spack/setup.sh && spack env activate star-env-root6 \
+ && export MYSQL=`spack location --install-dir mysql` \
+ && export LIBXML2_DIR=`spack location --install-dir libxml2` \
+ && export Vc_LIB_DIR=`spack location --install-dir vc`/lib \
+ && cd /usr/include && ln -s $MYSQL/include mysql \
+ && cd /star-sw \
+ && cons +asps/staf \
+ && cons +asps/Simulation/agetof \
```
genevb: I am unaware of the need to compile asps/staf and asps/Simulation/agetof before everything else. What led to adding these two lines?

docker/Dockerfile.root6
```diff
@@ -0,0 +1,98 @@
...
+ && cons +StarVMC/Geometry \
+ && cons %pams/sim/g2r %OnlTools \
```
genevb, line 92: One can use the SKIP_DIRS environment variable for many of these skipped subdirectories. Is there an advantage to using the "%" syntax?

docker/Dockerfile.root6
plexoos: Hi Gene: These lines are the result of my trial and error with cons build of the entire codebase because I never saw the actual build command in a script or documentation. If a test confirms that the build works without those two extra calls to cons I'd be happy to remove them.

docker/Dockerfile.root6
plexoos: For me the advantage of using the "%" syntax over SKIP_DIRS is that the former is documented by cons:

https://github.com/star-bnl/star-sw/blob/6a0d42373a26c1ce21b23dc256c8938bf135cf5a/mgr/cons#L124-L127

I am not aware of any documentation explaining the format for SKIP_DIRS or how it should be defined for multiple skipped directories. I hope the final result from using either approach is the same. If there are known differences please let us know.

plexoos:
```
ROOT5
succeeded ... in 1h 1m 52s

ROOT6
succeeded ... in 1h 48m 26s
```
Interesting. Templates?

PR #51, opened 2021-07-02
Code ownres from the best of my knowledge

yfisyak:


yfisyak:
Update code owners from the best my knowledge

plexoos (changes requested):
The syntax of the proposed changes is wrong. Merging it here or as part of #52 will cause the entire file to be ignored.

veprbl:
> The syntax of the proposed changes is wrong. Merging it here or as part of #52 will cause the entire file to be ignored.

Can you clarify which part is broken? Is it the maintainerless lines?

plexoos:
Yes
> The pattern is followed by one or more GitHub usernames

veprbl:
> Yes
> 
> > The pattern is followed by one or more GitHub usernames

Strange. That would mean that one can not remove all maintainers for a subdirectory. Should be easy to fix this PR though.

plexoos:
I believe this PR was superseded by #60. Shall we then close this one?

starsdong:
Dmitri, Yes. I would agree.

PR #52, opened 2021-07-02
Introduce run time switch for dN/dx calculation

yfisyak:
Add skipdNdx option to BFC in order to force disable dN/dx calculations

### DIFF
--- a/StRoot/StBFChain/StBFChain.cxx
+++ b/StRoot/StBFChain/StBFChain.cxx
@@ -793,7 +793,7 @@ Int_t StBFChain::Instantiate()
       mk->SetAttr("EmbeddingShortCut", kTRUE);
       mk->PrintAttr();
     }
-    if (maker == "StdEdxY2Maker" && (GetOption("SkipdNdx") || !(GetOption("CalcdNdx")))) {
+    if (maker == "StdEdxY2Maker" && !GetOption("CalcdNdx")) {
       mk->SetAttr("SkipdNdx", 1);
     } 
     if (maker == "StSvtDbMaker" || maker == "StSsdDbMaker"){
--- a/StRoot/StBFChain/StBFChain.cxx
+++ b/StRoot/StBFChain/StBFChain.cxx
@@ -793,6 +793,9 @@ Int_t StBFChain::Instantiate()
       mk->SetAttr("EmbeddingShortCut", kTRUE);
       mk->PrintAttr();
     }
+    if (maker == "StdEdxY2Maker" && GetOption("SkipdNdx")) {
+      mk->SetAttr("SkipdNdx", 1);
+    } 
     if (maker == "StSvtDbMaker" || maker == "StSsdDbMaker"){
       mk->SetMode(0);
       // If simulation running make sure pick up simu stuff from db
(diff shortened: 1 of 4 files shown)


nigmatkulov (approved):
Looks good to me.

starsdong:
Please take a review of the proposed changes. It is not ideal, but will certainly can do the work. And hopefully OK with everyone. If no objection is raised, I would propose to merge this to the main branch by COB tomorrow so we can prepare the library for upcoming BES-II production.

StRoot/StBFChain/StBFChain.cxx
```diff
@@ -793,6 +793,9 @@ Int_t StBFChain::Instantiate()
       mk->SetAttr("EmbeddingShortCut", kTRUE);
       mk->PrintAttr();
     }
+    if (maker == "StdEdxY2Maker" && (GetOption("SkipdNdx") || !(GetOption("CalcdNdx")))) {
```
plexoos: If `SkipdNdx` and `CalcdNdx` are mutually exclusive this is equivalent to just

```suggestion
    if (maker == "StdEdxY2Maker" && !GetOption("CalcdNdx")) {
```

if `skip = !calc` then

    (skip or !calc) = (skip or skip) = skip = (!calc or !calc) = !calc

klendathu2k: This is not about the name of the chain option, but whether the algorithm should be on by default or not.  Having it on by default 
1) Alters chains used in prior data productions. 
2) Requires all users to add the "skipdndx" option to their analysis chains, otherwise they incur a CPU penalty.

The default should be to disable dN/dx.

starsdong: Dmitri, logic wise, your interpretation is certainly correct. But I would suggest we should just move on in this case if the code does the job and has no bug.

plexoos: > The default should be to disable dN/dx.

Jason, please double check. I think it is the case right now. If you follow the if statement logic it literary says: "If the BFC option CalcdNdx is not set then the **Maker's** option SkipdNdx is set which in turn means dNdx is didabled". Since CalcdNdx is not set in any of the past chains and the production team is not going to insert it the upcoming ones we get the desired behavior by default.

klendathu2k: You're correct.  I didn't think through the logic completely.  The absence of CalcDnDx does instruct the maker to skip the dndx calculation.

plexoos: Xin, in the current implementation the SkipdNdx BFC option is redundant and should be removed. No, we cannot tolerate boolean expressions such as (var OR var) in our code. Maybe you don't care but I do. The only remaining option can be renamed to whatever you like, I don't care about the name. Otherwise, I am fine with the compromise solution.

starsdong:
Hi All, do we still need one approval to get the PR merged?

plexoos:
Xin, Are you ok with incorporating my suggestions above? If so, I can clean up and merge the squashed commits. I believe @genevb is fine with the change

starsdong:
OK, I am fine with the change. Let us merge them to assemble the library.

plexoos (approved):

PR #59, opened 2021-07-16
Star6

perevbnlgov:
"Cons" changes for Root6, merging of DEV Cons + Y.Fisyak + VP. Cons in STAR itself over complicated code.
It is very hard to support. Probably in future it is better to move to CMAKE. Current version does not work yet
for "pams" with non standard Fortran code.

StarGenerator: Full version. Buf i Cons fixed and now full version compiles successfully.

Some old SAR code, based on Root5 features and not used now is obsolete.
Few technical changes for for Root6.

PR #61, opened 2021-07-19
Remove unused Makefiles last touched 19 years ago

plexoos:
Inspired by changes proposed in #59


klendathu2k (approved):
Fine.

genevb (approved):
If the libraries compile without these, then I'm OK with removing them. We have old copies if they need restored for any reason.

PR #62, opened 2021-07-19
Clean up cons build script

plexoos:


genevb (commented):
I don't have much feedback on this. Compiling .DEV2 is not a concern for me.

veprbl (approved):
I don't know what .DEV2 is, but other changes look good.

PR #65, opened 2021-07-21
update loadParams(const int runNumber) and fix the index issue in StRoot/StBTofUtil/StBTofSimResParams.h 

Morning-Ye:
changed the void loadParams(const int date = 20160913, const int time = 175725, const char* Default_time = "2016-09-13 17:57:25")
to void loadParams(const int runNumber = 20076002)
and read the BTOF cell resolution table based on the runnumber

### DIFF
--- a/StRoot/StBTofUtil/StBTofSimResParams.h
+++ b/StRoot/StBTofUtil/StBTofSimResParams.h
@@ -38,7 +38,7 @@ class StBTofSimResParams : public StMaker {
 			if ( itray > 120 || imodule > 32 || icell > 6 )
 				return result;
 
-			return params[ itray ][ imodule * 6 + icell ];
+			return params[ itray-1 ][ (imodule-1) * 6 + (icell-1) ];
 		}
 
 		//! Loads BTOF Sim Params from database
@@ -52,7 +52,7 @@ class StBTofSimResParams : public StMaker {
 			St_db_Maker *dbMk = new St_db_Maker("db", "MySQL:StarDb", "$STAR/StarDb");
 			dbMk->SetDebug();
 			dbMk->SetDateTime(date,1);
-			cout<<"cout for test, data: "<<date<<endl;
+			cout<<"cout for test, date for BTOF resolution table: "<<date<<endl;
 			dbMk->SetFlavor("ofl");
 			dbMk->Init();
 			dbMk->Make();
--- a/StRoot/StBTofUtil/StBTofSimResParams.h
+++ b/StRoot/StBTofUtil/StBTofSimResParams.h
@@ -47,7 +47,7 @@ class StBTofSimResParams : public StMaker {
 		{
 			
 			const int yearNumber = runNumber/1000000 - 1 + 2000;
-			const int date       = yearNumber*1000*10 + 101; // year + 01-01(Date)
+			const int date       = (yearNumber-1)*1000*10 + 1001; // prior year + October 1st(Date)
 
 			St_db_Maker *dbMk = new St_db_Maker("db", "MySQL:StarDb", "$STAR/StarDb");
 			dbMk->SetDebug();
--- a/StRoot/StBTofUtil/StBTofSimResParams.h
+++ b/StRoot/StBTofUtil/StBTofSimResParams.h
@@ -52,7 +52,6 @@ class StBTofSimResParams : public StMaker {
 			St_db_Maker *dbMk = new St_db_Maker("db", "MySQL:StarDb", "$STAR/StarDb");
 			dbMk->SetDebug();
 			dbMk->SetDateTime(date,1);
-			cout<<"cout for test, date for BTOF resolution table: "<<date<<endl;
 			dbMk->SetFlavor("ofl");
 			dbMk->Init();
 			dbMk->Make();
@@ -73,9 +72,6 @@ class StBTofSimResParams : public StMaker {
 			{
 				TDatime val[2];
 				dbMk->GetValidity((TTable*)dataset,val);
-				
-				std::cout << "Dataset validity range: [ " << val[0].GetDate() << "." << val[0].GetTime() << " - "
-					<< val[1].GetDate() << "." << val[1].GetTime() << " ] " << std::endl;
 				
 				tofSimResParams_st* tofTable = static_cast<tofSimResParams_st*>(dataset->GetTable());
 				
--- a/StRoot/StBTofUtil/StBTofSimResParams.h
+++ b/StRoot/StBTofUtil/StBTofSimResParams.h
@@ -69,7 +69,6 @@ class StBTofSimResParams : public StMaker {
 			St_tofSimResParams *dataset = 0;
 			dataset = (St_tofSimResParams*) DB->Find("tofSimResParams");
 
-
 			if(dataset) 
 			{
 				TDatime val[2];
@@ -80,23 +79,24 @@ class StBTofSimResParams : public StMaker {
 				
 				tofSimResParams_st* tofTable = static_cast<tofSimResParams_st*>(dataset->GetTable());
 				
+				const int mNTray         = 120;
+				const int mNCellsPerTray = 192;
+
 				mAverageTimeResTof = 0;
-				for ( int i = 0; i < 120; i++ )
+				for ( int i = 0; i < mNTray; i++ )
 				{
-					for ( int j = 0; j < 192; j++ )
+					for ( int j = 0; j < mNCellsPerTray; j++ )
 					{
-						size_t index = i * 192 + j;
+						size_t index = i * mNCellsPerTray + j;
 						params[i][j] = tofTable[0].resolution[index];
 
 						mAverageTimeResTof += params[i][j];
 
 						LOG_DEBUG << "tray:" << i << ", mod cell:" << j << " = " << tofTable[0].resolution[index]  << " == " << params[i][j] << endm;
-						cout<<"index: "<<index<<endl;
-						LOG_INFO <<  "tray:" << i << ", mod cell:" << j << " = " << tofTable[0].resolution[index]  << " == " << params[i][j] << endm;
 					}
 				}
 
-				mAverageTimeResTof=mAverageTimeResTof/(120*192);
+				mAverageTimeResTof=mAverageTimeResTof/(mNTray*mNCellsPerTray);
 
 				LOG_INFO << "Loaded tofSimResParams. Average = " << mAverageTimeResTof << endm;
 				return;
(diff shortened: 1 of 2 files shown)


Morning-Ye:
Gene mentioned that the 19.6 GeV data production will start soon, shall we approve the merger?

genevb:
> Gene mentioned that the 19.6 GeV data production will start soon, shall we approve the merger?

Are you hoping to have this included in SL21c???

Morning-Ye:
Hi Gene,
Yes, these updates are done for improving the Startless mode BTOF T0s
determination as well as including the nSigmaBTOF values.
Thanks,
Zaochen

genevb:
I see.... So we'll need SL21c tagged and rebuilt yet again....
-Gene

plexoos:
It is so hard to see what was really changed in the code by looking at the diff. There are so many changes in the white space and additional print statements intertwined with presumably real changes... Hopefully, it will make more sense to the maintainers who will approve soon.

veprbl:
I second @plexoos . The change looks fine, but please don't reformat with tabs.

Morning-Ye:
> I second @plexoos . The change looks fine, but please don't reformat with tabs.

Hi Dmitry, Thank you very much, I will never do this again!

genevb:
(EDIT: I see from Zaochen's earlier reply to me that this is needed not just for simulation, but for production as well, reinforcing the point that this needs to be resolved now, so waiting to put these into an SL21c_embed is not an option anyhow.)

Given that embedding for BES-II will likely begin before the production completes, and that we cannot update the library (SL21c) mid-production, we have to wait for this code change to be added before we start the production (unless we create another library, SL21c_embed, though we would preferably not do that). Please understand that there is high pressure to get this production started, so please wrap up this code change (and any others?) that are needed for BES-II with urgency.

starsdong:
Frank and Daniel, would it be possible for you to take a review about this change and approve at your earliest convenience?
From my side, the change looks OK. I hope Zaochen has tested this with the Run19 data production?
One potential concern is that the initialization date is always 0101 (January 1st in each calendar year). It should be OK for the BES-II datasets (Run19, 20, 21 etc.). But for some year that starts before Jan. 1st. (e.g. the upcoming 2022 run),  one will need to take care of these runs happening before Jan. 1st.

Morning-Ye:
Hi Xin,

For the changes I made, I have tested with the Run19 daq files, it works
fine. For the initialization date, as you already pointed out, this should
be good for the BES-II. I originally planned to set the timestamp as the
input parameter to grep the tables from DB, however, I noticed that it will
need more changes in the StBTofCalibrationMaker Init() function. So I did
not do it that way. Do we want to make more changes on it now or leave it
as the current version and take care in future data production?

Thanks,
Zaochen

starsdong:
Zaochen, Glad that you have thought about this. Indeed, that would be the best and clean solution.
One more question, do we need to worry about the backward compatibility of this commit?

Morning-Ye:
Hi Xin, there is only one set of resolution table in the DB, the previous
uploaded table with messed index were overwritten with the new one
(2014-2017). However I want to mention it again here, the new BTOF
resolution table is obtained based on the Run18 Isobar 200 GeV
calibrations. For those years earlier than 2018, the resolution tables may
need to be restudied based on Run14 or Run16 data. These resolution tables
are used for nSigmaBTOF calculations, actually if people don't use this
variable for their analysis, there should be no problem. As the tables were
there, thus the backward compatibility should be ok.
However, I am not sure whether Bassam's changes could be backward
compatible?

genevb:
On this point...

On Jul 27, 2021, at 5:35 PM, Xin Dong ***@***.******@***.***>> wrote:

One potential concern is that the initialization date is always 0101 (January 1st in each calendar year). It should be OK for the BES-II datasets (Run19, 20, 21 etc.). But for some year that starts before Jan. 1st. (e.g. the upcoming 2022 run), one will need to take care of these runs happening before Jan. 1st.


...the safe bet is October 1st of the prior calendar year, which is the start of the fiscal year.

-Gene

Morning-Ye:
Hi Gene,

Good suggestion, so the easy solution will be
change "*const int date = yearNumber*1000*10 + 101*;" to "*const int date =
(yearNumber-1)*1000*10 + 1001;*"
If there are no obligations, I will make this change tomorrow.

Thanks,
Zaochen

starsdong:
Zaochen, thanks. The new commit looks good to me. If you can do a quick test with the 19.6 GeV daq file to be sure something goes wrong with this last commit. Thanks

starsdong (approved):
Thanks, Zaochen. The new commits look OK to me.

Morning-Ye:
I am thinking maybe I should let Bassam or someone else submit the pull
request, then I can also approve it as a code reviewer. Since Frank and
Daniel are not available.

starsdong:
Dmitri, I am not sure if Frank or Daniel is immediately available. Is it OK to add another reviewer here for approval so it can get merged later?

plexoos:
Ok, we can merge with the available feedback.
Can Zaochen provide here at least some instruction how he tested the change?

StRoot/StBTofUtil/StBTofSimResParams.h
```diff
@@ -20,80 +21,99 @@
...
+					for ( int j = 0; j < 192; j++ )
+					{
+						size_t index = i * 192 + j;
+						params[i][j] = tofTable[0].resolution[index];
+
+						mAverageTimeResTof += params[i][j];
+
+						LOG_DEBUG << "tray:" << i << ", mod cell:" << j << " = " << tofTable[0].resolution[index]  << " == " << params[i][j] << endm;
+						cout<<"index: "<<index<<endl;
+						LOG_INFO <<  "tray:" << i << ", mod cell:" << j << " = " << tofTable[0].resolution[index]  << " == " << params[i][j] << endm;
```
plexoos: Do these LOG_DEBUG and LOG_INFO print out exactly the same information?
Remove LOG_INFO. Remove cout

StRoot/StBTofUtil/StBTofSimResParams.h
```diff
@@ -20,80 +21,99 @@
...
+
+
+			if(dataset) 
+			{
+				TDatime val[2];
+				dbMk->GetValidity((TTable*)dataset,val);
+				
+				std::cout << "Dataset validity range: [ " << val[0].GetDate() << "." << val[0].GetTime() << " - "
+					<< val[1].GetDate() << "." << val[1].GetTime() << " ] " << std::endl;
+
```
plexoos: Is this print statement really necessary? Remove it

StRoot/StBTofUtil/StBTofSimResParams.h
```diff
@@ -20,80 +21,99 @@
...
+			const int date       = yearNumber*1000*10 + 101; // year + 01-01(Date)
 
-		if (!DB) {
-			LOG_WARN << "No data set found, creating new St_db_Maker... with date/time" << date << "/" << time << endm;
-			dbMk = new St_db_Maker("db", "MySQL:StarDb", "$STAR/StarDb");
+			St_db_Maker *dbMk = new St_db_Maker("db", "MySQL:StarDb", "$STAR/StarDb");
 			dbMk->SetDebug();
-			dbMk->SetDateTime(date,time); //! event or run start time, set to your liking
+			dbMk->SetDateTime(date,1);
+			cout<<"cout for test, data: "<<date<<endl;
```
plexoos: This clearly for debugging only. Remove this cout

plexoos (changes requested):

Morning-Ye:
I ran root4star -b -q -l 'bfc.C ( 1, 10, "DbV20210423 P2019a StiCA
-beamline3D btof BEmcChkStat CorrY -OPr13 EbyET0 evout PicoVtxVpdOrDefault
PicoCovMtxWrite -hitfilt btofCalib vpdCalib pppAMode",
"/star/u/bassam/forGrigory/scratch/daqFor2019AuAu19p6/st_physics_adc_20076002_raw_0000004.daq"
)' > & ! log_st_physics_adc_20076002_raw_0000004 &

and then look at the print out table values (in the i,j loop, currently
removed the cout lines) to what I uploaded to DB, and found they are the
same.

StRoot/StBTofUtil/StBTofSimResParams.h
Morning-Ye: Will do in next update

StRoot/StBTofUtil/StBTofSimResParams.h
Morning-Ye: Will remove it

StRoot/StBTofUtil/StBTofSimResParams.h
Morning-Ye: I only ran LOG_INFO in previous test, LOG_INFO could print out the full table without problem.
LOG_INFO line was removed

StRoot/StBTofUtil/StBTofSimResParams.h
```diff
@@ -20,80 +21,93 @@
...
-	void loadParams(const int date = 20160913, const int time = 175725, const char* Default_time = "2016-09-13 17:57:25")
-	{
-
-		St_db_Maker *dbMk = dynamic_cast<St_db_Maker*>( GetChain()->GetMakerInheritsFrom("St_db_Maker") );
+			return params[ itray-1 ][ (imodule-1) * 6 + (icell-1) ];
+		}
 
-		TDataSet *DB = GetDataBase("Calibrations/tof/tofSimResParams");
+		//! Loads BTOF Sim Params from database
+		//void loadParams(const int date = 20160913, const int time = 175725, const char* Default_time = "2016-09-13 17:57:25")
```
plexoos, line 45:
```suggestion
```

StRoot/StBTofUtil/StBTofSimResParams.h
```diff
@@ -20,80 +21,93 @@
...
-
-		if (dataset) {
-			TDatime val[3];
-			dbMk->GetValidity((TTable*)dataset,val);
-			tofSimResParams_st* table = static_cast<tofSimResParams_st*>(dataset->GetTable());
-
-      mAverageTimeResTof=0;
-			for ( int i = 0; i < 120; i++ ){ //  nTrays
-				for ( int j = 0; j < 192; j++ ){
-					size_t index = i * 120 + j;
```
plexoos, line 78: Do I understand correctly that the main objective of this PR was to fix the problem with this index?

```diff
-size_t index = i * 120 + j;
+size_t index = i * 192 + j;
```
If so, can it be mentioned in the PR title? The current title is not very informative. It will become a log message when we squash and merge

Morning-Ye: There are two main changes: 1) fix this index issue; 2) load the table according to the runnumber. Both of these changes are within the function loadParmas().

PR #66, opened 2021-07-21
Remove unmaintained packages

plexoos:
I investigated a bit the history of the StHbtMaker and StSpinMaker packages and found that they were excluded from all releases since at least SL15h and SL03g respectively. Propose to remove them as they appear to be unmaintained and outdated.

If anyone knows of any other unmaintained code that should be removed feel free to add it here.


genevb:
I don't think it would be appropriate to remove these without approval from the PWGs (conveners) in which HBT and Spin physics are analyzed.

From the Production Team side, I tried searching for whether these shared object libraries are loaded for any of the myriad nightly tests and found none, so I expect we probably don't need to keep these for the purpose of supporting our ability to re-run an old production.

genevb (commented):
I don't think it would be appropriate to remove these without approval from the PWGs (conveners) in which HBT and Spin physics are analyzed.

From the Production Team side, I tried searching for whether these shared object libraries are loaded for any of the myriad nightly tests and found none, so I expect we probably don't need to keep these for the purpose of supporting our ability to re-run an old production.

plexoos:
Thanks for checking. Given these packages have been excluded from all builds and releases for quite a long time it would be very strange if they were still used in the nightly tests.

I believe Akio is the POC for StSpinMaker but who should we contact for StHbtMaker?

marrbnl:
I will bring this into the attention of conveners, and see if anyone is still using them.

akioogawa:
Are you proposing to remove them from GitHub repository? Are we going back to CVS?

nigmatkulov:
Hi Dmitri, StHbtMaker was maintained by Mike Lisa in past.
Before moving forward, I think this is a very bad idea to remove packages that were pushed
to the repository even if there are no maintainers at the moment. 
Those are very important for the physics analysis. Since there is no area for the physics analysis codes
at the moment we MUST NOT remove such codes.

Why did you start looking for such makers?

plexoos:
> Are you proposing to remove them from GitHub repository?

Yes

> Are we going back to CVS?

No

akioogawa:
So... where are we putting those codes???

plexoos:
> Before moving forward, I think this is a very bad idea to remove packages that were pushed
> to the repository even if there are no maintainers at the moment.

Why is it a bad idea to remove something that is not currently used? Do you realize that we are talking about removing from the head of the main branch but all files remain in the history?

> Those are very important for the physics analysis. Since there is no area for the physics analysis codes
> at the moment we MUST NOT remove such codes.

I don't know what the "area for the physics analysis codes" means. Do you know that these "very important" codes cannot be built as is?

> Why did you start looking for such makers?

We need to simplify the build command. See:
https://github.com/star-bnl/star-sw/blob/8d4ab95842415f688cab59a673bce0c3edbd1c01/docker/Dockerfile.root5#L91-L96

There is a reason why we exclude these packages from the build

plexoos:
> So... where are we putting those codes???

Someone who cares about that code should decide. Do you care about it? Can you make it at least compilable?

marrbnl:
I have brought this to the attention of conveners, and they will discuss within relevant PWGs to see if these codes are needed. If so, we can discuss how to make it compile and how to maintain them moving forward.

akioogawa:
StSpinPoll was excluded from nightly build long time ago. But it is still used today, updated, and new makers are written and being committed here. I'm not sure what does "removing from the head of the main branch but all files remain in the history" means. Can I still fork/pull/commit/push and create PR?

nigmatkulov:
> Why is it a bad idea to remove something that is not currently used? Do you realize that we are talking about removing from the head of the main branch but all files remain in the history?

People will not look at the history. 

>I don't know what the "area for the physics analysis codes" means. 

The action you want to implement can be done ONLY in the case when codes for the physics analysis (and not only)
to which people invested huge amount of time will be stored somewhere. 
Cleaning something for the better compilation is good in a pure programing but not in physics.
For example, it means that there should be specific area in this repository that will be dedicated to the physics analysis codes.

> Do you know that these "very important" codes cannot be built as is?

Not sure I understand your question.

plexoos:
Akio, why are you bringing StSpinPool into this discussion? This PR is about removing StRoot/StSpinMaker https://github.com/star-bnl/star-sw/pull/66/commits/a65c6a735e40fb865c1462d8b7f61ecedd5b8f84

Are these two packages related somehow?

Although, StSpinPool could not compile either until recently I fixed it for you 531caf7a6cdbe2bd599e621f1fd819771edf16df Maybe you haven't even noticed that...

akioogawa:
So when it is off from nightly build, why do you care? I hope cons/make/whatever still support those makers when checked out...

nigmatkulov:
In other words, Dmitri, we need to and must keep the codes that are related to the physics analysis unless
there is a trivial way to get those codes and compile without any extra (non-standard) actions.

nigmatkulov:
> So when it is off from nightly build, why do you care? I hope cons/make/whatever still support those makers when checked out...

I second Akio's question.

akioogawa:
Sorry! I guess I did not read careful enough. Then why was I invited to review for StSpinMaker? Oh wait. I wrote this in 2001? I had no memory :D. Looking at it.... I'm 99.9% sure that this maker can go bye-bye, even from history.

plexoos:
Superseded by #69

PR #72, opened 2021-07-26
Remove reference to unsupported ROOT Qt class

plexoos:
Support for Qt library as dependency for STAR software has stopped some time ago.


genevb (approved):
Sure. Bye bye Qt.

veprbl (approved):

PR #74, opened 2021-07-27
Allow cons users to specify locations for external packages Vc and FastJet

plexoos:
Currently concerns only StJetFinder, TPCCATracker, and StiCA packages

### DIFF
--- a/mgr/Conscript-standard
+++ b/mgr/Conscript-standard
@@ -790,7 +790,12 @@ if ( $pkg !~ /^sim$/ && $pkg !~ /^gen$/ ) {
 	#print "-=-=-=-=-=-= SO_PKG = $SO_PKG\n";
 
     } elsif ( $pkg =~ m/^StJetFinder$/ ){
-	$CXXFLAGS .=  " -I".$env1->{ENV}->{OPTSTAR}."/include";
+
+        if ( defined($ENV{FASTJET_DIR}) ) {
+          $CPPPATH = $ENV{FASTJET_DIR}."/include".$main::PATH_SEPARATOR.$CPPPATH;
+        } else {
+          $CPPPATH = $env1->{ENV}->{OPTSTAR}."/include".$main::PATH_SEPARATOR.$CPPPATH;
+        }
 
     } elsif ( $pkg =~ m/^StFwdTrackMaker$/ ){ 
 		$CXXFLAGS .=  " -I".$env1->{ENV}->{OPTSTAR}."/include";
--- a/mgr/Conscript-standard
+++ b/mgr/Conscript-standard
@@ -971,13 +971,18 @@ if ( $pkg !~ /^sim$/ && $pkg !~ /^gen$/ ) {
 #	$CXXFLAGS .= " -MMD -DTBB_DO_ASSERT -DDO_ITT_NOTIFY -DUSE_PTHREAD -D__TBB_BUILD=1 -Wall -Wno-parentheses";
 #
 #
-    }  elsif ( $pkg =~ m/^TPCCATracker$/) {
+    } elsif ($pkg eq "TPCCATracker" or $pkg eq "StiCA") {
         $CPPPATH .=
 	  $main::PATH_SEPARATOR . "#" . $Dir . $main::PATH_SEPARATOR . "#" . $Dir . "/code";
 #	  $main::PATH_SEPARATOR . "#" . $Dir . $main::PATH_SEPARATOR . "#" . $Dir . "/code" .
 #	    $main::PATH_SEPARATOR . "#" . $Dir . "/code/KFParticle" .  $main::PATH_SEPARATOR . "#" . $Dir . "/code/Performance";
 	#    $CPPPATH .= $main::PATH_SEPARATOR . "#StRoot/TBB/include";
 	#    print "CPPPATH => $CPPPATH\n";
+
+        if ( defined($ENV{Vc_DIR}) ) {
+          $CPPPATH = $ENV{Vc_DIR}."/include".$main::PATH_SEPARATOR.$CPPPATH;
+          $LIBPATH = $ENV{Vc_DIR}."/lib";
+        }
 
 	my $extraCXXFLAGS = " -DNVALGRIND -DENABLE_VECTORIZATION -W -Wall -Wimplicit -Wswitch -Wformat -Wchar-subscripts";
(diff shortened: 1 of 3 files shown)


mgr/Conscript-standard
```diff
@@ -792,6 +792,10 @@ if ( $pkg !~ /^sim$/ && $pkg !~ /^gen$/ ) {
     } elsif ( $pkg =~ m/^StJetFinder$/ ){
 	$CXXFLAGS .=  " -I".$env1->{ENV}->{OPTSTAR}."/include";
 
+        if ( defined($ENV{FASTJET_DIR}) ) {
+          $CPPPATH = $ENV{FASTJET_DIR}."/include".$main::PATH_SEPARATOR.$CPPPATH;
+        }
```
veprbl: What about path to the libraries? Also, I wonder if you've considered querying paths using `fastjet-config`.

plexoos: > What about path to the libraries? 

Are you suggesting to link against the fastjet lib? Sure, why not. But I don't really know how to do it properly with cons. Can you make a concrete suggestion?

> Also, I wonder if you've considered querying paths using `fastjet-config`.

No, I just set the path by hand:

https://github.com/star-bnl/star-sw/pull/75/commits/3a4b404c0742a9eefe10b904a4169c682394985e

mgr/Conscript-standard
```diff
@@ -792,6 +792,10 @@ if ( $pkg !~ /^sim$/ && $pkg !~ /^gen$/ ) {
     } elsif ( $pkg =~ m/^StJetFinder$/ ){
 	$CXXFLAGS .=  " -I".$env1->{ENV}->{OPTSTAR}."/include";
```
veprbl: Should we change
```suggestion
	$CPPFLAGS .=  " -I".$env1->{ENV}->{OPTSTAR}."/include";
```
to ensure that `FASTJET_DIR` overrides `OPTSTAR`?

mgr/Conscript-standard
```diff
@@ -998,7 +1007,6 @@ if ( $pkg !~ /^sim$/ && $pkg !~ /^gen$/ ) {
 #	if ($STAR_SYS =~ 'gcc432$') {$CXXFLAGS .= " -DVC_IMPL=Scalar";}
 #	else                        {$CXXFLAGS .= " -DVC_IMPL=SSE"}
 	# c++ `root-config --ldflags` -W -Wall -Wimplicit -Wswitch -Wformat -Wchar-subscripts -Wparentheses -Wmultichar -Wtrigraphs -Wpointer-arith -Wcast-align -Wreturn-type -Wno-unused-function -pedantic -Wno-long-long -Wno-variadic-macros -msse -msse2 -Wshadow -fno-threadsafe-statics -Os -DNDEBUG -fPIC -rdynamic -DENABLE_VECTORIZATION `root-config --glibs` -lEG  ./StRoot/TPCCATracker/CA.C -I/star/u/ikulakov/tests/test5/StRoot/TPCCATracker/code -I/afs/rhic.bnl.gov/star/ROOT/5.99.99/.sl53_gcc450/rootdeb/include  -I/star/u/ikulakov/tests/test5/.sl53_gcc450/include  -L/star/u/ikulakov/tests/test5/.sl53_gcc450/lib -lVc -lTPCCATracker -lTBB -o CA
-	if ( defined($ENV{Vc_LIB_DIR}) ) { $LIBPATH = $ENV{Vc_LIB_DIR}; }
```
veprbl, line 1001: Perhaps it would be a good sense to impose a clear migration suggestion
```perl
if ( defined($ENV{Vc_LIB_DIR}) ) { die("Vc_LIB_DIR is not supported anymore. Please use Vc_DIR."); }
```

mgr/Conscript-standard
veprbl: > Are you suggesting to link against the fastjet lib? Sure, why not. But I don't really know how to do it properly with cons. Can you make a concrete suggestion?

I've forgotten that we load those dynamically:
https://github.com/star-bnl/star-sw/blob/ecfe0a03767393e85db5fcb91f9d05077c9747fd/StRoot/StJetMaker/macros/RunJetFinder2012pro.C#L34
I would say there are some benefits to linking directly, but that's outside of the scope of this PR.

> No, I just set the path by hand:
> 
> [3a4b404](https://github.com/star-bnl/star-sw/commit/3a4b404c0742a9eefe10b904a4169c682394985e)

For sure, this is the intention. Was wondering if you think it's a good idea. I guess, it's not that important.

mgr/Conscript-standard
plexoos: I have absolutely no problem modifying anything in Cons scripts as long as it works 😺 
But it seems unlikely that Cons provides a functionality to override $CPPFLAGS with $CPPPATH contents... Besides, StJetFinder may include something else from $OPTSTAR/include

veprbl: By override I meant that the `-I` that comes earlier will take precedence. My exact change is incorrect, as you pointed out, it should have been targeting CPPPATH.

```suggestion
	$CPPPATH = $env1->{ENV}->{OPTSTAR}."/include".$main::PATH_SEPARATOR.$CPPPATH;
```

> Besides, StJetFinder may include something else from $OPTSTAR/include

Actually, it's safe to say that it will not. IMO we should move OPTSTAR handling to the else branch of the condition.

plexoos: Fine, let's implement the "if-else" logic

plexoos:
Sorry for the force-push

veprbl:
This is strange. The CI complains about Vc_LIB_DIR even for ROOT6.

plexoos:
I forgot that the spack environment sets it
https://github.com/star-bnl/star-spack/commit/2f4ff6d8e6c2e6258db864ec96bdd0d6b20db5d7#diff-2d4e6f710604d87f720478538df079641194e2b08ec7a5256b4169a39764c086

Let's just ignore it as in the original commit. Not worth the sweat

veprbl:
> Let's just ignore it as in the original commit.

`unset Vc_LIB_DIR`?

plexoos:
Thanks @veprbl

veprbl (approved):
Tested to build on RCF (with OPTSTAR only).
This should not have effect in the STAR group environment as it doesn't set Vc_LIB_DIR.

PR #75, opened 2021-07-27
New CI with optimized docker build

plexoos:


plexoos:
Tried to merge the current main into this but found that it is much easier to deal with the conflicts by rebasing instead

docker/Dockerfile.root6
```diff
@@ -40,7 +34,11 @@ RUN source star-spack/setup.sh \
  && spack env create star-env-root6 star-spack/environments/star-x86_64-root-6.16.00-container.yaml \
  && spack env activate star-env-root6 \
  && spack install -j 5 --fail-fast \
- && spack gc -y
+ && spack gc -y \
+ && spack clean --all \
+ && spack uninstall --force -y llvm \
+ # Create star-spack/spack/var/spack/environments/star-env-root5/loads
+ && spack -e star-env-root5 env loads
```
veprbl: star-env-root6?

veprbl: NVM, it's renamed in the next commit

plexoos: Github marked this diff as Outdated. Somehow, you are looking at an old commit... Yes, force pushing is evil, confusing, and should be avoided

veprbl (approved):
It's a big changeset, but only affects CI. Diff LGTM.

docker/Dockerfile.root6
plexoos: > NVM, it's renamed in the next commit

You are actually correct. Sorry for the noise

PR #76, opened 2021-07-29
EbyET0 in simulation

genevb:
The EbyET0 corrected in StTpcHitMoverMaker (reconstruction) must also be un-corrected in StTpcRSMaker (simulation).

This modification will be needed for SL21c embedding as that library will use the EbyET0 correction in reconstruction.


genevb:
I had no intent to delete StRoot/StBFChain/doc/StBFChain.cxx_doc, if someone can instruct me how to exclude that deletion from the PR.

plexoos:
Easy peasy (assuming your remote origin = git@github.com:genevb/star-sw.git):

```
git checkout EbyET0-in-Simulation
# Do a soft reset of the last commit on the EbyET0-in-Simulation branch
git reset HEAD^
# Get rid of any changes in the following files
git restore StRoot/StBFChain/doc/StBFChain.cxx_doc StRoot/StBFChain/doc/index.html
# Stage the remaining intended changes
git add -u .
 # Commit and reuse the original commit message
git commit -C origin/EbyET0-in-Simulation
# You have to force push because this branch name already exists in the remote repo
git push -f origin
```

plexoos (approved):
I don't see any problem with this update

plexoos:
Gene, please let me know when you think you have everything ready and I will add it to SL21c (I may be slow to respond today due to limited internet access)

genevb:
On Jul 29, 2021, at 9:38 AM, Dmitri Smirnov ***@***.***> wrote:
> Gene, please let me know when you think you have everything ready and I will add it to SL21c (I may be slow to respond today due to limited internet access)

With this (PR#76) and the BTOF updates (PR#65), we are all set as far as I know for known issues that are being resolved. Anything else will need to be addressed at another time.

-Gene

genevb:
To be clear: this patch affects only simulation, not reconstruction.  In principle, production can proceed without this, but it should be used in embedding. If we don't get this into SL21c now, we would need to find an opportunity to update SL21c before embedding begins. It occurs to me that there may be an opportunity during the HPSS down time scheduled for next week if this doesn't make it into SL21c immediately (now). I would very much like to get the Run 19 19.6 GeV production moving before that HPSS downtime, but that implies getting SL21c frozen and built immediately.

plexoos:
Ok, it's done. Both #65 and #76 are on the SL21 branch and SL21c release is tagged. The libraries on afs/cvmfs can be updated using the SL21c tag

PR #78, opened 2021-07-30
TpcRS will need StEventUtilities lib for EbyET0

genevb:
The changes I proposed in #76 unfortunately add another dependency for TpcRS. This PR handles that dependency.


genevb:
We have already submitted some initial reconstruction jobs using SL21c to run over the weekend. There will be an opportunity to update SL21c next week when HPSS is down and no production can be run.

starsdong:
Gene, it is nice to have some test samples so we can take some quick look at the data quality to make sure everything is included. Thanks.

plexoos:
If reco data will be already produced with the current SL21c version it may make sense to bump the version when we include this fix. The new version can be either SL21c_1 or SL21d (the latter is perhaps somewhat easier to refer to in the future for major reconstruction jobs) New tags cost nothing and I don't think we will run out of letters.

In general, it is not a good idea to move tags as we did recently while adding more fixes and should be avoided in the future but at least there was no data produced.

plexoos (approved):

genevb:
On Jul 30, 2021, at 1:59 PM, Dmitri Smirnov ***@***.******@***.***>> wrote:

If reco data will be already produced with the current SL21c version it may make sense to bump the version when we include this fix. The new version can be either SL21c_1 or SL21d (the latter is perhaps somewhat easier to refer to in the future for major reconstruction jobs) New tags cost nothing and I don't think we will run out of letters.

I advocate for SL21c_v2 as the tag, keeping with past convention of patching an existing library. But I would also like to wait until next week to apply the new tag just in case any other last-second bugs & patches come.

That said, we shouldn't wait for the tagged version to get this particular patch into main so that DEV's nightly tests can help us with further verification.

Thanks,
-Gene

PR #80, opened 2021-08-04
Minor cleanup changes

plexoos:


genevb (approved):
I don't have a problem with removing CVS-specific files from git. I guess the git review process is intended to catch and prevent any unnecessarily committed PDFs, root files, etc.

veprbl (approved):

PR #91, opened 2021-08-10
ci: Run build action on all pushes and pull requests to any branch

veprbl:
It doesn't make sense to treat non-main branch PR's specially.

It's also desirable to run actions on direct-pushed tags and we also want users to be able to run CI in their forks before creating the PR.

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -1,8 +1,6 @@
 name: CI Build
 
-on:
-  pull_request:
-    branches: main
+on: [push, pull_request]
 
 jobs:
   ROOT5:


plexoos (approved):
That is fine

plexoos:
Hm, the number of jobs doubles: https://github.com/star-bnl/star-sw/pull/93 Now we have jobs triggered from push and pull_request events?

veprbl:
Strange. IIRC this is unlike on travis or on gitlab-ci that would not spawn a separate build.

Maybe use https://github.com/marketplace/actions/skip-duplicate-actions ?

PR #93, opened 2021-08-10
[root6] Include StRoot/StFgtPool into CI build

plexoos:
Declare nested types public in StRoot/StFgtPool

This affects persistent classes processed by ROOT to create dictionaries and
streamers. Unlike rootcint in ROOT5, rootcling seemingly respects the access
modifiers and needs all such types to be publicly defined.

Similar to 9bce66cfe89a6f3cc69b449547726e9b1d32b516


veprbl (dismissed):
Diff LGTM

veprbl:
So previous error was because of `StFgtStatusMaker::status_t`?

plexoos:
> So previous error was because of `StFgtStatusMaker::status_t`?

Yep, it should be public as well

veprbl (approved):
Diff still LGTM, and now this builds, so that's even better :)

PR #95, opened 2021-08-11
Revert "ci: run build action on all pushes and pull requests to any branch

plexoos:
This reverts commit c25a2fdf215f9deafdf4e727aea8fab19818c1ea.

The original commit caused duplication of CI jobs triggered by both push and
pull_request events

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -1,6 +1,8 @@
 name: CI Build
 
-on: [push, pull_request]
+on:
+  pull_request:
+    branches: main
 
 jobs:
   ROOT5:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-build.yml
```diff
@@ -1,6 +1,8 @@
 name: CI Build
 
-on: [push, pull_request]
+on:
+  pull_request:
+    branches: main
```
veprbl: It's worth building PR's to any branch
```suggestion
on: [pull_request]
```

veprbl (approved):
I support the revert. We want to always do the "pull_request" triggered build (with merging to main). The "push" builds (without merge) are nice too, but they also result in extra rebuilds.

PR #96, opened 2021-08-11
Include StRoot/StarGenerator/Kinematics into CI build

### LINKED ISSUE #81: Revisiting dependency on C++ boost
As far as I can tell, the only package that depends on boost is [StRoot/StarGenerator/Kinematics](https://github.com/star-bnl/star-sw/tree/main/StRoot/StarGenerator/Kinematics). It includes the following header

https://github.com/star-bnl/star-sw/blob/cd4338393c9bb095f94e223bc673732275b2a74d/StRoot/StarGenerator/Kinematics/StarKinematics.cxx#L7

and makes use of `boost:split()`

https://github.com/star-bnl/star-sw/blob/cd4338393c9bb095f94e223bc673732275b2a74d/StRoot/StarGenerator/Kinematics/StarKinematics.cxx#L85

Although some version of boost is available on the farm nodes, it is not targeted as direct dependency by the current Spack environment. Since boost is missing from the environment StRoot/StarGenerator/Kinematics cannot be compiled and is excluded from the build

https://github.com/star-bnl/star-sw/blob/cd4338393c9bb095f94e223bc673732275b2a74d/docker/Dockerfile.root5#L121-L130

Installing the entire boost package for just a single function (`boost:split`) does not seem very practical... Would it make more sense to replace the call to `boost::split()` with a few lines of equivalent code?

@klendathu2k

plexoos:
Resolves #81

### DIFF
--- a/StRoot/StarGenerator/Kinematics/StarKinematics.cxx
+++ b/StRoot/StarGenerator/Kinematics/StarKinematics.cxx
@@ -4,7 +4,6 @@
 #include "StarGenerator/UTIL/StarParticleData.h"
 #include "StarGenerator/UTIL/StarRandom.h"
 
-#include <boost/algorithm/string.hpp>
 #include <vector>
 #include <string>
 #include <algorithm>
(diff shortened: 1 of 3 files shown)


klendathu2k (approved):
FIne.  No sense carrying around an extra dependency for this.

perevbnlgov (approved):
To remove redundant include is always good

PR #98, opened 2021-08-12
Include StRoot/StHbtMaker into CI build

plexoos:


veprbl (approved):

mgr/Conscript-standard
```diff
@@ -731,8 +731,13 @@ if ( $pkg !~ /^sim$/ && $pkg !~ /^gen$/ ) {
 	    $CPPPATH .= $main::PATH_SEPARATOR . "#" . $Dir . "/" . $subdir;
 	}
 	if ( $env1->{CPPPATH} ) { $CPPPATH .=  $main::PATH_SEPARATOR . $env1->{CPPPATH};}
-	# $CPPPATH .= $main::PATH_SEPARATOR . "/opt/star/include";
-	$CPPPATH .= $main::PATH_SEPARATOR . $env1->{ENV}->{OPTSTAR}."/include";
+
+        if ( defined($ENV{GSL_DIR}) ) {
+          $CPPPATH = $ENV{GSL_DIR}."/include".$main::PATH_SEPARATOR.$CPPPATH;
+        } else {
+          $CPPPATH = $env1->{ENV}->{OPTSTAR}."/include".$main::PATH_SEPARATOR.$CPPPATH;
```
genevb, line 738: Why change the order of $CPPPATH components?

plexoos: No real reason. I just used the previous similar changes 3a55cdb and a3e2d4e for external FastJet and Vc packages as a template. 

I believe the order of $OPTSTAR and $CPPPATH does not matter because they are not supposed to have overlapping packages/codes. If I read the idea from the Cons scripts correctly, the former is responsible for the path(s) to all external packages except ROOT whereas the latter is responsible for all packages within this repository + ROOT. 

(For Vc the order is more important because it can be installed as part of ROOT in which case $CPPPATH would contain $ROOTSYS/include and thus to pick up Vc .h-files from a different location it has to appear before $ROOTSYS/include.)

veprbl: I also had the same question, but thought that we'd fix this later and in other places as well.

plexoos: What should be fixed?

plexoos:
Please let me know of any further comments/suggestions. If not, let's merge

PR #99, opened 2021-08-12
Remove unused install scripts from mgr/

plexoos:
This had to be done along with 830a472e


veprbl (approved):
> This had to be done along with [830a472](https://github.com/star-bnl/star-sw/commit/830a472e4cdec7eb29b94928f9cb865bafc8cae8)

That's right

genevb (approved):
I won't stop you.

PR #103, opened 2021-08-13
Update StarGenerator and starsim codes in preparation for geant4star integration

klendathu2k:
Changes to the StarGenerator and starsim codes in preparation for geant4star integration. 
- Introduce ageventread_() as a callback from starsim (agzio.age) triggering event generation.  This allows existing workflows with St_geant_maker to remain untouched.
-  Goal was to remove the dependency between StarGenerator and the starsim codes.  It will be difficult to eliminate the dependence on the agseventread_ (&etc...) methods in StarGenerator.  We will need to supply these dummy symbols in geant4star at integration.
- Introduce BFC chain options to load StarGenerator framework, and some select concrete generators
- Several code cleanups identified in review
- StarParticleData no longer created at library load (static instance of object vs dynamically instanstantiated)
- "usestarsim" attribute added (default) for StarPrimaryMaker.
- Introduced trivial test macro.


plexoos:
> cons seems to work fine when I compile on rcf.

That is exactly the problem with the environment on rcf. You move/rename a file in your local directory but cons can still pick up the old header files from the central location pointed by $STAR

StRoot/St_geant_Maker/St_geant_Maker.cxx
```diff
@@ -1177,6 +1180,17 @@ int St_geant_Maker::InitRun(int run){
...
 //_____________________________________________________________________________
 int St_geant_Maker::Make() {
+
+  static bool first = true;
+  if ( first ) {    // Event generator may have been plugged into chain after Init()
+
+    auto* stack = GetDataSet("PrimaryMakerStack");
+    auto* data  = GetDataSet("PDG_Database");
+    if ( stack != 0 )       AgStarReader::Instance().SetStack( (StarParticleStack*)stack->GetObject() );
+    if ( data  != 0 )       AgStarReader::Instance().SetParticleData( (StarParticleData*)data->GetObject() );
```
plexoos: Per STAR coding guidelines C-style type casting should be avoided.

StRoot/St_geant_Maker/St_geant_Maker.cxx
```diff
@@ -694,6 +694,9 @@
 #include "g2t/St_g2t_fts_Module.h"
 #include "g2t/St_g2t_stg_Module.h"
 #include "g2t/St_g2t_epd_Module.h"
+// Event Generator/starsim integration
+#include "AgStarReader.h"
```
plexoos: I suggest you prefix header files with their respective module names even when they belong to the same directory. This way it is easier for other developers to identify the module where it is coming from.

```suggestion
#include "St_geant_Maker/AgStarReader.h"
```

StRoot/StarGenerator/UTIL/StarParticleData.cxx
```diff
@@ -51,7 +51,7 @@ class G3TrackingCode : public StarTrackingCode {
 };
 
 // Instance should exist from the start
-StarParticleData StarParticleData::sInstance;
+StarParticleData* StarParticleData::sInstance = 0;
```
plexoos:
```suggestion
StarParticleData* StarParticleData::sInstance = nullptr;
```

StRoot/StarGenerator/macros/stargentest.py
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env python
+import datetime, os
+import ROOT
+import subprocess, shlex
+import sys
```
plexoos: datetime, ROOT, and sys are not used

StRoot/St_geant_Maker/St_geant_Maker.cxx
plexoos:
```suggestion
    if ( stack ) AgStarReader::Instance().SetStack( (StarParticleStack*)stack->GetObject() );
    if ( data )  AgStarReader::Instance().SetParticleData( (StarParticleData*)data->GetObject() );
```

klendathu2k: This is one of those cases where it is hard to avoid.  If I use static / dynamic casts, then the dynamic linker complains that StarParticleStack / Data are not completely defined when I load the St_geant_Maker library.  Same result if I push the cast down into the AgStarReader class.  The crime against good coding style is not the use of the c-style cast, but rather the use of the singleton pattern for AgStarReader.  Would prefer to keep the c-style cast here, rather than re-engineer the fortran interface.

StRoot/StarGenerator/macros/stargentest.py
klendathu2k: Removed... will push shortly.

StRoot/St_geant_Maker/St_geant_Maker.cxx
plexoos: Are you trying to avoid `#include "StarGenerator/UTIL/StarParticleData.h"` and `#include "StarGenerator/BASE/StarParticleStack.h"` from `StRoot/St_geant_Maker/St_geant_Maker.cxx`?

klendathu2k:
Yes.  Also trying to avoid any run time dependence between the two 
libraries, so that
St_geant_Maker can be loaded without StarGenerator, and StarGenerator 
can be loaded w/out
loading St_geant_Maker.

StRoot/St_geant_Maker/St_geant_Maker.cxx
klendathu2k: Yes.  Also trying to avoid any run time dependence between the two libraries, so that
St_geant_Maker can be loaded without StarGenerator, and StarGenerator can be loaded w/out
loading St_geant_Maker.

(Also, not sure if replying in email gets logged as part of the discussion on github?  If so, this may be redundant.)

plexoos:
Jason, Can you illustrate with a macro what you are trying to achieve. It is not very clear, how you expect to avoid a run time dependence if one library makes use of the other one. As far as I know, a dynamic loader does not complain about unresolved symbols but if you try to call something that depends on the second library the process will definitely crash.

klendathu2k:
Any (most) of the macros under StRoot/StarGenerator/macros will demonstrate... e.g.

$ ln -s StRoot/StarGenerator/macros/starsim.pythia6.C starsim.C
$ root4star -q -b starsim.C
...
As implemented, code runs.  To add a static or dynamic cast requires that I include the StarParticleStack and Data headers.  Including those... I get this when I run starsim.C...

dlopen error: /gpfs/mnt/gpfs01/star/simu/simu/jwebb/STAR/.sl73_gcc485/lib/libSt_geant_Maker.so: undefined symbol: _ZTI16StarParticleData
Load Error: Failed to load Dynamic link library /gpfs/mnt/gpfs01/star/simu/simu/jwebb/STAR/.sl73_gcc485/lib/libSt_geant_Maker.so

$ c++filt _ZTI16StarParticleData
typeinfo for StarParticleData

plexoos:
> As implemented, code runs.

I could not reproduce this. When run from the current main your test goes through and produces an fzd file but from G4star it seg faults.

klendathu2k:
I get a segmentation fault if I compile star-sw without compiling star-mcgen.  Otherwise it worked (to be confirmed again...)

klendathu2k:
Decided on a different path.  Instead of removing the dependence on starsim functions from StarGenerator (agsvert, agskine, etc...), the plan will be to satisfy these with dummy symbols inserted into the geant4 maker.  Alternatively I could create an interface base class, and a plugin library to support starsim (but this seems like overkill).

StRoot/StBFChain/BigFullChain.h
```diff
@@ -1363,6 +1363,15 @@ Bfc_st BFC[] = { // standard chains
   {"gstarLib","","",""                                                 ,"","gstar","Load gstar lib",kFALSE},
   {"flux"        ,"","","simu"                                           ,"","flux","Load flux lib",kFALSE},
   {"------------","-----------","-----------","------------------------------------------","","","",kFALSE},
+  {"Generators  ","-----------","-----------","------------------------------------------","","","",kFALSE},
```
klendathu2k, line 1366: DId not intend to commit this, but it is a harmless addition to our list of chain options (and will be utilized as we merge in the G4star codes).

genevb: If I understand correctly, if you update the branch from which you are making the PR not to have this line, then it will drop off the list of changed files. Then you can re-apply this change later when it is needed.

klendathu2k:
Yes, that's my understanding as well.  This would otherwise be in my 
next pull request, so not worth the effort to retract.

StRoot/StBFChain/BigFullChain.h
genevb: Can you clarify if it is only the single "Generators" line that isn't needed for this commit, or all the BigFullChain.h line changes in this PR that aren't really needed for this PR? git makes my approval mandatory for the whole shebang of this PR because I'm a code owner for a small part of it :-(

klendathu2k:
Hi Gene,

It's the group of ~5 event generator declarations which follows (as well 
as that "Generators" header).

Jason

klendathu2k:
Just pinging this.  Are there any other recommendations to push these changes?

genevb:
> Just pinging this. Are there any other recommendations to push these changes?

As long as your other pull request (for which these chain options matter) and this one point to each other, I'm not too particular.

genevb (commented):
No opposition.

StRoot/StarGenerator/BASE/AgStarReader.h
```diff
@@ -22,6 +22,10 @@ class AgStarReader : public TObject
   /// @param stack The particle stack
   void SetStack( StarParticleStack *stack ){ mStack = stack; }
 
+  /// Set the particle data
+  /// @param data Pointer to the particle data
+  void SetParticleData( StarParticleData* data ){ mParticleData = data; }
```
veprbl: Why add this setter and keep the field if you replace it with a call to `StarParticleData::instance()` in the `AgStarReader::ReadEvent()`?

plexoos:
The description states that "This removes from StarGenerator the dependence on code in the starsim library." I don't see how the submitted changes do anything to achieve this. The original intention to move the code that depends on St_geant_Maker from StarGenerator/ looked more like the move in the right direction but it was reverted.

Please update the description and title to reflect what is really being changed.

StRoot/StarGenerator/BASE/AgStarReader.h
klendathu2k: Solution to a problem that I had to go a different direction on.  Removed.

plexoos:
Thanks for updating the description Jason.

> It will be difficult to eliminate the dependence on the agseventread_ (&etc...) methods in StarGenerator.

This one is still confusing because `ageventread` is defined in StarGenerator... I am probably missing something but where is `ageventread` called from?

klendathu2k:
This is deep in the dark recesses of starsim...

https://github.com/star-bnl/star-sw/pull/103/files#diff-60be39524263b2d623a8cb4328ae1ce1e44cbf95a429a2bfad492e80127ea853

The block staring around line 113.  We ask comis for the address of the 
'AgUSRead'
subroutine.  If it doesn't exist, we ask it for the address of the 
'AgUSEvent' subroutine.
If an address was found, we call the subroutine at that point (line 
117).

plexoos:
Without tests it is very hard to judge whether these changes will break anything or not...
For example, I can't tell if it is safe to remove the definition of `agusread` https://github.com/star-bnl/star-sw/pull/103/files#diff-dcecc0e93bbe5fcf050ef4b946135b831eec3686ed0b1606a8355c62fca004e0L30 when there seems to be another one defined in https://github.com/star-bnl/star-sw/blob/main/pams/sim/gstar/gstar_input.g#L84

Anyway, if you are confident about these changes we can merge... or maybe it can wait until introduction of geant4star when these preparatory changes will be actually used.

klendathu2k:
Yes I am confident that the code functions and is correct.  agusread --> agusevent is simply a renaming of a subroutine, plus a branch under a switch statement to enable the different path.  The code runs.  stargentest.py confirms that events are produced.

plexoos (approved):
The author claims that all changes are necessary for Geant4 integration and have been tested

asps/Simulation/starsim/agzio/agzkine.age
```diff
@@ -110,8 +110,13 @@
       If (IKINEold==-2) call AgNTread(ier)
       If (IKINEold<=-3) call AgFread (ier)
       If (IKINEold<=-4)
-      { J=CsADDR('AGUSREAD'); If (J>0) Call CSjCAL(J,1,Ier,0,0,0,0,0,0,0,0,0) }
+      { 
 
+        J             =   CsADDR('AGUSREAD')
+        IF (J <= 0) J =   CsADDR('AGEVENTREAD')
+        If (J>0) Call CSjCAL(J,1,Ier,0,0,0,0,0,0,0,0,0)
```
veprbl, line 117: `J <= 0` and `J>0` will never coincide. Why set J at all if it's never gonna be used?

If it is used, it would help to swap the statements and add a comment
```suggestion
        If (J>0) Call CSjCAL(J,1,Ier,0,0,0,0,0,0,0,0,0)
c       Set J to be used blablabla
        IF (J <= 0) J =   CsADDR('AGEVENTREAD')
```

veprbl: I now remembered. Reading this last time I understood correctly. `J <= 0` means lookup fail, so we try `AGEVENTREAD`. Never mind then.

veprbl (approved):

PR #105, opened 2021-08-16
Update SCGL macro

genevb:
Several small improvements:

- sce_init calculation more robust by avoiding small denominators
- apply all initial guesses for gapf params if fit fails for FIX_GL
- VAR must be re-calculated at the end of FitWithOutlierRemoval() because the individual parameter fits can change the results (apparently for the better from what I can tell)
- function lines on plots vs. luminosity may need to go to negative luminosities in PCA fits for data with observed negative SpaceCharge

### DIFF
--- a/StRoot/macros/calib/Calib_SC_GL.C
+++ b/StRoot/macros/calib/Calib_SC_GL.C
@@ -166,7 +166,7 @@ int  Waiting();
 void SetMinMax(int n, Double_t* ar, double& min, double& max, double inclusion, double margin=0.075);
 int  SetMinuitPars(const int n, TString* names, Double_t* starts, Double_t* steps, int debug, Bool_t* fixing=0);
 void Log2Lin(int i);
-int  FitWithOutlierRemoval(int debug);
+int  FitWithOutlierRemoval(int npar, int debug);
 void DrawErrorContours(int npar, Bool_t* fix=0);
 TString PCA(int Nmax=32, int debug=0);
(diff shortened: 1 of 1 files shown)


StRoot/macros/calib/Calib_SC_GL.C
```diff
@@ -2036,18 +2041,23 @@ int FitWithOutlierRemoval(int debug) {
...
-  VAR_all = TMath::Sqrt(VAR_all/((Double_t) nMeasures));
   memcpy(outliers,outliers1,nBytes);
+  VAR = 0;
+  VAR_all = 0;
+  for (i=0; i<nMeasures; i++) {
+    if (!outliers[i]) VAR += devs[i]*devs[i];
+    VAR_all += devs[i]*devs[i];
+  }
+  VAR = TMath::Sqrt(VAR/((Double_t) nMeasuresI-1));
+  VAR_all = TMath::Sqrt(VAR_all/((Double_t) nMeasures-1));
```
plexoos, line 2060: In arithmetic expressions involving doubles integer types are implicitly converted to double.  So, the cast to `(Double_t)` in `VAR_all/((Double_t) nMeasures-1)` is not really needed.

plexoos (approved):

StRoot/macros/calib/Calib_SC_GL.C
genevb: Thanks for the point, Dmitri. The casting makes clear what is happening, and is not detrimental, so I would rather leave it and move on.

PR #106, opened 2021-08-17
Add test jobs to CI

plexoos:


plexoos:
Changes in StRoot/macro were approved in #71, the rest is CI-related only

PR #88, opened 2021-08-08
Yue Hang Leung parameters for 26p5GeV_fixedTarget_2018 TpcRS parameters

fisyakbnl:

PR #107, opened 2021-08-19
Add local DB parameters for TpcRS 26p5GeV_fixedTarget_2018

plexoos:
This replaces #88 due to the original PR branch removed by the submitter

Screen Shot 2021-08-19 at 11 16 58 AM

Close #88 when merged


plexoos:
Already approved by @iraklic and @yfisyak

iraklic:
@plexoos this is a little confusing. what happened here. Did you remove the comments and made a new PR?

plexoos:
Yes

iraklic:
got it. and looks like it does not need a new approval, right?

plexoos:
Yes, approved in the original #88. Thank you

genevb:
Just an FYI on this...
Because these parameters are stored as CINT files in the code library instead of in the database, any embedding/simulation libraries that need these parameters must be re-tagged when including this. At this point my understanding is that **SL19e** needs this and will need an `SL19e_2` tag. I'm looking into whether anything else needs updated in **SL19e** before we move forward with the tag.

PR #108, opened 2021-08-19
cons: remove dead platforms support

veprbl:
Since STAR has moved to c++11 standard it's unlikely that those could ever work again.

### DIFF
--- a/mgr/ConsDefs.pm
+++ b/mgr/ConsDefs.pm
@@ -314,63 +314,61 @@
     # Above was a generic gcc support
     # ============================================================
 
-    # cernlib if not intel WNT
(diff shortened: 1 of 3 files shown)


genevb:
Github requires my approval on this, but honestly, this is a lot of stuff I know almost nothing about.  I would again suggest getting the approval of @jlauret in my place.

plexoos (approved):
I looked through the proposed changes carefully and fully support the removal of the build configurations for the outdated platforms. As I understand, it is the responsibility of the Infrastructure Team to maintain the build system for various targets but I am not aware of any user interested in sun, alpha, or winNT systems.

PR #109, opened 2021-08-19
cons: drop x86_darwin support

veprbl:
Following discussion in https://github.com/star-bnl/star-sw/commit/61e9d4d1df854dd67d82c583fae0dde6564abbdc#

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
   - StRoot/StarGenerator/Kinematics (#96)
   - StRoot/StHbtMaker into CI build (#98)
 - Introduce new StFcsCosmicMaker in StRoot/SpinPool (#58)
+- Removed special support for x86_darwin/FINK platform
 
 
 ### Miscellaneous
--- a/mgr/ConsDefs.pm
+++ b/mgr/ConsDefs.pm
@@ -135,7 +135,7 @@
 	$LLIB = "lib";
     }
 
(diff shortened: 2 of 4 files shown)


genevb (commented):
I know the /sw paths helped support [fink](https://www.finkproject.org) installations on Mac OS X. We never ran productions on that platform, and my understanding is that containers now provide the capability to run on that platform and others. That said, there could be users compiling for their analyses that are leveraging the /sw paths, so I would suggest widely announcing this well in advance of removal to allow for potential feedback.

veprbl:
Right, I've added an anouncement in the changelog

plexoos:
Are we dropping support for MacOS for good?

yfisyak:
I am not.

 

Are we dropping support for MacOS officially?

—
You are receiving this because your review was requested.
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications on the go with GitHub Mobile for iOS or Android.

plexoos:
> I am not.

I am a bit puzzled by this comment. Care to elaborate?

starsdong:
Hi All, have we resolved the questions on this conservation? Is it still the intention now to merge this to main? Thanks

veprbl:
I don't think we run into danger of losing some valuable work here, as we can always pick out useful bits from the git history. I'm all for darwin support, but this stuff has not been updated for a decade and it seemingly was tied to an undocumented private environment. I'd do the removal, unless @yfisyak is actually using this nowadays.

PR #116, opened 2021-08-23
Add new BFC option for StPicoDstMaker: PicoVtxMtd

genevb:
It seems that adding a BFC option for PicoVtxMtd was never done. Here is a proposal for it. This will be needed for **SL21c**. I'll leave it to Dmitri whether we can just re-define `SL21c_1` with these changes (because we haven't actually done anything official with the `SL21c_1` code), or do we need an `SL21c_2` for some reason.


plexoos (approved):

veprbl:
Tags are free. Let's not ever retag, please.

plexoos:
> Tags are free. Let's not ever retag, please.

Yes, I agree. Although, all "minor" releases such as SL21c_1 or SL21c_2 still end up in the same location (i.e. in .../SL21c) on the afs and cvmfs volumes

veprbl:
I secretly hope that there is a backup of those on HPSS :D

genevb:
@nigmatkulov , please comment on this ASAP. The RACF farm is empty of production jobs because all jobs are waiting on the SL21c library. We cannot start jobs that don't need the PicoVtxMtd option because we don't want to recompile in the middle of an ongoing production.

plexoos:
It is midnight in GMT+3, I don't think we should wait to merge until morning

plexoos:
Added this commit to SL21 and tagged as SL21c_2

https://github.com/star-bnl/star-sw/commits/SL21

PR #117, opened 2021-08-23
Add EPD hit information to g2t_track data struct

klendathu2k:
Minor fix to MC track model.  Includes number of EPD hits and location of first hit on track.


veprbl (approved):

plexoos:
This PR would be more complete if the newly introduced struct members were used somehow in the code. I grepped for a few other members in this struct and confirmed that they are indeed either filled with values or being read from.

klendathu2k:
Code change is anticipating code in geant4 maker.  However, we should fill the structure in starsim as well.  Most recent push takes care of that.

plexoos (approved):

plexoos:
Any comments from @akioogawa or @perevbnlgov?

PR #118, opened 2021-08-24
Build docker containers via GitHub Actions

veprbl:
Currently all published containers https://github.com/orgs/star-bnl/packages?repo_name=star-sw (called "packages" in GitHub UI) are the ones built and pushed by @plexoos. The *-base containers are the ones used as cache in the CI builds, and any change to the base stage of the Dockerfiles would require for the containers to be updated as well. The proposed workflow allows to update the containers automatically.

Another responsibility of the new workflow is to build *-build containers for the release tags, e.g. ready to be used via Singularity. This has a potential to replace the current release workflow.

The following design decisions were taken:
 - Only build containers on pushes to any of the branches to the main star-bnl/star-sw GitHub repository. Might want to restrict this further to set `on.push.branches` to `['main', 'SL*']` to keep PR's clean.
 - The tag label has a form of `<branch>_latest`. Unfortunately "/" is not allowed in the Docker label, so a build for a branch called "pr/feature" would fail for now. Perhaps we could add `if: ${{ !contains(github.ref, "refs/heads/pr/") }}` or a step to substitute "/" with another character.
 - Got rid of hardcoded digest. Don't think we need it.
 - The "main" branch will build to "main_latest" Docker tag, but would better build to "latest". Unfortunately, simple string operations in GHA can only be implemented via running scripts in separate steps.

### DIFF
--- a/docker/Dockerfile.root5
+++ b/docker/Dockerfile.root5
@@ -28,12 +28,12 @@ RUN mkdir -p star-spack \
  && curl -sL https://github.com/spack/spack/archive/010e5761d6631eae102d1eec87e5e74a0e16ed82.tar.gz | tar -xz --strip-components 1 -C star-spack/spack
 
 RUN source star-spack/setup.sh \
+ && sed -e '/mesa/a\    variants: ~llvm' -i star-spack/environments/packages_gcc-4.8.5.yaml \
  && spack env create star-env-root star-spack/environments/star-x86_64-root-5.34.38-container.yaml \
  && spack env activate star-env-root \
  && spack install -j 5 --fail-fast \
  && spack gc -y \
  && spack clean --all \
- && spack uninstall --force -y llvm \
  # Create star-spack/spack/var/spack/environments/star-env-root/loads
  && spack -e star-env-root env loads
 
--- a/docker/Dockerfile.root5
+++ b/docker/Dockerfile.root5
@@ -28,12 +28,12 @@ RUN mkdir -p star-spack \
  && curl -sL https://github.com/spack/spack/archive/010e5761d6631eae102d1eec87e5e74a0e16ed82.tar.gz | tar -xz --strip-components 1 -C star-spack/spack
 
 RUN source star-spack/setup.sh \
- && sed -e '/mesa/a\    variants: ~llvm' -i star-spack/environments/packages_gcc-4.8.5.yaml \
  && spack env create star-env-root star-spack/environments/star-x86_64-root-5.34.38-container.yaml \
  && spack env activate star-env-root \
  && spack install -j 5 --fail-fast \
  && spack gc -y \
  && spack clean --all \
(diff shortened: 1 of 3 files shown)


.github/workflows/build-containers.yml
```diff
@@ -0,0 +1,80 @@
+name: Build Containers
```
veprbl: Once the first workflow runs, this name stays in the history indefinitely. So if we want to rename, better do it before this is merged.

plexoos:
Very nice. I want to see how long it takes to build the whole shebang. Let's merge

veprbl:
> Very nice. I want to see how long it takes to build the whole shebang. Let's merge

Thanks. It's about 3 hours for star-sw-root5-base and 4.5 for star-sw-root6-base

docker/Dockerfile.root5
```diff
@@ -28,12 +28,12 @@ RUN mkdir -p star-spack \
  && curl -sL https://github.com/spack/spack/archive/010e5761d6631eae102d1eec87e5e74a0e16ed82.tar.gz | tar -xz --strip-components 1 -C star-spack/spack
 
 RUN source star-spack/setup.sh \
+ && sed -e '/mesa/a\    variants: ~llvm' -i star-spack/environments/packages_gcc-4.8.5.yaml \
```
plexoos: Any reason to not modify the file directly in star-spack? A separate one can be created for containers if needed

veprbl: My goal was to contribute a thingy to build containers, so the focus here is to demonstrate its feasibility. We can make the change across repositories beforehand if you prefer.

veprbl:
> It's about 3 hours for star-sw-root5-base and 4.5 for star-sw-root6-base

Actually, I think I've mixed up those numbers. It looks like it's actually about 4 h 15 min for star-sw-root5-base and 5 h 15 min for star-sw-root6-base. Which makes it impossible to build root6-base and root6-build in a single shot with the current environment. The tag job should build on the second attempt due to caching of the base stage.

veprbl:
I'm becoming inclined towards limiting `on.push.branches` to `['main', 'SL*']`. This would avoid errors on `pr/` branches and, more importantly, avoid grey skipped jobs in all PR's (`on` is "stronger" than `if`).

starsdong:
Hi All, the request was made a while ago? Is the intention still to merge this in? If so, let us follow along and get this in. Otherwise, we can close the PR.

veprbl:
I'm not sure where are we with this. The original goal was to easily deploy modifications to the docker environment. This PR should work. The only downside I see with this approach is that it barely fits in the 6 hour limit of the GitHub Action CI. An alternative could be to cache at a spack package level (someone else would need to agree to look into that).

veprbl:
Closing in the light of https://github.com/star-bnl/star-sw/pull/349

PR #121, opened 2021-08-26
[root6] Install pcm files along with libraries

plexoos:


klendathu2k (approved):
Looks like a straightforward addition of files to the build that should otherwise leave the software libraries unchanged.  Using root-config to pickup the version of ROOT used to build the code is a nice touch.

veprbl (approved):
We do need to install precompiled headers

PR #122, opened 2021-08-26
docker/Dockerfile.root{5,6}: build mesa without LLVM

plexoos:
It's needed for compiling shaders, which ROOT should not need. This
change allows container compilation to fit within the 6 hour limit.


starsdong:
Dmitri, you indicated this is a draft? Shall we drop it at this moment or we want to continue to make it a real request?

PR #128, opened 2021-08-30
Fix compiler error: 'upc' was not declared in this scope

kehw:


OnlTools/Jevp/StJevpBuilders/l4Builder.cxx
```diff
@@ -930,6 +930,7 @@ void l4Builder::event(daqReader *rdr)
 
     XX(0);
     unsigned int decision = hlt_eve->hltDecision;
+    const int upc = 0x0;  // DAQ trigger id
```
plexoos, line 933: So, by setting it to zero you essentially skip all of the conditions such as this one:
https://github.com/star-bnl/star-sw/blob/948846a8dc8b9a891f42397889c2dbdfa1c644ec/OnlTools/Jevp/StJevpBuilders/l4Builder.cxx#L1034

If that's the intention I am fine with the change

kehw: Yes, this is the intention. Because we did not run HLT with UPC in the past a few years. Those UPC related code are meant to be skipped.

plexoos (approved):

PR #132, opened 2021-09-01
Star6 StTofHitMaker

perevbnlgov:
Move struct TofRawHit into private to avoid problems with Cling
Change ClassDef(StTofHitMaker, 1) to ClassDef(StTofHitMaker, 0) again to avoid Cling problem.
In the next update I will replace all maker ClassDef to 0. ClassDef non 0 means
I/O for maker, which is senseless

Star6 StTofHitMaker
Move struct TofRawHit into private to avoid problems with Cling
Change ClassDef(StTofHitMaker, 1) to ClassDef(StTofHitMaker, 0) again to avoid Cling problem.
In the next update I will replace all maker ClassDef to 0. ClassDef non 0 means
I/O for maker, which is senseless

In ROOT6 kCanRebbib is obsolete.
Star6 StQAMakerBase.cxx In class StarRoot/TH1Helper::SetCanRebin(..) was developed intead and used

Look comments in St.h
St.h St.cxx St::call(...)
New class which support Cling macro call. In Root5 when Cint called macro from macro
these maccro are called independently. When submacro is defined but not called even if
such submacro does not exist or could not be called Cint is working smoothly.
In Cling, all the submacros are compiled even if this submacro is not really called.
In result your macro does not work if you have any problem in subbmacro, which is not even called.
To solve this preoblem using gROOT->ProcessLine(...) is proposed.
It is easy to do if there is no parameters. But parameters, especially with pointers is much
complicated. You must convert this pointers to the addresses.In result this process line
becames extremely long and unreadable. Example:
gROOT->ProcessLine("subMacro(char*)0x1234567,char*)0x1234567, ....")
Using method St::call it becames mor easy:
char *name-...; char *name2=..;
St::call("subMacro",name,name2)
Class name St was selrcted only to be maximally short and simple.
Arguments for St::call could are different and most popular. If it is needed to add
one, class Sr could be easy  changed

Star6 Load.C
Macro load completely rewrittem. It loads mpstly used set of classes

UPD g2r
Directory pams/sim/g2r removed. It is not usd anymore.
It is assigned to Star6 branch because y sreange reason Cons failed there in Root6
Sure it is bug in my version of Cons. It is not fixed yet


StRoot/St_QA_Maker/StQAMakerBase.cxx
```diff
@@ -593,6 +593,7 @@ void StQAMakerBase::BookHistFMS(){
                        title.c_str(),
                        kNChannels, 0., kNChannels,  // Channel axis bins
                        200, 0., kNAdc);             // ADC axis bins
+//    h->SetBit(TH1::kCanRebin);
```
genevb, line 675: Why add a commented line?

fgeurts (approved):
I have no objections to these updates. Implementations look fine.
A couple of comments:
- I am surprised to see the update to StVpdSimConfig.h but that update -from a self-consistency p.o.v.- makes sense.
- As far as I can tell StTofHitMaker has never been used in any data production. It has been replaced by StBTofHitMaker. In fact, StTofHitMaker will bail out if any Run9+ data is presented to it. Perhaps we should archive it?

genevb:
On Sep 1, 2021, at 12:30 PM, Frank Geurts ***@***.******@***.***>> wrote:

  *   As far as I can tell StTofHitMaker has never been used in any data production. It has been replaced by StBTofHitMaker. In fact, StTofHitMaker will bail out if any Run9+ data is presented to it. Perhaps we should archive it?

Thanks for the note, Frank. We do run some nightly tests with some old datasets, but they generally use data production chains, so if you're statement is correct, then those nightly tests should be unaffected.

-Gene

plexoos:
> * As far as I can tell StTofHitMaker has never been used in any data production. It has been replaced by StBTofHitMaker. In fact, StTofHitMaker will bail out if any Run9+ data is presented to it. Perhaps we should archive it?

If I interpret `StBFChain/BigFullChain.h` correctly, the following options depend on StTofHitMaker:

    pp2008, P2008 -> B2008a -> ToFx -> tofXDat -> StTofHitMaker

Then on this page https://www.star.bnl.gov/devcgi/dbProdOptionRetrv.pl I see a few entries using the high level options pp2008, P2008, or B2008a which probably tells us that we can't simply remove StTofHitMaker

PR #130, opened 2021-08-30
Star6

perevbnlgov:
In StMessage.h ostrstream replaced tp std::St_ostrstreamm, see Stsstreamm.h comments.
In Stsstream.h ostrstream was replaced by std::St_ostrstream to avoyd clash with real ç
and #define std::St_ostrstream ostrstream, to avoid clash with the real ostrstreamm

New class which support Cling macro call. In Root5 when Cint called macro from macro
these maccro are called independently. When submacro is defined but not called even if
such submacro does not exist or could not be called Cint is working smoothly.
In Cling, all the submacros are compiled even if this submacro is not really called.
In result your macro does not work if you have any problem in subbmacro, which is not even called.
To solve this preoblem using gROOT->ProcessLine(...) is proposed.
It is easy to do if there is no parameters. But parameters, especially with pointers is much
complicated. You must convert this pointers to the addresses.In result this process line
becames extremely long and unreadable. Example:

In Cint you need:
.x subbmacro(name1,name2);
TString Line(".x submacro(";

In Cling:
TString ts(".x subacro(");
ts += TString::ULLtoa((ULong64_t)name1,16); ts +=",";
ts += TString::ULLtoa((ULong64_t)name2,16); ts +=",";
ts+=")"
gROOT->ProcessLine("subMacro(char*)0x1234567,char*)0x1234567, ....")

Using method St::call it becames mor easy:
St::call(".x subMacro",name,name2)

Issue #21, opened 2021-06-09
PicoDst causing slowness in reconstruction chains


plexoos:
It's been reported (Gene, Lidia) that PicoDst code changes between SL21a and SL21b have caused reconstruction chains to become increasingly slower over the processing of hundreds of events. To reproduce the problem one can use the following command:

    root4star -b -q -l 'bfc.C(1000,"DbV20210417 P2019a StiCA -beamline3D btof BEmcChkStat CorrY -OPr13 evout PicoVtxVpd PicoCovMtxWrite -hitfilt","/star/data03/daq/2019/087/20087007/st_physics_20087007_raw_2000001.daq")'

PR #134, opened 2021-09-02
[root6] Rename s/ostrstream/St_ostrstream/

### LINKED ISSUE #126: Error due to ambiguous ostrstream while running bfc.C in ROOT6
When I try to run one of our tests in ROOT6 environment I get the following error:

```
Processing bfc.C(10, "pp2017,StiCA,btof,mtd,pp2pp,fmsDat,fmsPoint,fpsDat,BEmcChkStat,QAalltrigs,CorrX,OSpaceZ2,OGridLeak3D,-hitfilt", "/star-test-data/quick/st_physics_18069061_raw_2000021.daq")...
In file included from input_line_37:1:
In file included from /star-sw/StRoot/macros/bfc.C:19:
In file included from /star-sw/.sl88_gcc789/include/StBFChain.h:22:
In file included from /star-sw/.sl88_gcc789/include/StChain.h:22:
In file included from /star-sw/.sl88_gcc789/include/StMaker.h:26:
/star-sw/.sl88_gcc789/include/StMessMgr.h:108:26: error: reference to 'ostrstream' is ambiguous
class StMessMgr : public ostrstream {
                         ^
/star-sw/.sl88_gcc789/include/Stsstream.h:15:7: note: candidate found by name lookup is 'ostrstream'
class ostrstream : public std::ostringstream {
      ^
/usr/lib/gcc/x86_64-redhat-linux/4.8.5/../../../../include/c++/4.8.5/backward/strstream:143:9: note: candidate found by name lookup is 'std::ostrstream'
  class ostrstream : public basic_ostream<char>
```

Any suggestions how to fix this? @perevbnlgov @klendathu2k

plexoos:
Fixes #126 Error due to ambiguous ostrstream while running bfc.C in ROOT6

### DIFF
--- a/StRoot/StDbBroker/DbUse.cxx
+++ b/StRoot/StDbBroker/DbUse.cxx
@@ -100,7 +100,7 @@ unsigned int num_rows;
 //unsigned int num_struct;
 
 //const int MAXBUF=1024;
-ostrstream Query;
+St_ostrstream Query;
 char temps[128];
 
 char validFrom[20];
--- a/StRoot/StHbtMaker/Base/StHbtRoot1DCF.cc
+++ b/StRoot/StHbtMaker/Base/StHbtRoot1DCF.cc
@@ -105,7 +105,7 @@ void StHbtRoot1DCF::Finish(){
 
 //____________________________
 StHbtString StHbtRoot1DCF::Report(){
-  ostrstream tStr; 
+  St_ostrstream tStr; 
   tStr << GetName() << " Correlation Function Report:"<< endl;
   tStr << "Number of entries in numerator: " << mNumerator->GetEntries() << endl;;
   tStr << "Number of entries in denominator: " << mDenominator->GetEntries() << endl;
--- a/StRoot/StHbtMaker/Base/StHbtRoot2DCF.cc
+++ b/StRoot/StHbtMaker/Base/StHbtRoot2DCF.cc
@@ -107,7 +107,7 @@ void StHbtRoot2DCF::Finish(){
 
 //____________________________
 StHbtString StHbtRoot2DCF::Report(){
-  ostrstream tStr; 
+  St_ostrstream tStr; 
   tStr << GetName() << " Correlation Function Report:"<< endl;
   tStr << "Number of entries in numerator: " << mNumerator->GetEntries() << endl;;
   tStr << "Number of entries in denominator: " << mDenominator->GetEntries() << endl;
--- a/StRoot/StHbtMaker/Base/StHbtRoot3DCF.cc
+++ b/StRoot/StHbtMaker/Base/StHbtRoot3DCF.cc
@@ -108,7 +108,7 @@ void StHbtRoot3DCF::Finish(){
 
 //____________________________
 StHbtString StHbtRoot3DCF::Report(){
-  ostrstream tStr; 
+  St_ostrstream tStr; 
   tStr << GetName() << " Correlation Function Report:"<< endl;
   tStr << "Number of entries in numerator: " << mNumerator->GetEntries() << endl;;
   tStr << "Number of entries in denominator: " << mDenominator->GetEntries() << endl;
--- a/StRoot/StHbtMaker/Base/StHbtThPair.cc
+++ b/StRoot/StHbtMaker/Base/StHbtThPair.cc
@@ -86,7 +86,7 @@ void StHbtThPair::UpdateWeight() {
     mWeightNum=(mWeightNum-1.)*mPairPurity+1.;
     mWeightDen=mWeight->GetWeightDen();
     if(mWeightNum<=0. || mWeightNum>1000.){
-      ostrstream tCom;
+      St_ostrstream tCom;
       tCom << "echo ---> " << mWeightNum << " " 
 	   << mEmPoint1 << " " << mEmPoint2 << " " 
 	   << mMomentum1 << " " << mMomentum2 << " >> Err.txt" << ends;
@@ -101,7 +101,7 @@ void StHbtThPair::UpdateWeight() {
 }
 
 StHbtString StHbtThPair::Report() {
-  ostrstream tStr; 
+  St_ostrstream tStr; 
   tStr << "Default StHbtThPair Report" << endl;
   if (mWeight) {
     tStr << mWeight->Report() << endl;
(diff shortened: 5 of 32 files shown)


genevb (commented):
I have no objection to replacing ostrstream with St_ostrstream in the logger / message manager. What people are doing with ostrstream outside of that, I don't know.

dmarkh:
Personally, I never understood a need for St_* streams. Victor explained that they needed during times when STL was unstable and they encountered issues with streams - thus wrapper was created. In my code (StDbLib), I never expect St_* streams, but rather standard std:: streams. Moreover, StDbLib cannot depend on ROOT, as it has a standalone compile option - used by Jeff in the online domain. If St_* streams are ROOT-dependent, please do not use them as a substitute for std:: streams in the StDbLib codes.

perevbnlgov:
I do not remember all the history, but as a result
we have class
class ostrstream : public std::ostringstream {
But there is an official std::ostrstream.
In very old code there were methods :

const char *str()
int   pcount() const
void freeze(bool)

which are not in ostrstream.
So this fake ostrstream was used.
But when Root6 + new compiler is used this compiler found
that there are two ostrstring classes, one in Star code and one in std::
and clashes arised.

Now I replaced my class ostrstring to St_ostrstring
and added #define ostrstream St_ostrstream
No clashes any more

plexoos:
> Now I replaced my class ostrstring to St_ostrstring and added #define ostrstream St_ostrstream
> No clashes any more

First, I think you meant to say that you did s/ostrstream/St_ostrstream/ rather than s/ostrstring/St_ostrstring/
This is exactly what this PR is proposing to do across the entire code base.

Then, why do you think the `#define ostrstream St_ostrstream` is necessary?

Finally, if you look at the result of the CI checks (which you can access from this PR page) you see that the code with the above modification compiles fine but the test jobs in ROOT5 fail with the following message:

```
Error: operator<< not defined for St_ostrstream /star-sw/StRoot/macros/bfc.C:144:
Error: << Illegal operator for pointer 3 /star-sw/StRoot/macros/bfc.C:144:
```

So, clearly just the renaming of ostrstream to St_ostrstream is not enough to take care of this problem.

plexoos:
I don't know if Jason is back from his vacation but he wanted to try something different https://github.com/star-bnl/star-sw/issues/126#issuecomment-907532000

I think the idea was to replace the STAR's ostream with `std::ostringstream`. If @klendathu2k can come up with a nice PR for this we can see if it works.

klendathu2k:
Hi Dmitri,

I am back from vacation.  I'll assemble a PR this weekend.

Jason

perevbnlgov:
> s/ostrstream/St_ostrstream/ rather than s/ostrstring/St_ostrstring/
No, every there in the Star code is used ostrstream

> This is exactly what this PR is proposing to do across the entire code
> base.
    I do not want to change the entire code. In addition ostringstream 
does not have
    method const char* str(). It has std::string str(). So you must 
change also all using of str()
    I do not think that at the end of STAR we need to change a lot of 
code. (obout 100+ lines)

> Then, why do you think the #define ostrstream St_ostrstream is
> necessary?
As I explained to avoid clash between fake ostrstream and the real one.
Everywhere where is declared #include Sstrstream this #define will 
convert it into
St_ostrstream and everywhere where there is no such include and hence no 
#define,
the official std::ostrstream will be used. As a result no clash.

> Error: operator<< not defined for St_ostrstream
> /star-sw/StRoot/macros/bfc.C:144:
> Error: << Illegal operator for pointer 3
> /star-sw/StRoot/macros/bfc.C:144:
No this is fixed

Victor

plexoos:
I believe we still don't have a working solution for #126 among the existing PRs

perevbnlgov:
Do you mean that my solution does not work?
If so, please show me where I can see it.

plexoos:
> Do you mean that my solution does not work? If so, please show me where I can see it.

Did you create a pull request with your solution?

perevbnlgov:
In my Star6 origin there is the class
class St_ostrstream : public std::ostringstream {

But in the main Star6 repository
class ostrstream : public std::ostringstream {

In my comments I have:
In StMessage.h ostrstream replaced tp std::St_ostrstreamm, see 
Stsstreamm.h comments.
In Stsstream.h ostrstream was replaced by std::St_ostrstream to avoyd 
clash with real ç
and #define std::St_ostrstream ostrstream, to avoid clash with the real 
ostrstreamm

I did all the pull requests and I do not know why my updates were not 
propagated.
But this github is so complicated that I can nor be sure in anything

Victor

plexoos:
>  I do not know why my updates were not propagated.

We cannot merge changes if they cause errors in our builds. You did create a few PRs but none of them passed the test build successfully. For example, see errors here https://github.com/star-bnl/star-sw/pull/130/checks?check_run_id=3477975211

klendathu2k:
Hi Victor,

Bottom of the pull request page...
https://github.com/star-bnl/star-sw/pull/134
... it doesn't pass tests...

Digging into the test, it looks like it fails when a bfc.C macro is 
loaded
under ROOT5.  That would break all production chains...

Cheers,
Jason

perevbnlgov:
Looking on compilation I see the problem with StShadowMaker and 
StarGeoetry.
No one of them was modified by me. my compilation is successful.

Victor

plexoos:
Victor, Are you looking at this line? https://github.com/star-bnl/star-sw/pull/130/checks?check_run_id=3477975211#step:3:15375

`
#21 ERROR: process "/bin/bash -c source /etc/profile  && export Vc_DIR=/opt/software/linux-scientific7-x86_64/gcc-4.8.5/vc_-0.7.4-gqbhzu2x5u2dbe5tafxnl36xj5qbwov4  && cd /star-sw  && cons +StarVMC/Geometry  && cons %OnlTools          %StRoot/StShadowMaker  && find /star-sw/.$STAR_HOST_SYS -name *.o -exec rm '{}' \\;" did not complete successfully: exit code: 2
`

You need to look at the lines prior to this one.

The problem is not with either `StarVMC/Geometry` or `StShadowMaker` which is excluded form the build

plexoos:
I am not sure what you mean by "your" compilation. The CI jobs build the entire code base (with the exception of `StShadowMaker`)

Actually, you probably want to comment about the compilation problems in your PR #130
This PR #134 has a problem with ostrstream in run time

perevbnlgov:
On 2021-09-15 13:09, Dmitri Smirnov wrote:
> I am not sure what you mean by "your" compilation. The CI jobs build
> the entire code base (with the exception of StShadowMaker)
It is very simple. I have full code in the local repository and it
compiled without errors

perevbnlgov:
I found that the problem in StHbtMaker. It is out of my compilation. It 
is no used anymore.
Last comment was 7 years ago.

plexoos:
Superseded by #150

PR #136, opened 2021-09-03
Remove StShadowMaker implementation

genevb:
The existing implementation can be restored again later if needed.

Removing it allows compilation of libraries without explicitly excluding the directory for it.


plexoos:
Is there a good reason to keep the header file without implementation?

genevb:
> Is there a good reason to keep the header file without implementation?

Removing just the implementation is, in my opinion, the simplest approach without altering any other codes.

plexoos:
Gene, the missing implementation will cause jobs to crash during run time when another code calls the functions defined in StShadowMaker. I see that it can be called from StRoot/StBFChain/StBFChain.cxx and StRoot/StTagsMaker/StTagsMaker.cxx. Removing the implementation but keeping the BFC option ("shadow") is not a very graceful solution.

If for some reason you don't want to expose the original implementation, it may be better to remove the option and all calls to StShadowMaker all together. If it is done in a single commit it will be easy to inspect it and/or restore if needed in the future. It is also possible to add some dummy implementation in order to avoid crashes. But in this case I am also not sure if it is really worth keeping such unuseful code.

genevb:
It sounds like you know quite clearly what you want to do, and there's no understood need for me to maintain my responsibility. I'll close this.

plexoos:
So, are you OK with removing all references to StShadowMaker?

PR #146, opened 2021-09-14
Star buildgeom fix

klendathu2k:
Propose fix to issue #45.   When building the (.h, .cxx) files from the .xml sources, the dependence on the agmlParser.py is no longer required.  This may fix issue #45 as well.  When I remove the dependency, I can reliably build the geometry in a freshly cloned directory.

### DIFF
--- a/docker/Dockerfile.root5
+++ b/docker/Dockerfile.root5
@@ -122,7 +122,6 @@ diff --git a/asps/rexe/Conscript b/asps/rexe/Conscript \n\
 
 RUN source /etc/profile \
  && cd /star-sw \
- && cons +StarVMC/Geometry \
  && cons %StRoot/StShadowMaker \
  && find /star-sw/.$STAR_HOST_SYS -name *.o -exec rm '{}' \;
 
--- a/docker/Dockerfile.root6
+++ b/docker/Dockerfile.root6
@@ -125,7 +125,6 @@ RUN source /etc/profile \
  # Override default Vc version
  && export Vc_DIR=/opt/software/linux-scientific7-x86_64/gcc-4.8.5/vc_-0.7.4-gqbhzu2x5u2dbe5tafxnl36xj5qbwov4 \
  && cd /star-sw \
- && cons +StarVMC/Geometry \
  && cons %StRoot/StShadowMaker \
  && find /star-sw/.$STAR_HOST_SYS -name *.o -exec rm '{}' \;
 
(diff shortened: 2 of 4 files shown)


genevb (commented):
To be sure I understand, you're saying this helps prevent the incomplete compile, but doesn't remove the need to first compile `StarVMC/Geometry`, right?

klendathu2k:
Yes to the first.  Maybe to the second -- I did not test to see if a 
single issue of cons
fully builds the geometry.  Will try this shortly.

klendathu2k:
This still requires cons +StarVMC/Geometry before the full build.

Printing out the dependencies reveals a big hint about why the build 
fails
out of the box.

.$STAR_HOST_SYS/obj/StarVMC/StarGeometry/BbcmGeo.o: 
.$STAR_HOST_SYS/obj/StarVMC/StarGeometry/BbcmGeo.cxx

The object file only depends on the implementation file for the geometry 
modules, not the
header file.  But otherwise, the dependency tree looks correct.  (.h and 
.cxx both depend on XML, and
the "InstallAs" targets look correct as well.)

It looks like the cons "logic" for discovering the dependencies is a 
little fragile here.  When I
generate the implementation files, I simply do an
#include "BbcmGeo.h"

It looks like cons happy with
#include "StarVMC/StarGeometry/BbcmGeo.h"

After I confirm this compiles, I may be pushing a complete solution.  
TBD.

Jason

klendathu2k:
The latter commit should remove the need for a separate pass with cons +StarVMC/Geometry.

klendathu2k:
NOTE:  this requires starting from a _clean_ StarVMC/StarGeometry/ directory.

genevb:
I confirm that starting with a clean StarVMC/StarGeometry, and not first executing `cons StarVMC/Geometry`, then executing only `cons`, I get...
...without these patches, nearly empty StarVMC/xgeometry and StarVMC/StarGeometry directories
...with these patches, numerous .age files in StarVMC/xgeometry and numerous .h and .cxx files in StarVMC/StarGeometry.
Is there anything else I should check?

plexoos:
I removed `cons +StarVMC/Geometry` from the CI build to see if all libraries can be built with a single call to cons.

> Removing the dependency on the parser allows a freshly cloned repo to compile,
and may resolve issue #45.  It is not clear to me why.

This is a bit worrisome because naturally one would expect the build system to re-generate the source code from xml files when `agmlParser.py` is changed

plexoos:
Superseded by #148

PR #150, opened 2021-09-17
Use std::ostringstream in place of STAR's ostrstream

### LINKED ISSUE #126: Error due to ambiguous ostrstream while running bfc.C in ROOT6

plexoos:
Another attempt to resolve #126

To summarize, we get rid of `ostrstream` as implemented in ` StRoot/St_base/Stsstream.h` and switch to using `std::ostringstream` class instead. This transition is more or less straightforward as the STAR's implementation of `ostrstream` inherited from `std::ostringstream`.

### DIFF
--- a/StRoot/StDbBroker/DbUse.cxx
+++ b/StRoot/StDbBroker/DbUse.cxx
@@ -176,7 +176,7 @@ uint num_latest;
 int latestDirDate;
 int latestDirTime;
 
-if (mysql_real_query(&mysql,Query.str(),Query.pcount()-1))
+if (mysql_real_query(&mysql, Query.str().c_str(), int(Query.tellp())-1))
   {
     LOG_ERROR << "Failed to query: Error: " <<  mysql_error(&mysql) << endm;
     mysql_close(&mysql);
@@ -252,7 +252,7 @@ int latestStrID=99999;
 uint sizeOfDbStruct=99999;
 uint nDbVar=99999;
 
-if (mysql_real_query(&mysql,Query.str(),Query.pcount()-1))
+if (mysql_real_query(&mysql, Query.str().c_str(), int(Query.tellp())-1))
   {
     LOG_ERROR << "Failed to query: Error: " <<  mysql_error(&mysql) << endm;
      mysql_close(&mysql);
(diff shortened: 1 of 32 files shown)


perevbnlgov:
Dmitri, I see you ignore my one line correction and introduce your 
hundred ones.
I can not work with my code, when you modify my code without permission.
I see you have a plenty of free time. Then it will be better to take 
your responsibility
for Star Root6 development. And I will promise you not to intervene in 
your development.
Victor

veprbl:
@perevbnlgov 
I believe you were pointed out that your change breaks StHbtMaker. It also introduces extra definitions to the `std::` namespace, which can be argued to be suboptimal.

dmarkh (approved):
LGTM

plexoos:
Hi Victor, we really need to move forward with ROOT6. This task has been pending for too long already, at least a year I think. Unfortunately, the changes you submitted so far could not be compiled for various reasons, but having a compileable code is a prerequisite to start looking into it seriously.

Hopefully, this PR addresses the problem with the deprecated `ostrstream` and does not introduce any side effects, but still it is just one more step towards the ROOT6 goal. I am sure there are still many problems to solve. Your contributions will be highly appreciated.

veprbl:
I also want to add that contributions to the development are not solely about modifying source code files. It also includes the discussion and all proposed variants are a part of it.

veprbl (approved):
~~This is pretty much equivalent to Victor's change, but is more explicit about its effect.~~ The StMess* classes expose string streams in its API, but it's likely that nobody ever had a use of those return values anyway.

veprbl:
Cross linking to Mattermost: https://chat.sdcc.bnl.gov/star/pl/j7j415iuhty85bpo99397fujoy

PR #153, opened 2021-09-22
Remove StRoot/StShadowMaker and "shadow" option

plexoos:
This code cannot be used so I propose to remove it. Alternatively, the authors could fix the implementation by providing the valid code in [StRoot/StShadowMaker](https://github.com/star-bnl/star-sw/blob/68ae73a95f62da13742401879f26f5646727954b/StRoot/StShadowMaker/StShadowMaker.cxx) but my understanding from #136 is that they are fine with just removing it.


genevb (commented):
My understanding is that this is no longer my responsibility.

plexoos:
As you wish. You can remove the corresponding entry from the CODEOWNERS file in order to avoid automatic notifications
https://github.com/star-bnl/star-sw/blob/68ae73a95f62da13742401879f26f5646727954b/.github/CODEOWNERS#L84

klendathu2k:
That is not quite my reading of the discussion in #136.

StShadowMaker was used in the blinding/unblinding of data for the isobar data.  That is the reason that the implementation file is encrypted.  This code is (I believe) needed in order to be able to read / process the isobar data, and thus we need to maintain this in our library releases.  Now that the data has been unblinded, it is not clear to me that StShadowMaker should remain encrypted.  But that decision should be taken up as a topic for an S&C meeting.

plexoos:
Just like I said:

> Alternatively, the authors could fix the implementation by providing the valid code in StRoot/StShadowMaker

Who is the author? Jerome? @jlauret

jlauret:
This code does not need to be deployed (it is not part of the chain except the blind analysis). We can leave it encrypted and never use in a checkout. Only the header can / should be deployed.
Furthermore, I no longer have responsibility for the fate of any code in STAR remember? So, while the suggestion is already to leave it "as-is" and  not checkout upon new library release (only the header file is needed, not the code itself), @genevb  should IMO be the new owner of this piece of code.

genevb:
A couple things that appear to need clarifying:
- StShadowMaker is not needed to read the unblinded data, which is what is now in the hands of analyzers. It is only necessary for blinding purposes.
- Jerome & I were jointly involved in preparing the StShadowMaker code and encrypting it. If there is a directive from the Collaboration (particularly the Analysis Blinding Committee) to provide it unencrypted, this is of course possible.

EDIT: I see Jerome responded at about the same time I did, so there is some overlap in what we have said.

plexoos:
>  We can leave it encrypted and never use in a checkout.

Why? It cannot be compiled, it cannot be used.
Why do you want to complicate the life of our users by requiring a sub-standard clone procedure? I hope you do realize that the production team is not the only user who can clone and compile the entire repo?

klendathu2k:
Alternative to removing it:  modify cons to ignore by default.

jlauret:
> Alternative to removing it: modify cons to ignore by default.
>

Or rename .cxx.enc and move on. So many possibilities ...

plexoos:
> modify cons to ignore by default.

Exactly my proposal #69 since July 26

genevb:
> As you wish. You can remove the corresponding entry from the CODEOWNERS file in order to avoid automatic notifications
> 
> https://github.com/star-bnl/star-sw/blob/68ae73a95f62da13742401879f26f5646727954b/.github/CODEOWNERS#L84

My understanding != my wish
And apparently CODEOWNERS != responsibility for maintaining code (this point was made in #69 )

plexoos:
> My understanding != my wish

It's up to you

> And apparently CODEOWNERS != responsibility for maintaining code (this point was made in #69 )

Your point != policy adopted by the collaboration. If there is a document describing the procedure of who and how can modify the code then please let us know.

PR #159, opened 2021-09-30
Fix float->int for setId, adding access funciton for neighbor

akioogawa:
Fixing a bug float->int for setId function's argument
Adding access function neighbor


klendathu2k (approved):
Change to interface / underlying data structure remains the same.

veprbl (approved):

plexoos (approved):
LGTM

ullrich-bnl (approved):

PR #162, opened 2021-10-04
TPC SpaceCharge & GridLeak calibration: a few improvements based on lessons learned from Run 20 AuAu9.2

genevb:
A few improvements based on lessons learned from calibrating Run 20 AuAu9.2:
- Do not execute MINOS error calculation for parameters who are at their limits from the standard fit as this was allowing the
 limits to be exceeded
- PCA : check on sanity of g5 from iteration 0 before using in iteration 1
- FIX_GL : 3D fit for GridLeak now used only to provide g2 in this case

### DIFF
--- a/StRoot/macros/calib/Calib_SC_GL.C
+++ b/StRoot/macros/calib/Calib_SC_GL.C
@@ -165,6 +165,8 @@ int  Init(const char* input);
 int  Waiting();
 void SetMinMax(int n, Double_t* ar, double& min, double& max, double inclusion, double margin=0.075);
 int  SetMinuitPars(const int n, TString* names, Double_t* starts, Double_t* steps, int debug, Bool_t* fixing=0);
+void GetMinuitPar(const int n);
+void GetMinuitPars(const int n, Bool_t* fixing=0, Double_t* fixedValues=0);
 void Log2Lin(int i);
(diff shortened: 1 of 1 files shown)


plexoos:
Is there a mechanism to match the calibration values inserted into the database with the code revision used to produced them?

genevb:
Hi, Dmitri

On Oct 4, 2021, at 7:56 PM, Dmitri Smirnov ***@***.******@***.***>> wrote:

Is there a mechanism to match the calibration values inserted into the database with the code revision used to produced them?

Just the time it was entered (the database entries will be uploaded once this code is merged into the main branch). The latest version of the code is always used at the time of upload, and we have both of those times.

-Gene

genevb:
I'd like to see this get merged soon. Does it fall to the Infrastructure Team to approve codes that no one else really cares much about?

-Gene

plexoos (approved):
Yes, let's merge it.

Just a friendly reminder that you probably want to run it with ROOT6 if not yet already. The sooner we learn about posible issues with rootcling the better.

genevb:
> Just a friendly reminder that you probably want to run it with ROOT6 if not yet already. The sooner we learn about posible issues with rootcling the better.

Good point - thanks for the reminder. I'll try to address any such incompatibilities in a separate PR.

PR #164, opened 2021-10-06
Getter/setter for user weights

klendathu2k:
Adding a "get" and "set" method for the user weights defined in the event generator record.  These were previously missing, and will be required for upcoming hard diffraction simulations.

### DIFF
--- a/StRoot/StarGenerator/EVENT/StarGenEvent.h
+++ b/StRoot/StarGenerator/EVENT/StarGenEvent.h
@@ -205,6 +205,12 @@ class StarGenEvent : public TObject
   /// Returns the filter result
   UInt_t GetFilterResult(){ return mFilterResult; }
 
+  /// Attach a user-weight to the event
+  void AddUserWeight( double w ){ mWeights.push_back(w); }
+
+  /// Retrieve the user weights vector
+  const std::vector<double>& GetUserWeights(){ return mWeights; }
+
  private:
   /// Copy ctor is private.  
   StarGenEvent( const StarGenEvent &other );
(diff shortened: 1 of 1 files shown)


StRoot/StarGenerator/EVENT/StarGenEvent.h
```diff
@@ -205,6 +205,12 @@ class StarGenEvent : public TObject
   /// Returns the filter result
   UInt_t GetFilterResult(){ return mFilterResult; }
 
+  /// Attach a user-weight to the event
+  void AddUserWeight( double w ){ mWeights.push_back(w); }
+
+  /// Retrieve the user weights vector
+  const std::vector<double>& GetUserWeights(){ return mWeights; }
```
plexoos:
```suggestion
  const std::vector<double>& GetUserWeights() const { return mWeights; }
```
If you want to be a c++ purist 🐕 `

plexoos (approved):

plexoos:
I guess waiting for Victor's approval on this unambiguous addition is futile...

StRoot/StarGenerator/EVENT/StarGenEvent.h
```diff
@@ -205,6 +205,14 @@ class StarGenEvent : public TObject
   /// Returns the filter result
   UInt_t GetFilterResult(){ return mFilterResult; }
 
+  /// Attach a user-weight to the event
+  void AddUserWeight( double w ){ mWeights.push_back(w); }
+
+  /// Retrieve the user weights vector
+  const std::vector<double>& GetUserWeights()       { return mWeights; }
```
plexoos, line 212: Ah shoot, just noticed after merging. This getter signature does not make much sense and should be removed or rewritten to return a non-const reference

perevbnlgov:
Well, if such setter/getter is needed, no objection. The only remmark, it could be slightly safer to make the getter const

PR #169, opened 2021-10-11
Initial Implement of FST Online Monitor

sunxuhit:
The FST Online Monitor code (fstBuilder) is based on istBuilder with some updates due to different geometry.
The code is tested on my local RCF account with daq file took by Run Control.

### DIFF
--- a/OnlTools/Jevp/StJevpBuilders/fstBuilder.h
+++ b/OnlTools/Jevp/StJevpBuilders/fstBuilder.h
@@ -2,7 +2,7 @@
 #include <stdlib.h>
 
 #include "JevpBuilder.h"
-#include "DAQ_READER/daqReader.h"
+class daqReader;
 #include <TH1F.h>
 #include <TH2F.h>
 #include <TH1S.h>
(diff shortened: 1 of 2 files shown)


sunxuhit:
Dear Expert,

Not sure what is this CI Build/ROOT5 failure means here. 
I could compiled the fstBuilder at my rcf account with ROOT 5.34 without any error.
Could you please point me where is the problem?

Xu

OnlTools/Jevp/StJevpBuilders/fstBuilder.h
```diff
@@ -0,0 +1,651 @@
+#include <stdio.h>
+#include <stdlib.h>
+
+#include "JevpBuilder.h"
+#include "DAQ_READER/daqReader.h"
```
plexoos: Try to forward declare instead

```suggestion
class daqReader;
```

plexoos:
Something is wrong with the header file as rootcint chokes on it for some reason. I would suspect the unions but you say it compiles in a different environment...

jml985 (approved):
I've tested the changes and they produce appropriate plots, the resource requirements are not out of bounds, and the architecture seems correct.

PR #174, opened 2021-10-16
Update code in StRoot/StLaserAnalysisMaker

fisyak:


plexoos (approved):
>  kehw requested review from kehw and plexoos

This PR changes files only in StRoot/StLaserAnalysisMaker. I do not expected it to affect any result from reco jobs directly.

kehw (approved):

PR #175, opened 2021-10-20
D edx fxt2019

fisyak:


genevb:
Just a comment on the commits to StarDb to make sure everyone understands:

The point of using the offline database over files in StarDb is to enable simple run-time selection from the history of parameters via chain options, without the need to use local files to make that selection. It additionally allows for parameters to evolve even after a library is frozen.
- A good example of this is how we are right now easily able to continue the Run 14 AuAu14.5 GeV production in SL21c with updated ETOF calibrations without the need to modify the SL21c library.
- An example against this is how we have to update various libraries for updates to TpcRS parameters which are not stored in the database as in #107 .

I believe Yuri mentioned at today's S&C meeting that he will be putting parameters into the database, though it is for Yuri to say how that works together with these proposed commits.

iraklic:
I have a general question on the approvals of this sort. I can see that there are many new tables added for some correction/calibrations, but there are also changes of the source codes, most of them are benign and I can understand and approve but there are quite a few which in order to understand I need to spend some time and work with Yuri and so on. So is my approval here something completely symbolic or what is the approach here? If one is to double check codes owned by someone else one need to work on that. Or does the approval mean that "it looks reasonable enough" for me, and the CI tests will take care of the compilation/build, and the physics level changes are left to the developer? <- is this the case (I suspect it it, because others don't make sense to me).

starsdong:
Irakli, I agree with your elaboration. Most of reviews are from the structure, coding standard point of view. For all subsystems, we rely on the developers to test/demonstrate the physics performance, they are the experts in these directions.

iraklic:
> Irakli, I agree with your elaboration. Most of reviews are from the structure, coding standard point of view. For all subsystems, we rely on the developers to test/demonstrate the physics performance, they are the experts in these directions.

Hi Xin. Thanks. This sounds good to me and I kind of agree with this approach. I will review these to the standards that I always adhered to. Just one advice to all: smaller and more frequent commits/PRs are preferred, at least to me and I think in general. They take less time to review and have quicker turnaround as well as are easier to trace back in case of issues later on.

iraklic (approved):

iraklic:
I approved this PR. It took me some time to dig where `memDelay` in `LanaTrees.C` was used to but as I understand this is not a calibration for lasers (I wanted to make sure the same parameters are not use for all the previous runs), but it is just Yuri's analysis macro. @fisyak correct me if I am wrong.

dmarkh (approved):
LGTM (I only reviewed IDLs). There is a failing CI check though.

starsdong:
Yuri and All,

I see the PR didn't pass the CI build. Can you take a look and fix it? I see Dmitry and Irakli have approved. Should be able to get merged once it passes the CI build automatically, to my understanding.

fisyak:
Hi Xin, do you have instruction how I can reproduce "CI build" ?

starsdong:
Hi Yuri,

I will have to ask Dmitri to comment on this. But I see in the summary, there is one failing check. You may click the "Details" to find out the details.

fisyak:
Hi Xin, I have found a problem, fixed it, and I am waiting for check results.

PR #180, opened 2021-10-31
Add fix for fixed Target run with broken yellowIntensity information

fisyak:


dmarkh (approved):
LGTM

plexoos:
Unfortunately, the CI tests wouldn't be able to detect problems from such changes. I hope someone is looking at the nightly test results and will warn us if there are issues.

starsdong:
Dmitri, shall we merge this request in? I believe the nightly test and later QA will need to check these carefully.

plexoos:
Are you fine with the change and approve it, Xin? If so, we can merge unless there are real opposing solutions. (Not sure if #161 can qualify as a solution yet...) I personally don't have an opinion based on the information provided here and will abstain.

Regarding the nightly tests, last time I checked Birdview it was password protected and I could not easily find what I was looking for... but if you can follow up on this that would be great and good enough for me.

starsdong:
Hi Dmitri,

I think we need both this one (I believe Yuri uses this in the TPC related code) and PR#161 (still needed for others). Gene, is it right?

Thanks

/xin

genevb:
> On Nov 1, 2021, at 1:29 PM, Xin Dong ***@***.***> wrote:
> I think we need both this one (I believe Yuri uses this in the TPC related code) and PR#161 (still needed for others). Gene, is it right?
> 
Hi, Xin

Other codes depend on the solution proposed in PR 161.

-Gene

nigmatkulov (approved):

PR #191, opened 2021-11-12
fstBuilder with two input ped files

sunxuhit:
Updated fstBuilder to read in two pedestals files.

### DIFF
--- a/OnlTools/Jevp/StJevpBuilders/fstBuilder.h
+++ b/OnlTools/Jevp/StJevpBuilders/fstBuilder.h
@@ -2,7 +2,8 @@
 #include <stdlib.h>
 
 #include "JevpBuilder.h"
-#include "DAQ_READER/daqReader.h"
+// #include "DAQ_READER/daqReader.h"
+class daqReader;
 #include <TH1F.h>
 #include <TH2F.h>
 #include <TH1S.h>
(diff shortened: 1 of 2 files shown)


sunxuhit:
I could compile and run on my own RCF account. Not sure why failed the test.

OnlTools/Jevp/StJevpBuilders/fstBuilder.h
```diff
@@ -2,10 +2,7 @@
 #include <stdlib.h>
 
 #include "JevpBuilder.h"
-
-class daqReader;
-
-
+#include "DAQ_READER/daqReader.h"
```
plexoos: This is the reason why it can't compile in the CI environment. I suggest you keep the forward declaration for daqReader

sunxuhit: Thanks, I remembered the solution now....

sunxuhit: Just tried the forward declaration for daqReader on RCF but the compile failed.

plexoos: If the build on the farm failed we would see it [here](https://www.star.bnl.gov/public/comp/prod/Sanity/AutoBuild.html) or [here](https://www.star.bnl.gov/public/comp/prod/Sanity/AutoBuild-64bits.html)

PR #205, opened 2021-11-21
Add BUILD_CMD arg to Dockerfile

plexoos:
The users may optionally specify their own build command when building containers.
This may be used to build older releases where some packages need to be excluded from the build, e.g.:

```
docker build --build-arg BUILD_CMD="cons %OnlTools" ...
```

or when a two-stage build is required, e.g.:

```
docker build --build-arg BUILD_CMD="cons +StarVMC/Geometry && cons %ShadowMaker" ...
```

### DIFF
--- a/docker/Dockerfile.root6
+++ b/docker/Dockerfile.root6
@@ -101,6 +101,8 @@ RUN ln -s $MYSQL/include /usr/include/mysql
 
 FROM base-stage AS build-stage
 
+ARG BUILD_CMD="cons"
+
 COPY . /star-sw
 
 # XXX TEMP XXX: Don't force mysql static libraries to be linked into root4star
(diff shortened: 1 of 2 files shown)


nigmatkulov (approved):
Looks okay.

docker/Dockerfile.root6
```diff
@@ -125,7 +127,8 @@ RUN source /etc/profile \
  # Override default Vc version
  && export Vc_DIR=/opt/software/linux-scientific7-x86_64/gcc-4.8.5/vc_-0.7.4-gqbhzu2x5u2dbe5tafxnl36xj5qbwov4 \
  && cd /star-sw \
- && cons \
+ && install -m 755 <(echo "${BUILD_CMD}") build_cmd.sh \
+ && ./build_cmd.sh \
```
veprbl:
```suggestion
 && echo "${BUILD_CMD}" | bash -xe \
```

veprbl (commented):
The fact that we need to implement arbitrary code execution in the docker file probably means that it's not designed right. Would make more sense if the "build-stage" was ready to run cons immediately.

plexoos:
I agree but right now I don't see how to implement in the base stage what used to be a two-stage build with cons, e.g. "cons +StarVMC/Geometry && cons %StSkipPackage"

veprbl:
If you need an override, you write a separate dockerfile and import a stage of the existing one?

PR #206, opened 2021-11-22
Sl20c embedding patch

klendathu2k:
Draft pull request to identify codes which need to be updated in SL21c.


klendathu2k:
This would not be appropriate b/c The embedding patch was based on SL20c, so too many conflicts.

PR #209, opened 2021-11-22
StEvent additions for sTGC (Ftt) detector

jdbrice:
These additions add StEvent structures for:
- StFttRawHit
- StFttCluster
- StFttPoint
- StFttCollection

The second commit also adds the StFttRawHitMaker and a BFC option for running it.

A bare minimum chain for testing is:
root4star -b -q -l 'bfc.C( 5, "in, db, StEvent, FttDat", "input.daq" )'

Note: I have daq files that can be used for testing at:
/star/data03/pwg/jdb/FWD/daq/stgc-cosmic/


starsdong:
Hi Daniel, for the StFttRawHitMaker, this is a new piece of code. Shall we request reviews from a couple of peers for this new one?

StRoot/StFttRawHitMaker/StFttRawHitMaker.h
```diff
@@ -0,0 +1,63 @@
...
+ ***************************************************************************/
+#ifndef STFTTRAWHITMAKER_H
+#define STFTTRAWHITMAKER_H
+
+#include "StRTSBaseMaker.h"
+
+#include <vector>
+
+#ifndef __CINT__
+// copied from RTS ... trying to include directly fails
```
veprbl: What is the error message?

plexoos: It is likely due to the inclusion of "sys/types.h" for u_char and u_short. I saw rootcint choking on it in https://github.com/star-bnl/star-sw/pull/101#issuecomment-899818930

veprbl: I've never heard of u_char or u_short. The ISO C standard defines `<stdint.h>` that comes with (u)int?_t, I wonder if that would be more portable.

jdbrice: I think it should be fine to change those types. However, this is a direct copy from RTS code.

jdbrice:
Hi Xin, yes. For the StFttRawHitMaker I agree. I submitted together since they require the StEvent additions to run

StRoot/StFttRawHitMaker/StFttRawHitMaker.h
veprbl: I will submit some PRs to address the types change. However, can we, please, update the comment to mention which header was supposed to be included and what is the error message that prevents us from using it?

akioogawa (approved):

StRoot/StFttRawHitMaker/StFttRawHitMaker.h
jdbrice: I may be wrong but I dont think the RTS/include/... are setup in such a way that they can be included from elsewhere. They use internally include paths that are not supported by our standard set of include paths. We had the same issue in all of the past detectors that I remember working on. If there is a solution to that I am happy to update it here

klendathu2k:
Hi Daniel,

Is there a reason why Monte Carlo truth information is not being propagated to clusters and points?  Is the plan to calculate this directly from the Monte Carlo hits and (if so) how does one reach the MC hits used in a given cluster / point / track?

Jason

jdbrice:
Hi Jason,

That's a good idea. I can add this. What do I need, just an int for McId? Let me know and I'll add that. 

Best,
Daniel

starsdong:
Hi Hongwei and/or Jeff,

This PR includes a new StFttRawHitMaker as the offline DAQ decoder.  May I ask either or both of you to help review the StFttRawHitMaker? Thank you

/xin

StRoot/StFttRawHitMaker/StFttRawHitMaker.h
veprbl: Works for me
```diff
diff --git a/StRoot/StFttRawHitMaker/StFttRawHitMaker.h b/StRoot/StFttRawHitMaker/StFttRawHitMaker.h
index 9eb42367d7..dcb6df41ea 100644
--- a/StRoot/StFttRawHitMaker/StFttRawHitMaker.h
+++ b/StRoot/StFttRawHitMaker/StFttRawHitMaker.h
@@ -15,18 +15,10 @@
 
 #include <vector>
 
-#ifndef __CINT__
-// copied from RTS ... trying to include directly fails
-struct stgc_vmm_t {
-    u_char feb_vmm ;    
-    u_char ch ;
-    u_short adc ;
-    u_short bcid ;
-    short tb ;  // from the trigger
-} ;
-#endif
 #ifdef __CINT__
 struct stgc_vmm_t;
+#else
+#include <RTS/src/DAQ_STGC/daq_stgc.h>
 #endif
 
 class StEvent;
```

jdbrice: Ah great. 
I must have remembered wrong. I'll update this 👍

kehw:
> Hi Hongwei and/or Jeff,
> 
> This PR includes a new StFttRawHitMaker as the offline DAQ decoder. May I ask either or both of you to help review the StFttRawHitMaker? Thank you
> 
> /xin

Sure. I will look into this.

StRoot/StBFChain/BigFullChain.h
```diff
@@ -1698,7 +1698,8 @@ Bfc_st BFC[] = { // standard chains
    "StFcsClusterMaker","StFcsClusterMaker","Fill FCS clusters",                                     kFALSE},
   {"fcsPoint"   ,"","", "StEvent,fcsDb",
    "StFcsPointMaker","StFcsPointMaker,libMinuit","Fill FCS points",                                 kFALSE},
-
+   {"FttDat","","","StEvent","StFttRawHitMaker","StFttRawHitMaker,StEvent",
```
genevb, line 1701: I think the 4th element should probably be "event", ensuring that StEvent exists before StFttRawHitMaker runs in the chain.

jdbrice: Actually using StEvent was intentional. event or MakeEvent also add detDb which I was avoiding to keep the minimal chain simple and fast

genevb: The `"StEvent"` option also turns on `detDb`:
```
{"StEvent"   ,"","","globT,SCL,TRGDef,StBichsel,EmcUtil,TbUtil,detDb","","StEvent","Load StEvent",kFALSE},
```
Perhaps it didn't do so sometime in the past, but now it does. Regardless, `"Event"` ensures that there will be an StEvent for you to fill, not just that StEvent libraries are loaded/available at run-time.

jdbrice: Ok I see, I am fine with that then. Thanks

jml985 (approved):
The code looks fine to me and is needed.

plexoos:
@jdbrice Should we wait for more feedback or changes from you?

StRoot/StEvent/StFttPoint.cxx
```diff
@@ -0,0 +1,20 @@
...
+ * Description: Implementation of StFttPoint, the StEvent point structure
+ *
+ ***************************************************************************/
+#include "StFttPoint.h"
+#include "St_base/StMessMgr.h"
+
+
+StFttPoint::StFttPoint() :  StObject() { /* no op */ }
+
+StFttPoint::~StFttPoint() { /* no op */ }
```
kehw, line 17: A minor comment on the performance side. If the destructor does nothing, there is no need to define an empty one. If we do, we may prevent the compiler to generate the move and move assignment constructors under C++11. In that case, we lost some optimization opportunities.

kehw (approved):
The new data structures look straightforward and decoding part is simple.

StRoot/StBFChain/BigFullChain.h
jdbrice: Hi Gene,

I just had a chance to test this. In fact I find that the test chain that I provided:
```
root4star -b -q -l 'bfc.C( 5, "in, db, StEvent, FttDat", "input.daq" )'
```
runs fine if FttDat uses "StEvent" as the 4th item. It also saves the data into an StEvent object, no problem.  
However, if I change it to "Event" or "MakeEvent" then the inclusion of "StTpcDb" leads to an error

```
root4star: .sl73_gcc485/obj/StRoot/StTpcDb/StTpcDb.cxx:276: float StTpcDb::DriftVelocity(Int_t): Assertion `mUc > 0' failed.
Abort
```

Is there a simple way to solve this? Even if this is not an issue for production chains, I still need a minimal chain that I can run for development purposes

jdbrice:
@plexoos this last commit implemented the comments from @kehw (remove empty dtor) and from @veprbl (directly include daq_stgc.h instead of defining structure myself)

I have not taken the comment from @genevb wrt the BFC chain, see above comment. We can always update that as needed later.

Once the checks pass please merge the code in ASAP. We have a high-priority FastOffline request that cannot run until this is included

starsdong:
Thank you, Daniel. I would agree that we should go with your initial proposal on the chain option.

Thomas, if you get a chance, I would appreciate that you can take a review of the StEvent container structure. Because of the urgency to provide feedback to online operation, we really would like to get this PR merged ASAP. If you cannot do it today, I would suggest we should merge this PR in now and we should still have a chance to modify them later. Thanks

PR #211, opened 2021-11-23
StRoot/Stv*: replace sys/types.h with stdint.h

veprbl:
ROOT5 doesn't always work with sys/types.h, types from stdint are the ISO C standard


akioogawa (approved):

perevbnlgov (approved):
I see that uint was replaced to unsigned int. I do not see the difference.
Especially it is not important, be couse this code is frozen. It is not used now.
But anyway there is no any danger to change it
Victor

PR #215, opened 2021-11-29
starsim: make seed overflow into a fatal error

veprbl:
The value of bigger than 0x7FFFFFFF will overflow ISEQ and make it
wrap around into a negative values. The current behavior is to print a
warning message and reset the seed to a default value. Because of that
an unsuspecting user can easily initialize a fraction of their
simulation production with the same seed.

### DIFF
--- a/asps/Simulation/starsim/atutil/aranlux.F
+++ b/asps/Simulation/starsim/atutil/aranlux.F
@@ -256,8 +256,11 @@ C                    Entry to initialize from one or three integers
 #endif
       ENDIF
       IN24 = 0
-      IF (INS .LT. 0)  WRITE (6,'(A)')
+      IF (INS .LT. 0)  THEN
+        WRITE (6,'(A)')
      +   ' Illegal initialization by RLUXGO, negative input seed'
+        CALL ABORT
+      ENDIF
       IF (INS .GT. 0)  THEN
         JSEED = INS
 #ifndef SILENT_RANLUX


plexoos:
Will it abort before the simulation starts?

veprbl:
> Will it abort before the simulation starts?

Yes

plexoos (approved):

plexoos:
Any comments, Victor? @perevbnlgov

klendathu2k (approved):
Is there a trouble ticket / github issue associated with this PR?   It would be nice to know how this error mode is encountered.  Typical production usage sets the RNG seed equal to the $JOBINDEX when running on the batch system...  integers O(1000).  Or by hand in some cases.  

Regardless, this is a reasonable change.

veprbl:
@klendathu2k I don't think there is any officially maintained simulation/embedding production scripts in this repo. If one rolls their own, it's not unreasonable to imagine that they would supply reasonable big number seeds.

klendathu2k:
Hi Dmitry,

Users do not roll their own production scripts / macros.  Simulation and
embedding production campaigns are handled by the simulation and 
embedding
teams.  Users may contribute code, but it is vetted by the relevant 
experts
before large data samples are produced.

The embedding team maintains its macros under StRoot/macros.  Ranlux 
does not appear
to be used in the embedding workflow (default GRNDM is used).

Simulation macros are generally created on an as-needed basis.  I have 
these archived (but
not in the repo... yet).

Jason

PR #217, opened 2021-11-29
Initial Run 22 chains, including FTT and FCS

genevb:
This turns on chain options for FCS (which have existed for a little while) and FTT (just added today: #209 ).

Example test job:
```
root4star -b -q -l 'bfc.C(10,"pp2022 StiCA BEmcChkStat","/star/data03/daq/2021/326/22326042/st_cosmic_adc_22326042_raw_0000005.daq")'
```
...currently crashing in filling MuDst with FCS information on the second event. @akioogawa and @jdbrice are aware and there is an effort to resolve it. But that should not be an issue of chain options, so I think the addition of the chain options can proceed regardless.

-Gene


plexoos (approved):
ok

genevb:
I understand @klendathu2k is unlikely to be available for helping review/approve this week. So I requested @starsdong as a second reviewer, but I'm not too picky about who reviews this.

starsdong (approved):
Looks good. Thanks

PR #222, opened 2021-12-02
StMuDstMaker, init mFcsCollection to zero

jdbrice:
Fix small bug, mFcsCollection not initialialized in second ctor, prevents actually reading the fcs data since mFcsCollection doesnt get set properly.


akioogawa (approved):
This resolved an Issue of MuDst reading for FCS.

plexoos (approved):

PR #226, opened 2021-12-03
Adding option to read from MuDst in StFcsRawHitMaker

akioogawa:
Adding option to read from MuDst in StFcsRawHitMaker, and example macro to read from MuDst, re-create StFcsHits in StEvent on memory, and re-run FCS reconstruction makers.

### DIFF
--- a/StRoot/StFcsRawHitMaker/StFcsRawHitMaker.h
+++ b/StRoot/StFcsRawHitMaker/StFcsRawHitMaker.h
@@ -25,7 +25,8 @@ public:
 
     void setReadMode(int v) {mReadMode=v;}
     void setDebug(int v=1) {SetDebug(v);}  //!backward compatubility     
-    
+    void setReadMuDst(int v=1) {mReadMuDst=v;} //!reading Mudst/StMuFcsHit into StEvent/StFcsHit
+
     // Get CVS
     virtual const char *GetCVS() const;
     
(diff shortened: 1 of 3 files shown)


StRoot/StFcsRawHitMaker/StFcsRawHitMaker.h
```diff
@@ -25,7 +25,8 @@ class StFcsRawHitMaker : public StRTSBaseMaker {
 
     void setReadMode(int v) {mReadMode=v;}
     void setDebug(int v=1) {SetDebug(v);}  //!backward compatubility     
-    
+    void setReadMuDst(int v=1) {mReadMuDst=v;} //!reading Mudst/StMuFcsHit into StEvent/StFcsHit
```
genevb, line 28: Not a big deal, but why assign an int to an unsigned int?

genevb: Also, not that you have to make a change, but all makers inherit a mode data member from StMaker. It's pretty harmless to add your own mode data member as you've done, but you can certainly make use of what's already there if you like:
```
> grep -i mode $STAR/StRoot/StChain/StMaker.h
   Int_t           m_Mode;              //!Integer mode of maker
   virtual void         SetMode(Int_t mode=0)   {m_Mode=mode;}   // *MENU*
   virtual Int_t        GetMode() { return m_Mode;}
```
And you can get fancy with `enum` if you want 0 and 1 to represent "normalMode" and "readMode" or something like that. Just throwing it out there.

veprbl (approved):
Makes sense

plexoos (approved):

jdbrice (approved):

PR #231, opened 2021-12-07
[JevpProduction] Update fstBuilder for beam

sunxuhit:
Switch the signal cuts back to default for the beam.

### DIFF
--- a/OnlTools/Jevp/StJevpBuilders/fstBuilder.cxx
+++ b/OnlTools/Jevp/StJevpBuilders/fstBuilder.cxx
@@ -38,8 +38,8 @@ const float fstBuilder::maxTbFracOK       = 0.9;
 const float fstBuilder::landauFit_dn      = 10.0;
 const float fstBuilder::landauFit_up      = 2000.0;
 const float fstBuilder::cmnCut            = 3.0;
-const float fstBuilder::hitCut            = 8.0;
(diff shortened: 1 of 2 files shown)


starsdong:
Any concern from the reviewers on this request? The requested changes seem to straightforward.
We do also have one issue with the CI Build/ROOT5 failing, any comment?

veprbl:
The failure on the "JevpProduction" branch is a known issue and needs to be addressed separately from this PR.

jml985:
@starsdong The issue we are having is that changes to the offline (I guess) have stopped the StRoot/RTS directory from compiling because of data type support being removed.   There was a pull request from Dmitry that fixes this for offline, but would break the experiments DAQ if back ported our code.        I'm on vacation (and still supporting the run, but not this nonsense) till january so there will be nothing done on my end till then.       On the plus side, it isn't crucial whether or not the Jevp pull requests are accepted as they are not used in the offline code anywhere.   As far as the running QA at the experiment, the PRs are incorporated and tested in the online as soon as they are submitted whether or not they pass the CI tests.   This is ugly, but even if the CI tests were working properly the review process is not quick enough for the QA updates during the run.

starsdong:
Closing this PR. Jevp is update-to-date with the recent PR by Jeff.
https://github.com/star-bnl/star-sw/pull/292

Issue #213, opened 2021-11-24
Replacing sys/types.h with stdint.h


veprbl:
This is a tracking issue to replace non-standard types with types defined in the ISO C and C++ standards. Specifically, the following:

 - u_char, u_short, u_int -> uint8_t, uint16_t, uint32_t
 - ushort, uint, ulong -> unsigned short, unsigned int, unsigned long

The ultimate goal is to get rid of unnecessary includes of sys/types.h. This will help with ROOT5 being incompatible with the newest versions of libc (see https://github.com/NixOS/nixpkgs/blob/master/pkgs/applications/science/misc/root/purify_include_paths_root5.patch for an alternative solution).

 - [x] StRoot/Stv* (#211)
 - [x] StRoot/StTriggerUtilities (#212)
 - [x] StRoot/StSpinPool (#224)
 - [ ] StRoot/RTS (#225)
 - [ ] OnlTools (#225)
 - [ ] StRoot/St{,E}Emc*
 - [ ] StRoot/StDb*
 - [ ] StRoot/StDaqLib, StRoot/StDAQMaker
 - [x] StRoot/StGenericVertexMaker
 - [x] #234
 - [x] Disable _BSD_SOURCE?

plexoos:
I don't think we have any _BSD_SOURCE in the code

veprbl:
I don't remember what I was thinking, but it is probably controlled via other settings implemented in libc's. In any case, this can't be done unless we get rid of using that functionality first, which I don't think is going to happen at this point (we are blocked on having Jeff do the work on VxWorks bits).

plexoos:
#443 should take care of the `StRoot/StGenericVertexMaker` bullet

fgeurts:
Hi @veprbl and @plexoos,
Is this issue still tracking the non-standard type replacements? It hasn't been updated in a couple of years. Should we leave it open, assign to others, or has this been addressed and we should close it. Please advise.

veprbl:
We weren't able to complete this migration fully. This is not a strict requirement for ROOT6-based builds. Let's close.

PR #212, opened 2021-11-23
StRoot/StTriggerUtilities: replace sys/types.h with stdint.h

veprbl:
ROOT5/cint doesn't always work with sys/types.h, types from stdint are the ISO C standard

PR #234, opened 2021-12-11
misc: replace sys/types.h with stdint.h

veprbl:
For motivation see:
- #213
- https://github.com/star-bnl/star-sw/pull/212#discussion_r761571455

### DIFF
--- a/StRoot/StTofHitMaker/StTofHitMaker.cxx
+++ b/StRoot/StTofHitMaker/StTofHitMaker.cxx
@@ -146,8 +146,8 @@ Int_t StTofHitMaker::Make()
    LOG_INFO << " getting the tof collection " << tofCollection << endm;
    if (tofCollection) {
       if ( GetNextRaw() ) {
-         // u_int ddl[4][10000] ;	/// content of up to 4 fibers
-     	   // u_int ddl_words[4] ;	/// the count of words (32bit) for above
+         // uint32_t ddl[4][10000] ;	/// content of up to 4 fibers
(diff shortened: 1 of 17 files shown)


StRoot/StTofHitMaker/StTofHitMaker.cxx
```diff
@@ -146,8 +146,8 @@ Int_t StTofHitMaker::Make()
    LOG_INFO << " getting the tof collection " << tofCollection << endm;
    if (tofCollection) {
       if ( GetNextRaw() ) {
-         // u_int ddl[4][10000] ;	/// content of up to 4 fibers
-     	   // u_int ddl_words[4] ;	/// the count of words (32bit) for above
+         // uint32_t ddl[4][10000] ;	/// content of up to 4 fibers
```
klendathu2k, line 149: Guessing this is done with a refactoring tool... which explains why comments are left in place (which I'm fine with).  Gives a little more confidence that the types remain invariant under the refactoring.

klendathu2k (approved):
I don't see any issues with this.  Approved.

fgeurts (approved):

plexoos (approved):

StRoot/StTofHitMaker/StTofHitMaker.cxx
veprbl: Sorry to disappoint, it's just a tab character. My refactoring tool is `find . -type f -exec sed -i {} -e 's/u_int/uint32_t/g' \;` and `git add -p`. However, `clang-rename` can handle this too, but would require some extra setup.

PR #208, opened 2021-11-22
[SL21c] Add Anti-H4 Lambda decay to Anti-He4 + pion- 

klendathu2k:
Update to SL21c codes for Run19 AuAu19 embedding.    [PR#176 did the same for SL20c.]

* Add H3LambdaBar decay to gstar library

* Add H4LambdaBar, He4LambdaBar & He5LambdaBar hypernuclei to StarClassLibrary.
Add only H4LambdaBar to the particle table (others are placeholders assuming
a request is made at some point).

(cherry picked from commit c109e4f41158956e72663d5fd00988cd6603e1b0)

PR #239, opened 2021-12-13
Fix the charge of He5Lambda

plexoos:
Motivated by [this fix](https://github.com/star-bnl/star-sw/pull/208/commits/24c2071bae1d88bce84af59463d446b4aed889db) on SL21c. See #208


klendathu2k (approved):
Looks good.

veprbl (approved):
This is a forward port of 24c2071bae1d88bce84af59463d446b4aed889db 

Helium5 is an isotope of the second periodic table element hence charge +2, the Lambda0 baryon is neutral, hence total electric charge +2.

Also if I understood correctly, first argument of hid in `encoding=hid(2,4,1),` is the charge, so makes sense here as well.

PR #241, opened 2021-12-13
Modifications to StBTofCalibMaker for FXT

bkimel:
A number of small changes to StBTofCalibMaker for running in FXT mode: select vertex within target, use iterative outlier rejection developed by Bassam, added specific cases to outlier rejection for nT0==1 or 2. These changes were made to both the functions that run over StEvent and MuDst so that there is consistency. Additional changes were made to StBFChain to add "btofFXT" as a chain option (it was not present and there was no way to turn on FXT mode in StBTofCalibMaker) as well as have the "FXT" chain option automatically use this . Checks were performed and confirmed that all changes behaved exactly as expected.


starsdong (approved):
The changes look good to me.

Reviewers, would appreciate if you can review/approve the PR at your earliest convenience. We need this to update the SL21d for Run19 FXT data production. Thanks

genevb (approved):
I'm ok with the changes to StRoot/StBFChain/.

fgeurts (approved):
Proposed update is ok. The zeroing of NT0 (in the case of only 2 tracks that are too far apart in time) kind of breaks the expected behavior in other modes, but since this is isolated to FXTmode only this is fine

PR #225, opened 2021-12-03
[JevpProduction] OnlTools: replace sys/types.h with stdint.h

veprbl:
For motivation see:
- #213
- https://github.com/star-bnl/star-sw/pull/212#discussion_r761571455

PR #251, opened 2021-12-14
Updates to Jevp for 2022 run

jml985:
This is the current running Jevp code online, after modifications to make it run stable.   Most changes to code in Jevp/StJevpBuilders is already part of other specific pull requests from the detector owners.   I don't know how, or whether it is important to separate those out.


veprbl:
@jml985 This breaks ROOT5 build in CI. I was looking to fix that in #225, or we could find some other workaround, but this better be solved if we are to merge this.

jml985:
Yeah, I saw that but I don't understand it.   I use root5 on evp where this runs:   All the code changed is in the Jevp tree.  A straight:

stardev
git init
git config core.sparseCheckout true
git remote add upstream https://github.com/star-bnl/star-sw.git
echo "OnlTools/Jevp" > .git/info/sparse-checkout
echo "OnlTools/PDFUtil" > .git/info/sparse-checkout
git pull upstream pull/251/head
cons

Works fine.    The logfile showing the error is not comprehensible to me.   It appears that the problem is in rtsSystems.h?   But not at all clear why, and in any event it wasn't changed in this pull request.

veprbl:
Seems to be a small mismatch between our containers running Scientific Linux 7.9 and the Scientific Linux 7.3 that you may be running. Here is a reproducer:
```
cd star-sw
singularity build star-sw-root5.sif docker://ghcr.io/star-bnl/star-sw-root5-base:main_latest
singularity exec -e -B $PWD:/star-sw star-sw-root5.sif /bin/bash -lc 'cd /star-sw && cons +asps +pams +OnlTools +StDb +StRoot/StarMagField'
```

plexoos:
I just would like to add that trying to sort out the differences between the environments in the container and online nodes does not seem like a practical task. Since we know there is a fix (#255) that works for the both environments it would make more sense to apply it and make our code more "robust" and portable so, it can work not just in SL7.3 on online nodes. I want to remind that the proposed fix replaces the **older** integral types such as `u_int`, `u_long`, `u_char` (never standardized) with the **equivalent** types define by the language standard `uint32_t`, `uint8_t` etc.

jml985:
We aren't talking about sorting out the difference between two linux environments.   These are the actual header files from the DAQ code which runs everywhere from linux to vxworks to specialized VHDL implemented mini-C environments.    These are provided as a service so that offline can use the readers we build.    We don't have stdint.h on all these platforms and even if we did we wouldn't change and test all of our source code to match the header changes because you don't like sys/types.h.    We can easily make a stub defining the types removed in the latest version of __CINT__ or whatever but you can't just change the data types that have been in use for 20 years because you don't like them.

PR #252, opened 2021-12-14
StEvent StFcsHit adding forgotten protection against bad inputs

akioogawa:
Adding forgotten protection against bad inputs to StEvent StFcsHit.

### DIFF
--- a/StRoot/StEvent/StFcsHit.cxx
+++ b/StRoot/StEvent/StFcsHit.cxx
@@ -73,7 +73,7 @@ int   StFcsHit::nPeak()    const {return mNPeak;}
 float StFcsHit::energy()   const {return mEnergy;}
 
 void StFcsHit::setDepCh(unsigned short ns, unsigned short ehp, unsigned short dep, unsigned short ch) { 
-    mDepCh = (ns << 15) | (ehp<<13) | (dep<<8) | ch; 
+    mDepCh = ((ns&0x1) << 15) | ((ehp&0x3)<<13) | ((dep&0x1f)<<8) | (ch & 0xff);
 }
 void StFcsHit::setNS(unsigned short val)       { setDepCh(val,ehp(),dep(),channel());}
 void StFcsHit::setEHP(unsigned short val)      { setDepCh(ns(),val,dep(),channel());}


StRoot/StEvent/StFcsHit.cxx
```diff
@@ -73,7 +73,7 @@ int   StFcsHit::nPeak()    const {return mNPeak;}
 float StFcsHit::energy()   const {return mEnergy;}
 
 void StFcsHit::setDepCh(unsigned short ns, unsigned short ehp, unsigned short dep, unsigned short ch) { 
-    mDepCh = (ns << 15) | (ehp<<13) | (dep<<8) | ch; 
+    mDepCh = ((ns&0x1) << 15) | ((ehp&0x3)<<13) | ((dep&0x1f)<<8) | (ch & 0xff);
```
veprbl, line 76:
```suggestion
    assert((ns & (~0x1)) == 0);
    assert((ehp & (~0x3)) == 0);
    assert((dep & (~0x1f)) == 0);
    assert((ch & (~0xff)) == 0);
    mDepCh = ((ns&0x1) << 15) | ((ehp&0x3)<<13) | ((dep&0x1f)<<8) | (ch & 0xff);
```

klendathu2k (approved):
I don't see any issues here.

I will comment that it would be nice to see doxygen comments specifying what the inputs are, and what the expected values of them are.    But not going to hold this up on account of that.

StRoot/StEvent/StFcsHit.cxx
akioogawa: I thought using assert was strongly discouraged...

akioogawa:
StFcsHit.h has
    UShort_t mDetId=0;        // 1 bit ZS, 3 bits DetectorId, 12 bits id                                                              
    UShort_t mDepCh=0;        // 1 bit for NS, 2 bits for EHP, 5 bits for DEP, 8 bits for channal

klendathu2k:
If you are writing corrupt data into StEvent... that should not be a 
reason to kill
the rest of the chain.  So yes, assert is not the correct thing here.

I assume that the bit masks you are adding are intended to mask off 
noisy bits?  If so,
then this is a reasonable solution.  If not... if nonzero bits indicate 
data corruption...
then I would think that zeroing out the event / issuing a warning would 
be correct.

Cheers,
Jason

StRoot/StEvent/StFcsHit.cxx
veprbl: Why?

akioogawa: Stpping the whole chain and delaying QA process for all other detectors is worse than one detector getting a bugged data?

veprbl: Turns out we don't define the `NDEBUG` for release builds, so this was not a good suggestion.

Crashing the chain is better than redoing the production.

veprbl (approved):

akioogawa:
Other side is controlled by Tonko. He used ch=32~36 for DEP data which is a 32ch device. Luckly I used 8bit(255 max) for channel and it is safe. So far I have not seen anything corrupting this mDepCh. StFacRawHitMaker has a trick to put all "non-standard" hits into none-existing detectorId=6 array, which includes intermediate "trigger data" from trigger algorithms and other possibly useful info. So even if there is crazy data which can corrupt mDepCh, it should not corrupt real raw data from detector.

PR #256, opened 2021-12-16
StFmsBsQaMaker: fix invaild ClassDef

veprbl:
With ROOT6:

    error: static assertion failed: ClassDef(Inline) macro: the specified class version number is not an integer.


plexoos (approved):
Thanks!

akioogawa (approved):

PR #257, opened 2021-12-16
StMiniMcMaker: use const reference to TString in setters

veprbl:
QA :INFO  - ProcessLine ((StMiniMcMaker *) 0x101ebfc0)->setFileName("rcf12002_1_100evts.root");
    input_line_617:2:46: error: non-const lvalue reference to type 'TString' cannot bind to a value of unrelated type 'const char [24]'
     ((StMiniMcMaker *) 0x101ebfc0)->setFileName("rcf12002_1_100evts.root");
                                                 ^~~~~~~~~~~~~~~~~~~~~~~~~
    .sl73_x8664_gcc485/obj/StRoot/StMiniMcMaker/StMiniMcMaker.h:265:30: note: passing argument to parameter 'val' here
      void  setFileName(TString& val)               { mInFileName = val; }
                                 ^


plexoos (approved):
Makes perfect sense. Thanks!

starsdong (approved):

PR #263, opened 2021-12-20
add four zdcsmd adc plots to Jevp

kehw:
Per Aihong's requests, I added four 2D plots for ZDMSMD ADC plots to trgBuilder. It is also implemented Akio's suggestion to the 1D ADC plots.


kehw:
Jeff has accepted the changes and put it in the run. I am closing this PR.

PR #264, opened 2021-12-20
docker: install the same version of pyparsing as on SDCC

veprbl:
This is not urgent, but our dockerfiles seem to be broken right now as they install an incompatible version (3.x) of pyparsing needed for mgr/agmlParser.py

### DIFF
--- a/docker/Dockerfile.root5
+++ b/docker/Dockerfile.root5
@@ -69,7 +69,7 @@ RUN yum update -q -y \
  && yum clean all
 
 # Install extra python modules used by the STAR software
-RUN pip install pyparsing
+RUN pip install pyparsing==2.2.0
 
 ENV MODULEPATH=/opt/linux-scientific7-x86_64
 
--- a/docker/Dockerfile.root6
+++ b/docker/Dockerfile.root6
@@ -69,7 +69,7 @@ RUN yum update -q -y \
  && yum clean all
 
 # Install extra python modules used by the STAR software
-RUN pip install pyparsing
+RUN pip install pyparsing==2.2.0
 
 ENV MODULEPATH=/opt/linux-scientific7-x86_64
 


plexoos:
How do you see that it is broken?

veprbl:
You could try rebuilding the container. Here is my attempt with CentOS 7.3, but it should be the same on SL 7.9:
https://github.com/veprbl/star-sw/runs/4569985951
```
#16 [base-stage 7/9] RUN pip install pyparsing
#16 0.154 /bin/bash: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8)
#16 0.294 sh: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8)
#16 0.299 sh: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8)
#16 0.374 sh: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8)
#16 0.387 Collecting pyparsing
#16 0.766   Downloading https://files.pythonhosted.org/packages/ab/61/1a1613e3dcca483a7aa9d446cb4614e6425eb853b90db131c305bd9674cb/pyparsing-3.0.6.tar.gz (882kB)
#16 1.021     Complete output from command python setup.py egg_info:
#16 1.022     Traceback (most recent call last):
#16 1.022       File "<string>", line 1, in <module>
#16 1.022       File "/tmp/pip-build-Mnetrn/pyparsing/setup.py", line 8, in <module>
#16 1.022         from pyparsing import __version__ as pyparsing_version
#16 1.022       File "pyparsing/__init__.py", line 100
#16 1.022         major: int
#16 1.022              ^
#16 1.022     SyntaxError: invalid syntax
#16 1.022     
#16 1.022     ----------------------------------------
#16 1.029 Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-build-Mnetrn/pyparsing/
#16 1.128 You are using pip version 8.1.2, however version 21.3.1 is available.
#16 1.128 You should consider upgrading via the 'pip install --upgrade pip' command.
#16 ERROR: process "/bin/bash -c pip install pyparsing" did not complete successfully: exit code: 1
```

klendathu2k:
First, a little history...

The pyparsing package was used to implement the ageParser.py code, which reads in the old AgSTAR geometry files (a.k.a. the ".g" files, kept under pams/geometry).  This is the major part of the code which uses pyparsing.   If it were the only place, then I think we could safely retire the pyparsing.py code from usage.  However, there are minor parts of the syntax (dealing with matching agstar internal variables) which are still present in the AgML parsing framework.   So we need to maintain a version of pyparsing in the software stack.

However, I don;t think it matters much which version is used.  I expect things haven't changed enough to impact the places where it is needed.

veprbl:
> However, I don;t think it matters much which version is used. I expect things haven't changed enough to impact the places where it is needed.

Can your code run under python3? If not, then it matters.

veprbl:
Also pinning a specific version improves reproducibility.

klendathu2k:
I have not tested python 3.  When I get a chance, I will try running it.

plexoos (approved):

PR #273, opened 2022-01-05
StBTofCalibMaker: fix variable shadowing

veprbl:
The code in question looks like this:

```c++
    int doPID = 0;
    for(;;) {
       // ...
       Bool_t doPID = kFALSE;
       // ...
       if(!doPID) continue;
       doPID++;
    }
    LOG_INFO << doPID << endm;
```

The doPID variable from the outer scope gets shadowed by the boolean flag variable. That should make it always report 0.

This was found with a following compiler diagnostic:

    .sl79_gcc485/obj/StRoot/StBTofCalibMaker/StBTofCalibMaker.cxx: In member function 'void StBTofCalibMaker::processStEvent()':
    .sl79_gcc485/obj/StRoot/StBTofCalibMaker/StBTofCalibMaker.cxx:1241:14: error: use of an operand of type 'bool' in 'operator++' is forbidden in C++17
     1241 |         doPID++;
          |              ^~
    .sl79_gcc485/obj/StRoot/StBTofCalibMaker/StBTofCalibMaker.cxx: In member function 'void StBTofCalibMaker::processMuDst()':
    .sl79_gcc485/obj/StRoot/StBTofCalibMaker/StBTofCalibMaker.cxx:1496:18: error: use of an operand of type 'bool' in 'operator++' is forbidden in C++17
     1496 |             doPID++;
          |                  ^~

### DIFF
--- a/StRoot/StBTofCalibMaker/StBTofCalibMaker.cxx
+++ b/StRoot/StBTofCalibMaker/StBTofCalibMaker.cxx
@@ -1129,7 +1129,7 @@ void StBTofCalibMaker::processStEvent()
     // keep a few counters:
     int nohitfound(0), ismchit(0),trayoutofbounds(0),notrack(0),
         gnopidtraits(0),gtraitisfalse(0), pidtoffalse(0),calibfailed(0),
-        pgnopidtraits(0), ptraitisfalse(0),tofisset(0), doPID(0);
+        pgnopidtraits(0), ptraitisfalse(0),tofisset(0), num_doPID(0);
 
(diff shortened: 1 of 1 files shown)


fgeurts (approved):
That's a good catch. Recent code updates introduced a counter that had the exact same name as a boolean flag that was used inside a loop. The good news is that this shadowing did not have any detrimental to the calibrations as the counter was only used for logging purposes and the boolean "reset" happened inside the loop for each individual track - as originally intended.

starsdong (approved):

PR #275, opened 2022-01-07
Resolve a memory leak with created lists

genevb:
Some additional tiny other changes, but predominantly to resolve the memory leak

### DIFF
--- a/StRoot/StAnalysisUtilities/StHistUtil.cxx
+++ b/StRoot/StAnalysisUtilities/StHistUtil.cxx
@@ -352,6 +352,7 @@
 #include "TObjString.h"
 #include "TMath.h"
 #include "TString.h"
+#include "TRegexp.h"
 #include "TPaveLabel.h"
 #include "TPaveText.h"
 #include "TLegend.h"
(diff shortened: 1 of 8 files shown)


plexoos:
If I understand correctly, a (partial) copy of a TList with histogram objects can be created by this line https://github.com/star-bnl/star-sw/pull/275/files#diff-ab754b35c990f9626b367fecb9746fb194cfb2d02b550b48a50d91bd25d6169cR1750
But according to this document https://root.cern/root/html534/TList.html#TList:_TList TList does not own the objects by default unless you call `dList2->SetOwner();` in https://github.com/star-bnl/star-sw/blob/8a817eea3e956ebb0fcb6ac192e8b55fdd347a45/StRoot/StAnalysisUtilities/StHistUtil.cxx#L1755-L1756

So, if you do take the ownership of the objects in the list and all works as expected then you should be able to just `delete dirList` and `dirList2` without the need to introduce the new boolean flags `created` and `createdR`.

genevb:
> So, if you do take the ownership of the objects in the list and all works as expected then you should be able to just `delete dirList` and `dirList2` without the need to introduce the new boolean flags `created` and `createdR`.

Thanks for the suggestion, @plexoos ....I'll look into it.
-Gene

PR #278, opened 2022-01-10
bfc.C: fix for cling

veprbl:
cling needs to have libraries loaded before running a macro.

This disables the #include statements when ACLiC is not
used. Currently, cling can't parse those headers: there seems to be an
issue with how forward declarations interract with
`$clingAutoload$`. Fortunately, the `#pragma load` takes care of the
headers, like in ROOT5, so things just work.

### DIFF
--- a/StRoot/macros/bfc.C
+++ b/StRoot/macros/bfc.C
@@ -10,6 +10,9 @@
 class StBFChain;        
 class StMessMgr;
 
+#pragma cling load("StarRoot")
+#pragma cling load("St_base")
+#pragma cling load("StChain")
 #pragma cling load("StUtilities")
 #pragma cling load("StBFChain")
 
--- a/StRoot/macros/bfc.C
+++ b/StRoot/macros/bfc.C
@@ -9,7 +9,11 @@
 //////////////////////////////////////////////////////////////////////////
 class StBFChain;        
 class StMessMgr;
-#if !defined(__CINT__) || defined(__MAKECINT__)
+
+#pragma cling load("StUtilities")
+#pragma cling load("StBFChain")
+
+#if !(defined(__CINT__) || defined(__CLING__)) || defined(__MAKECINT__)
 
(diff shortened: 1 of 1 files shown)


genevb (commented):
Hi, @veprbl . I understood @perevbnlgov was working on the transition of macros for ROOT6. Did this get discussed with him?

Thanks,
-Gene

veprbl:
> Hi, @veprbl . I understood @perevbnlgov was working on the transition of macros for ROOT6. Did this get discussed with him?
> 
> Thanks, -Gene

I am only aware of what's discussed in threads https://github.com/star-bnl/star-sw/pulls?q=author%3Aperevbnlgov+ . There was no private discussion. If discussion is needed, it can happen here, I suppose.

StRoot/macros/bfc.C
```diff
@@ -9,7 +9,11 @@
 //////////////////////////////////////////////////////////////////////////
 class StBFChain;        
 class StMessMgr;
-#if !defined(__CINT__) || defined(__MAKECINT__)
+
+#pragma cling load("StUtilities")
```
klendathu2k, line 16: So... I did a search for documentation on this feature of cling... and only was able to find the load pragma in their github repo under tests...

https://github.com/root-project/cling

Regardless,... this appears to solve the problem, enabling the necessary libraries to be preloaded.  I *think* you could also achieve this by loading libraries in the standard rootlogon.C.  But this would be more general (and allows individual users to implement their own root logons).

klendathu2k (approved):
LGTM

StRoot/macros/bfc.C
veprbl: Yes, rootlogon.C, might be possible, but, if it can work without the "secret sauce", I prefer that)

plexoos (approved):

plexoos:
Last call for comments?

Issue #155, opened 2021-09-22
Errors while runnig bfc.C jobs with ROOT6


plexoos:
In #154 I set up a branch ([pr/ci_root6_tests ](https://github.com/star-bnl/star-sw/tree/pr/ci_root6_tests)) with a few test jobs running in the ROOT6 environment. This way we can see the errors reported in the log files and try to address them

plexoos:
Numerous errors from `cling` with "Missing FileEntry for" for various header files

https://github.com/star-bnl/star-sw/pull/154/checks?check_run_id=3678994719#step:5:5

```
Error in cling::AutoloadingVisitor::InsertIntoAutoloadingState:
   Missing FileEntry for TGiant3.h
   requested to autoload type StarMC
Error in cling::AutoloadingVisitor::InsertIntoAutoloadingState:
   Missing FileEntry for TGiant3.h
   requested to autoload type Gcflag_t
Error in cling::AutoloadingVisitor::InsertIntoAutoloadingState:
   Missing FileEntry for TGiant3.h
   requested to autoload type TGiant3
```

```
Error in cling::AutoloadingVisitor::InsertIntoAutoloadingState:
   Missing FileEntry for StThreeVector.hh
   requested to autoload type StThreeVector
Error in cling::AutoloadingVisitor::InsertIntoAutoloadingState:
   Missing FileEntry for StLorentzVector.hh
   requested to autoload type StLorentzVector
Error in cling::AutoloadingVisitor::InsertIntoAutoloadingState:
   Missing FileEntry for StMatrix.hh
   requested to autoload type StMatrix
Error in cling::AutoloadingVisitor::InsertIntoAutoloadingState:
   Missing FileEntry for StHelix.hh
   requested to autoload type StHelix
```

---

Proposed solution: https://github.com/star-bnl/star-sw/pull/154/commits/3b7c72f3b4b41ae64511d389ccfb54379c345feb

plexoos:
https://github.com/star-bnl/star-sw/pull/154/checks?check_run_id=3678994719#step:5:14

```
Error in <TCling::RegisterModule>: cannot find dictionary module root4star_Cint_rdict.pcm
```

---

Proposed solution: https://github.com/star-bnl/star-sw/pull/154/commits/2b201d9a7613e39073680260df78b2c6e84960f8

PR #280, opened 2022-01-15
cling bfc.C test

veprbl:
I wonder if the CI failure is happening in the rootlogon.C macro even before bfc.C is processed. If that is the case, then there seems to be an error in CI that I don't see in my environment.

- Enable ROOT6 tests
- Disable rootlogon <- expect an error to change
- Enable fix <- why not try

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -3,29 +3,6 @@ name: CI Build
 on: [pull_request]
 
 jobs:
-  ROOT5:
-    runs-on: ubuntu-latest
-    steps:
-      - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v1
-        with:
-          driver-opts: env.BUILDKIT_STEP_LOG_MAX_SIZE=10000000
-
-      - name: Build with Docker
-        uses: docker/build-push-action@v2
-        with:
-          cache-from: type=registry,ref=ghcr.io/star-bnl/star-sw-root5-base@sha256:4bed8a8b729fa9e6c8e92b2bdecd6b73869e67a07bfbde969936f921123d4c3e
-          push: false
-          tags: ghcr.io/star-bnl/star-sw-root5-build
-          file: docker/Dockerfile.root5
-          outputs: type=docker,dest=/tmp/star-sw-root5-build.tar
-
-      - name: Save built image for test jobs
-        uses: actions/upload-artifact@v2
-        with:
-          name: star-sw-root5-build
-          path: /tmp/star-sw-root5-build.tar
-
   ROOT6:
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


veprbl:
Sorry for the noise, I meant to open this against my private fork.

This is surprising, but the test went as expected. In my tests I was using a vanilla ROOT 6.24/06 and a basic environment

<details>
```
source /path/to/root/bin/thisroot.sh
export STAR=$PWD
export USE_64BITS=1
export CERN=/cern64
export CERN_LEVEL=pro
export CERN_ROOT=$CERN/$CERN_LEVEL
export XOPTSTAR=/afs/rhic.bnl.gov/opt/star/sl73_x8664_gcc485
export OPTSTAR=/opt/star/sl73_x8664_gcc485
export STAR_SYS=x8664_sl7
export STAR_HOST_SYS=sl73_x8664_gcc485
#export NODEBUG=yes
export STAR_LIB=$STAR/.${STAR_HOST_SYS}/lib
export STAR_BIN=$STAR/.${STAR_HOST_SYS}/bin
export ROOT_INCLUDE_PATH=$STAR/StRoot:$(find $STAR/.${STAR_HOST_SYS}/include $STAR/.${STAR_HOST_SYS}/obj/StRoot -type d | xargs printf ":%s")
export PATH=$CERN_ROOT/bin:$STAR_BIN:$STAR/mgr:$PATH
export LD_LIBRARY_PATH=$STAR_LIB:$LD_LIBRARY_PATH:/opt/star/sl73_x8664_gcc485/lib
```
</details>

Just applying the #278 would get bfc.C going in my environment, as you can see above, not on the CI. One difference to CI is that I wasn't including the rootlogon.C file, and I did test locally, with it, by using `HOME=StRoot/macros root4star -b -q -l 'bfc.C` (that prints the ` *** Start at Date : `), but that worked just fine. So then came this PR where I removed the rootrc from the docker container and that partially solved the problem.

Then I'm forced to add some additional `#pragma load` for `St_base` and `StChain` as those are dependencies of the `StUtilities` and `StBFChain`:
https://github.com/star-bnl/star-sw/blob/ec4f3ffd0ab71dc66f2c64ca3b18d26ce8b71276/StRoot/macros/bfc.C#L13-L16
That is corresponding to what is done in `Load` for ROOT5:
https://github.com/star-bnl/star-sw/blob/ec4f3ffd0ab71dc66f2c64ca3b18d26ce8b71276/StRoot/macros/bfc.C#L116-L128

Now what I don't get is why my ROOT6 knows that it needs to load `St_base` and `StChain`. I think this is also the reason why it is able to execute rootlogon.C in my test, possibly by loading dependency libraries on the fly. This could be because ROOT version on CI, which is 6.16, is too old for the feature.

cc #155

PR #285, opened 2022-01-25
Misalignment updates FST and Simplified FST Geometry

klendathu2k:
Updates to misalignment tables (names, correct timestamps, etc...) for the FST.  Implementation of simplified FST geometry model for track reconstruction.  Support for misaligned geometries in g2t_volume_id.


genevb (commented):
I don't have much to speak of on any of this PR (I'm sure misalignment will be useful). A tangential question though: what else is needed for `y2022a`?

klendathu2k:
Depends on whether forward tracking is to be supported with y2022a.  If 
we limit
ourselves to support central tracking with the y2022a tag, I can define 
that tomorrow.
Otherwise, I have to double check where we are on the sTGC.

akioogawa (approved):

genevb:
> Depends on whether forward tracking is to be supported with y2022a. If we limit ourselves to support central tracking with the y2022a tag, I can define that tomorrow. 

If you define y2022a now (sans forward tracking), is there any difference from y2022 (initial)?

klendathu2k:
There should not be.  It will be a straight y2022 --> y2022a copy.  
There have
been no updates outside of the forward detectors.

genevb:
> There should not be. It will be a straight y2022 --> y2022a copy. There have been no updates outside of the forward detectors.

Thanks for the reply. In that case, I am not pressing for freezing y2022a now.

plexoos:
Should we store the misalignment values from StarDb/Geometry/fst/*.C in the database instead?

klendathu2k:
@plexoos The misalignments are not expected to vary over the course of the run, so we only expect one entry once they have been measured.   The practice for HFT misalignments, and I believe TPC and others as well, has been to store them StarDb area.

klendathu2k:
@genevb @perevbnlgov  This geometry update is stalled waiting for a code owner to approve.  Could one of you have a look?  Thanks.

StRoot/StChain/GeometryDbAliases.h
```diff
@@ -155,12 +155,15 @@ static const DbAlias_t fDbAlias[] = {// geometry  Comment            old
 
   {"y2021",       20201215,     0, "y2021",    "y2021 development geometry, AgML,xgeometry"},   
   {"y2021a",      20201215,     1, "y2021a",   "y2021 first production geometry, AgML,xgeometry"},   
-
+  
   {"y2022",       20211015,     0, "y2022",    "y2022 first cut geometry, AgML,xgeometry"},   
 
-  {"dev2021",     21201210,     1, "dev2021",  "-deprecated- geometry for 2021+ forward program,AgML,xgeometry"},  
+  {"dev2021",     21201210,     1, "dev2021",  "-deprecated- geometry for 2021+ forward program,AgML,xgeometry"},
   {"dev2022",     21211210,     1, "dev2022",  "development geometry for 2022+ forward program,AgML,xgeometry"},
```
genevb, line 162: Why not use 20211015, as is the case for y2022, for all the dev2022* timestamps?

StRoot/StChain/GeometryDbAliases.h
```diff
@@ -155,12 +155,15 @@ static const DbAlias_t fDbAlias[] = {// geometry  Comment            old
...
 
+  {"dev2022m",     20221210,     2, "dev2022m",  "development geometry for 2022+ forward program,AgML,xgeometry"},
+  {"dev2022sm",     20221210,     3, "dev2022sm",  "development geometry for 2022+ forward program,AgML,xgeometry"},
```
genevb: Can the comments here indicate the difference between `dev2022`,` dev2022m`, and `dev2022sm`?

genevb:
> The misalignments are not expected to vary over the course of the run, so we only expect one entry once they have been measured. The practice for HFT misalignments, and I believe TPC and others as well, has been to store them StarDb area.

For the TPC, I think surveys and ideal geometries have been stored as CINT files, but the misalignments used in production are stored in the database. If my memory serves me correctly, I think for the HFT that the CINT files were generated from the database tables, but I could be wrong.

-Gene

StRoot/StChain/GeometryDbAliases.h
klendathu2k: I have been trying to keep the R&D / development geometries well advanced in time, so as not to accidentally run them in a production context (thus the +100 years).

StRoot/StChain/GeometryDbAliases.h
klendathu2k: @genevb Comments updated.  Any other issues?

klendathu2k:
Looks like trigger simulator is breaking tests?

genevb:
> Looks like trigger simulator is breaking tests?

I have created an issue: #303 .

-Gene

veprbl:
FYI I've restarted CI after #304 was merged, the tests are green again

starsdong:
Are we good with this one? @jdbrice or @perevbnlgov need your approval to proceed.

plexoos (approved):
I trust this been tested and ready

starsdong:
Dmitri, let us move ahead and merge this in then. Thanks

PR #293, opened 2022-02-01
Rdo clean up

fisyak:
Remove junk cluster due to masked out RDO


starsdong:
Hi Yuri,

Thanks. I see the new added file  ...y2016.C. Just want to be sure that this PR is also needed for the Run19 BES-II embedding? Thanks

StRoot/StTpcHitMaker/StTpcHitMaker.cxx
```diff
@@ -397,7 +399,7 @@ Int_t StTpcHitMaker::InitRun(Int_t runnumber) {
       for(Int_t row=1;row<=St_tpcPadConfigC::instance()->numberOfRows(sector);row++) {
         Int_t numPadsAtRow = St_tpcPadConfigC::instance()->padsPerRow(sector,row);
         totalSecPads += numPadsAtRow;
-        if (StDetectorDbTpcRDOMasks::instance()->isRowOn(sector,row) &&
+        if (StDetectorDbTpcRDOMasks::instance()->isRowOn(sector,row,1) &&
```
iraklic, line 402: Hi @fisyak ,

Quick question before I do the approval. Where is this function defined `isOn(int, int, int)`? I see one here:
https://github.com/star-bnl/star-sw/blob/4641c7dfe27bcb88ee0ff673769e55258ead82a6/StRoot/StDetectorDbMaker/St_tpcRDOMasksC.h#L36

but it only has two arguments. What am I missing?

zhux97: Irakli, this is isRowOn

iraklic: Oh, thanks. My bad.

iraklic (approved):

starsdong (approved):

genevb:
Regarding this unanswered question...

> On Feb 1, 2022, at 5:44 PM, Xin Dong ***@***.***> wrote:
> 
>  Just want to be sure that this PR is also needed for the Run19 BES-II embedding?

My understanding is that this PR is not to be back-propagated as it affects real data reconstruction and was not used in the real data production for old libraries.

That said, I believe this exclusion of hits on dead RDOs *should* be in place for BES-II productions, along with exclusion of hits on lowered anode HV sections (lowered to the point of negligible gains, e.g. less than 1%). These were to-do items for the TPC group because some noise was seen during BES-II under both conditions that could have been picked up on reconstructed tracks. Since this one is merged, a separate PR will need be made for the anode HV exclusion. And then both of these will be included in future libraries (maybe it's about time to freeze SL22a).

Thanks,
-Gene

StRoot/StTpcHitMaker/StTpcHitMaker.cxx
fisyak: Hi Irakli, there are two implementations of isOn with 2 and 3 arguments.

PR #295, opened 2022-02-05
Fix EMC data in picoDst data in DAQ production

starsdong:
- picoDstMaker change to extract the emcCollection from StEvent during daq production
- BFC option added for TriggerSimuMaker + options

### DIFF
--- a/StRoot/StBFChain/BigFullChain.h
+++ b/StRoot/StBFChain/BigFullChain.h
@@ -1949,6 +1949,8 @@ Bfc_st BFC[] = { // standard chains
 #endif /* __NoStrangeMuDst__ */
   {"RMuDST"    ,"","","CMuDST"   ,"","","reads Common MuDST, do not disactivate if no output files",kFALSE},
 
+  {"trgSimu"        ,"","",""       ,"StTriggerSimuMaker","StTriggerUtilities","trigger simu maker",kFALSE},
+
   {"picoWrite" ,"","PicoChain","picoDst","StPicoDstMaker",""               ,"Writes picoDST format",kFALSE},
   {"picoRead"  ,"","PicoChain","picoDst","StPicoDstMaker",""           ,"WritesRead picoDST format",kFALSE},
   {"PicoVtxDefault" ,"","",""                                       ,"" ,"","pico Vtx default mode",kFALSE},
--- a/StRoot/StBFChain/BigFullChain.h
+++ b/StRoot/StBFChain/BigFullChain.h
@@ -1951,7 +1951,7 @@ Bfc_st BFC[] = { // standard chains
 
   {"trgSimu"        ,"","",""       ,"StTriggerSimuMaker","StTriggerUtilities","trigger simu maker",kFALSE},
 
-  {"picoWrite" ,"","PicoChain","picoDst","StPicoDstMaker",""               ,"Writes picoDST format",kFALSE},
+  {"picoWrite" ,"","PicoChain","trgSimu,picoDst","StPicoDstMaker",""       ,"Writes picoDST format",kFALSE},
   {"picoRead"  ,"","PicoChain","picoDst","StPicoDstMaker",""           ,"WritesRead picoDST format",kFALSE},
   {"PicoVtxDefault" ,"","",""                                       ,"" ,"","pico Vtx default mode",kFALSE},
   {"PicoVtxVpd"     ,"","","-PicoVtxDefault"             ,"" ,"","pico Vtx cut on Tof and VPD mode",kFALSE},
--- a/StRoot/StBFChain/StBFChain.cxx
+++ b/StRoot/StBFChain/StBFChain.cxx
@@ -657,13 +657,7 @@ Int_t StBFChain::Instantiate()
 
     // trigger simu maker
     if ( maker == "StTriggerSimuMaker" && GetOption("picoWrite") ) {
-      TString cmd(Form("StTriggerSimuMaker *pTSMk = (StTriggerSimuMaker*) %p;",mk));
-      cmd += "pTSMk->setMC(false);";                                 // Set TriggerSimuMaker parameters
-      cmd += "pTSMk->useBemc();";
-      cmd += "pTSMk->useEemc();";
-      cmd += "pTSMk->useOfflineDB();";
-      cmd += "pTSMk->bemc->setConfig(StBemcTriggerSimu::kOffline);";
-      ProcessLine(cmd);
+      mk->SetMode(10);  // picoDst production
     }
 
     // MuDST and ezTree. Combinations are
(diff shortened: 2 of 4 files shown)


genevb (commented):
Hi,

Glad to see this progressing to a PR! The maintenance of StTriggerSimuMaker was an issue in the past, and we seemed to move past that. However, if StTriggerSimuMaker hasn't been updated in over 4 years, I take that as a warning sign that lack of maintenance could become an issue again. Now it appears all our picoDst-producing production chains will depend on it.

If the maintainer of StTriggerSimuMaker is on-board with this PR, then I have a recommendation:

The 7 lines of code in StBFChain.cxx from 660->666 could be replaced with a single line:

    mk->SetMode(10);   // Set parameters necessary for filling PicoDst in BFC

Then near the start of StTriggerSimuMaker::Init() add these lines of code:

    if (GetMode() == 10) {
        // Set parameters necessary for filling PicoDst in BFC
        setMC(false);
        useBemc();
        useEemc();
        useOfflineDB();
        bemc->setConfig(StBemcTriggerSimu::kOffline);
    }

This keeps the complexities of the StTriggerSimuMaker code local to StTriggerSimuMaker. Others have used the ProcessLine() functionality in StBFChain, so I'm not saying this would set a bad precedent (that has already happened), but I would encourage not spreading these special instructions into the BFC code. (Mode 10 is a random choice, just leaving room for other possibly important modes to be used as 1, 2, etc., but I don't really care what mode number is chosen.)

-Gene

starsdong:
Updated following Gene's suggestion

genevb:
The relocation of setup specifics to StTriggerSimuMaker looks good to me.

One more thing:
I expect that the "picoWrite" option in BigFulLChain.h should now include "trgSimu" among the options it turns on (fourth argument becomes "trgSimu,picoDst"), unless the default should be that we wouldn't want trgSimu (but I get the impression that we do). This in particular ensures that the StTriggerSimuMaker runs before the StPicoDstMaker in the chain.

Thanks,
-Gene

starsdong:
Thanks for the suggestion, Gene. The picoWrite option is updated.

starsdong:
Per today's discussion, we would like to include this PR for the upcoming SL22a lib release. Could you please review the latest commit, comment or approve it soon? Thanks

klendathu2k (approved):
I don't see any problems with this pull request.  I approve of the PR, but note the following:

1) The trigger simulator has never gone through a formal code review. For long term maintenance (i.e. before STAR goes into an archival mode) we should be putting some effort into making sure that the trigger simulator can be used long term.  As I recall, there is some amount of expertise needed to make sure that it correctly simulates the trigger.

2) The trigger simulator is sitting in a "Utilities" directory, rather than a "Maker" directory.  This implies different build rules during compilation, and makes it less obvious where to find the code in the directory structure.

zlchang (approved):

starsdong:
Jason, I agree with your concerns and am also in strong favor that we have long term maintainer and maybe have it reviewed too.
One question, maybe to Gene. The TriggerSimuMaker is included in genDst.C that is used for our MuDst->picoDst production.  So presumably is it already included in our production lib build?

genevb:
Hi,

> One question, maybe to Gene. The TriggerSimuMaker is included in genDst.C that is used for our MuDst->picoDst production. So presumably is it already included in our production lib build?

1) StTriggerSimuMaker was in genDst.C from its original entry into CVS by @nigmatkulov (I never altered that part):
https://www.star.bnl.gov/cgi-bin/protected/viewvc.cgi/cvsroot/StRoot/StPicoDstMaker/macros/genDst.C?hideattic=0&revision=1.1&view=markup
There was a software peer review in 2007, but noted as unofficial because it wasn't used in production at the time. Another review was done in 2010 as it began to be used in official simulation/embedding productions, and it moved forward with a flag about database maintenance:
https://drupal.star.bnl.gov/STAR/comp/sofi/soft-n-libs/peer-review

2) We skip Pool directories, but not Utils/Utilities directories in compilation.

-Gene

starsdong:
Gene, thank for pointing out this history. It seems the review was concluded and the code moved forward with some concern on the DB access. I am not sure if we want to reopen this and address the concern especially at this moment considering we are all tied with many things. I would suggest we move forward and if this causes an issue with the production, we should flag as priority and address it asap.

zlchang:
Hi Xin,

I believe the concern was the older version was using MySQL query to access online database server. I have fixed that seven years ago to query the trigger information from offline DB. That’s why Zhenyu is now trying to populate the offline database. Please let me know if any further review is required.

Zilong

On Feb 15, 2022, at 5:31 PM, Xin Dong ***@***.***> wrote:

﻿

Gene, thank for pointing out this history. It seems the review was concluded and the code moved forward with some concern on the DB access. I am not sure if we want to reopen this and address the concern especially at this moment considering we are all tied with many things. I would suggest we move forward and if this causes an issue with the production, we should flag as priority and address it asap.

starsdong:
Thanks Zilong for the comment. It looks to be that there is no major concern regarding the StTriggerSimuMaker (besides its location under StTriggerUtilities as Jason mentioned). I would then suggest we go ahead and merge the request in and prepare for SL22a library. Dmitri, could you help help merge this one?

genevb:
Merging this PR immediately broke FastOffline jobs in DEV after the nightly AutoBuild ran. I backed out a couple of the changes to get it running again. Someone needs to debug this.

Example job:

    root4star -b -q -l 'bfc.C(4,"pp2022,StiCA,BEmcChkStat,epdhit,-hitfilt,-evout,-fcsCluster,-fcsPoint","/gpfs01/star/daq/2022/032/23032031/st_physics_adc_23032031_raw_6500079.daq")'

What I'm seeing at the crash:

    StTriggerSimuMaker:INFO  - runNumber=23032031 with DB timestamp 2022-02-01 16:59:15
    root4star: .sl73_gcc485/OBJ/StRoot/StTriggerUtilities/Eemc/StEemcTriggerSimu.cxx:737: void StEemcTriggerSimu::getEemcFeeMask(): Assertion `mDbE' failed.

-Gene

veprbl:
The same crash is now seen in CI: https://github.com/star-bnl/star-sw/runs/5221232278?check_suite_focus=true
Could be an interplay between two PRs that were merged together?

genevb:
> The same crash is now seen in CI: https://github.com/star-bnl/star-sw/runs/5221232278?check_suite_focus=true
> Could be an interplay between two PRs that were merged together?

Which two? I thought this was the only one for Feb. 16, other than a CODEOWNERS one.

-Gene

veprbl:
> > The same crash is now seen in CI: https://github.com/star-bnl/star-sw/runs/5221232278?check_suite_focus=true
> > Could be an interplay between two PRs that were merged together?
> 
> Which two? I thought this was the only one for Feb. 16, other than a CODEOWNERS one.
> 

It's just a guess. The CI for this PR was run 9 days ago, so it couldn't have taken into account whatever was merged in that time.

I'm trying reverting this in my fork https://github.com/veprbl/star-sw/pull/15

genevb:
> I'm trying reverting this in my fork [veprbl#15](https://github.com/veprbl/star-sw/pull/15)

I reverted the changes to `StPicoDstMaker.cxx` and `BigFullChain.h` to get DEV working a few hours ago (after making sure the jobs ran with some local tests). Seems to be fine without those. I suspect it's more to do with support for recent data (e.g. 2022) with StTriggerSimuMaker than a conflicting code commit. A maintenance issue.

PR #299, opened 2022-02-11
Fix crash in StFcsPi0FinderForEcal with MuDst VPD vertex in FastOffli…

### LINKED ISSUE #297: st_fwd crashing in StFcsPi0FinderForECal after PR #291
It appears that every st_fwd stream file produced in DEV for FastOffline since PR #291 was merged is seg faulting. The log files from FastOffline all show lines like this:
```
Start using StFcsPi0FinderForECal
current event:2
StMaker:FATAL - TUnixSystem::DispatchSignals : segmentation violation
```
(though it does not always appear to be event 2; sometimes it is event 0 or 1, etc.)

-Gene

akioogawa:
Fix crash in StFcsPi0FinderForEcal with MuDst VPD vertex in FastOffline/BFC.

StFcsWaveformFitMaker's default for mEnergySelect separated for Ecal,Hcal(gaus fit) & Pres(Sum8) to speed it up. Also when there are too many peaks found, take normalized sum8 instead of just quit fitting.

### DIFF
--- a/StRoot/StFcsWaveformFitMaker/StFcsWaveformFitMaker.cxx
+++ b/StRoot/StFcsWaveformFitMaker/StFcsWaveformFitMaker.cxx
@@ -129,6 +129,9 @@ namespace {
 StFcsWaveformFitMaker::StFcsWaveformFitMaker(const char* name) : StMaker(name) {
     mChWaveData.SetClass("TGraphAsymmErrors"); //Initialize with only one graph at first
     mPulseFit = new StFcsPulseFit( (TGraphAsymmErrors*)mChWaveData.ConstructedAt(0) );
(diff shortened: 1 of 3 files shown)


zlchang (approved):

genevb (approved):
The checks on null pointers look good to me. I'm guessing that a null StMuDst::btofHeader() was the reason for the crashes we were seeing. Other than those fixes, most of it is details for experts to understand. I don't see anything obviously questionable in this PR, so I don't object.

veprbl (approved):

jdbrice (approved):
These changes look good to me. I am approving and merging (since I am second to sign off)

PR #300, opened 2022-02-15
ci: grep output log to ensure there was no segfault

### LINKED ISSUE #249: Status of failed CI tests should reflect when a job fails
Here is an example of a test that is expected to fail https://github.com/star-bnl/star-sw/runs/4516531310?check_suite_focus=true due to the test data being more recent than the code.

```shell
 *** Start at Date : Tue Dec 14 07:37:08 2021

Processing bfc.C(2000, "P2021a,StiCA,EbyET0,BEmcChkStat,-hitfilt,QAalltrigs", "/star/rcf/test/daq/2021/157/st_physics_22157006_raw_2000022.daq")...
Load system libraries	 + liblog4cxx.so + libStStarLogger.so
Create chain bfc
BFC:WARN  - StBFChain::Setup	File BFC.C has been found as /star-sw/StRoot/StBFChain/BFC.C
BFC:ERROR - TSystem::ExpandFileName : input: $STAR_VERSION, output: $STAR_VERSION
QA :INFO  - ==============================================
QA :INFO  - ============= You are in $STAR_VERSION ===============
QA :INFO  - Requested chain bfc is :	P2021a,StiCA,EbyET0,BEmcChkStat,-hitfilt,QAalltrigs
QA :INFO  -  Invalid Option p2021a. !! ABORT !! 
BFC:FATAL - TUnixSystem::DispatchSignals : segmentation violation
```

Nevertheless, the status indicates that the test has passed successfully.
One solution would be to extend the test command with a check parsing the log file for "segmentation violation" and/or check for empty output files.

veprbl:
Fixes: #249

This is only a very simple check. Ideally, we would check that the produced files are correct, but this is better than nothing.

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -76,7 +76,9 @@ jobs:
       - run: docker run --name star-test-data --volume /star ghcr.io/star-bnl/star-test-data:v2
       - run: |
              TEST_CMD=$(docker run --rm ghcr.io/star-bnl/star-sw-root5-build star-sw/tests/executest.py -c ${{ matrix.test_id }})
-             docker run --volumes-from star-test-data ghcr.io/star-bnl/star-sw-root5-build sh -c "$TEST_CMD"
+             # Workaround https://sft.its.cern.ch/jira/browse/ROOT-7660 in ROOT 5 by checking the output log
+             docker run --volumes-from star-test-data ghcr.io/star-bnl/star-sw-root5-build \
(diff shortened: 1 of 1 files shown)


veprbl:
Would be also nice to additionally get some `pp2022` testing going to prevent regressions like #297.

veprbl:
Here is the demonstration of how this rightfully fails on SL19b: https://github.com/veprbl/star-sw/pull/14

plexoos (approved):

PR #312, opened 2022-02-23
ci: run tests with MALLOC_CHECK_=3

veprbl:
Enables additional heap consistency checks in glibc
https://www.gnu.org/software/libc/manual/html_node/Heap-Consistency-Checking.html


plexoos (approved):

plexoos:
Grigory, does the proposed change look ok to you? Please let us know.

PR #315, opened 2022-03-03
Simu pp200 run14 pythia8 hfjets

klendathu2k:
Addition of fast jet filter for HF jet simulation.

Fix to StarGenerator/DECAY -- advance past any pythia "system" entries when checking decay chains for validity.

Setup, macros  and workflow for HF D0-jet simulation.


genevb (approved):
Honestly, pretty much all of this PR is outside of my familiarity and I'm not very confident about what I'm approving here. Trusting Jason on this one.

starsdong (approved):
Trust Jason on this development. This has been used for embedding.

starsdong:
Dmitri, if you don't object, let us also merge this into main. Thanks

StRoot/StarGenerator/macros/production/setup.simulation.y2014x.pythia8.HFjets
```diff
@@ -0,0 +1,28 @@
+#!/bin/csh -f
+
+echo ""
+echo "Building FastJetFilter maker"
+echo ""
+
+cd star-sw-1
```
plexoos, line 14: Shouldn't this directory be created first?
This script does not seem like it can be used out of the box.

StRoot/StarGenerator/macros/starsim.pythia8.C
```diff
@@ -50,7 +53,7 @@ void Pythia8( TString config="pp:W" )
   //
   StarPythia8 *pythia8 = new StarPythia8();    
   if ( config=="pp:W" )
-    {
+    {-
```
plexoos: Does this code compile?

mgr/Conscript-standard
```diff
@@ -568,6 +568,9 @@ if ( $pkg !~ /^sim$/ && $pkg !~ /^gen$/ ) {
 	print "CPPPATH = $CPPPATH\n" if $param::debug;
 
 
+    } elsif ( $pkg =~ m/^FastJetFilter/ and $Dir =~ m/StarGenerator/ ){
+	$CPPPATH = $env1->{ENV}->{OPTSTAR}."/include".$main::PATH_SEPARATOR.$CPPPATH;
```
plexoos: The config variable provided by the package should be preferred

https://github.com/star-bnl/star-sw/blob/11f02d137a851c7cb00a6ec945dd0451b1d46ebc/mgr/Conscript-standard#L770-L774

StRoot/StarGenerator/macros/production/simulation.y2014x.pythia8.HFjets.input.list
```diff
@@ -0,0 +1,10000 @@
+/star/simu/simu/gstardata/Simu2022/rcf22000/vertex_files/15165031_0014
```
plexoos, line 1: Is it possible to run the submitted macros as is without regenerating these files?

StRoot/StarGenerator/macros/production/setup.simulation.y2014x.pythia8.HFjets
klendathu2k: This macro assumes the user has checked the code out of my repository, and builds.  Once the pull request goes through, this is for documentation purposes only.

StRoot/StarGenerator/macros/starsim.pythia8.C
klendathu2k: No, it won't compile.  But surprisingly, CINT doesn't care about this typo (as long as you don't  follow the pp:W branch of the code).  Updating.

StRoot/StarGenerator/macros/production/simulation.y2014x.pythia8.HFjets.input.list
klendathu2k: Those files are an input dependency to the job.  The format of the files are:

Run#   Event#   Vx   Vy   Vz...

Updating the macro with the specification.   These files would need to be regenerated when comparing against any production of the data... (any change which alters vertex finding would require a new set of inputs).

StRoot/StarGenerator/macros/production/setup.simulation.y2014x.pythia8.HFjets
plexoos: That is a pretty strong assumption if you think about the long term code preservation and reproducibility.

StRoot/StarGenerator/macros/production/simulation.y2014x.pythia8.HFjets.input.list
plexoos: If you explain the format of the input files and how to regenerate them then there is not much point in including this particular list which, as you say, need to be regenerated for other data anyway.

starsdong:
Victor, it seems this PR needs your approval before merging. If you can take a look and comment/approve it. Thanks

perevbnlgov (approved):
I do not see anything suspicious

genevb:
This PR appears to have failed the AutoBuild. I've opened issue #324 .

klendathu2k:
This should be resolved by PR#325.  Details in the commit message.  But
the short summary is that I broke the build when I reorganized the
source tree (moved FastJetFilter one level higher).  Cons no longer
matched the directory name, and did not apply the include for the
fast jet library.

(We were already in production for the simulation at that point, so
this did not impact the results).

PR #318, opened 2022-03-06
Star6 01

perevbnlgov:
Outdated StKF* and KFP* and KFV* removed. Code which is used it also removed.
According LinDef is changed. By request of Y.Fisyak


ullrich-bnl (approved):

starsdong:
Victor, thank you very much for pushing this forward. One quick question, the changes proposed in StBFChain in this PR don't seem to be related to the need for ROOT6. Or did I miss something here?
On the proposed changes to mgr/, Dmitri or Grigory would be nice to hear your comments. Thanks

StRoot/StBFChain/doc/BFC.h
```diff
@@ -0,0 +1,1693 @@
+Bfc_st BFC[] = {
```
klendathu2k, line 1: Files under StBFChain/doc are generated during the build process.  They should not be committed into this directory.  (Use "git rm" to remove the files).

StRoot/StBFChain/doc/BFC2.h
```diff
@@ -0,0 +1,1693 @@
+Bfc_st BFC[] = {
```
klendathu2k, line 1: Remove

StRoot/StBFChain/doc/StBFChain.cxx_doc
```diff
@@ -1,4 +1,4 @@
-// @(#)StRoot/StBFChain:$Name:  $:$Id: StBFChain.cxx_doc,v 1.180 2021/03/30 14:45:42 genevb Exp $
+// @(#)StRoot/StBFChain:$Name:  $:$Id: StBFChain.cxx,v 1.670 2021/03/21 03:32:28 genevb Exp $
```
klendathu2k, line 1: ... looks like we are replacing a newer revision with an older one (1.670 vs 1.180)... but again, these files are generated at compile time.  Should remove this.

mgr/ConsDefs.pm
```diff
@@ -173,8 +174,9 @@
     $CXXFLAGS = $CFLAGS = $LDFLAGS = $SOFLAGS = "$XMACHOPT";
 
     $CXXFLAGS      .= " -fPIC -w";
+$CXXFLAGS .= "-fdiagnostics-color=always -msse4.2 -std=c++11 -pipe  -Wshadow -Wall -W -Woverloaded-virtual -fsigned-char -pthread -DNDEBUG";
```
klendathu2k, line 177: Not clear why we need the -msse4.2 option.  Is this for root 6, or are we just picking up a change for the heck of it?

Also... defines NDEBUG... which I am not convinced is the right thing to do here.  (Doesn't this break logic for creating debug vs nodebug builds?  Has this change to the CXXFLAGS been tested... beyond "it compiles, it runs, and plots can be made that sorta kinda look right")

mgr/RootCint.pl
```diff
@@ -437,9 +440,10 @@
     #	$cmd  = "rootcint -f $Cint_cxx -c  -D__NO_STRANGE_MUDST__ -DROOT_CINT -D__ROOT__ $CPPFLAGS $h_files";
     #	print "cmd (+extra) = ",$cmd,"\n";
     #} else {
-    	$cmd  = "rootcint -f $Cint_cxx -c -DROOT_CINT -D__ROOT__ $CPPFLAGS $h_files";
+##    	$cmd  = "rootcint  -f $Cint_cxx -c -DROOT_CINT -D__ROOT__ $CPPFLAGS $h_files";
+    	$cmd  = "rootcling -rmf $Cint_map -f $Cint_cxx  -DROOT_CINT -D__ROOT__ $CPPFLAGS $h_files";
```
klendathu2k, line 444: Does this work for our ROOT5 builds as well?  (This commit _cannot_ break the ROOT5 build.)

mgr/RootCint.pl
```diff
@@ -437,9 +440,10 @@
...
      	print "cmd (normal)= ",$cmd,"\n";
-    #	die;
+##    	die;
```
klendathu2k, line 446: Yes.  Let's _really_ make sure this line is commented out ;)

mgr/RootCint.pl
klendathu2k: ... probably should have logic testing against the root version we compile against, and use root5/rootcint versus root6/rootcling.

StRoot/StBFChain/StBFChain.cxx
```diff
@@ -638,11 +638,14 @@ Int_t StBFChain::Instantiate()
     }
 
     if (maker == "StStrangeMuDstMaker" && GetOption("CMuDST")&& GetOption("StrngMuDST") ) {
-      TString cmd(Form("StStrangeMuDstMaker *pSMMk = (StStrangeMuDstMaker*) %p;",mk));
-      cmd += "pSMMk->DoV0();";                                 // Set StrangeMuDstMaker parameters
-      cmd += "pSMMk->DoXi();";
-      cmd += "pSMMk->DoKink();";
-      cmd += "pSMMk->SetNoKeep();";                            // Set flag for output OFF
+printf ("UUU Mk = %p\n",mk);
```
klendathu2k, line 641: This introduces a change to the StStrangeMuDstMaker configuration... and also creates meaningless output.

The old version needs to be restored.

klendathu2k (changes requested):
Thank you for pruning this down to a minimal set of changes.  There are several issues identified, and a couple of questions.  

But I note that the CI builds have failed for both ROOT5 and ROOT6.  Regardless of getting two approvals, the code must actually compile under both ROOT5 and ROOT6... (under both cons and the CI builds).

plexoos:
> On the proposed changes to mgr/, Dmitri or Grigory would be nice to hear your comments. Thanks

I agree with the Jason's remarks. Before we continue with any further discussion the code should compile.

StRoot/StBFChain/doc/BFC.h
perevbnlgov: Yes, I know, I am trying not to touch these doc files, ut they are arised again. I did not find how to ignore them.
Best of all to remove them at all. Automatically created files should not be in ane CVS or Git repository.
They shoukd be created in some temporary directories like .o, .so ...

perevbnlgov:
pull request Star6_01 closed without merging. Wrong request.

PR #319, opened 2022-03-08
Tpc rdo mask fix

fisyak:
Two Updates:
1) An additional correction for dE/dx handled by a new table identified by StTpcdEdxCorrection::kzCorrectionC which is currently activated only for the Run 19 AuAu200 data. The authors claim that the change won't affect any other dataset. A slide illustrating the related problem is available at https://drupal.star.bnl.gov/STAR/event/2022/03/09/SC-Management-Meeting/Run19-AuAu200-Calibration-Status

The change in StRoot/StDetectorDbMaker looks pretty safe as it only adds a diagnostic routine St_TpcAvgPowerSupplyC::PrintC()

2) TPC RDO masking to deal with timestamp issue with full simulation for Run19 data (no impact on embedding).

### DIFF
--- a/StDb/idl/tpcCorrection.idl
+++ b/StDb/idl/tpcCorrection.idl
@@ -17,6 +17,7 @@ struct tpcCorrection {
                 // type = 6 X = abs(x);
                 // type = 10 (== log(1. + OffSet/x) + poly(x,npar))                  -"-
                 // type = 11 (== log(1. + OffSet/x) + poly(x,npar) for log(ADC) and |Z|
+                // type = 20 has extra 2 parameters to add expo => a[npar]*exp(a[npar+1]*x) to polynomical fit 
                 // type = 200 cut on range [min,max]
                 // type = 300 don't correct out of range [min,max]
                 // type = 400 special for TpcZCorrectionB, to enable Gating Grid correction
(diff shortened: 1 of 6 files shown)


genevb:
I see there are changes to the TPC dE/dx codes here. Is this PR needed for dE/dx in Run 19 AuAu200?

-Gene

iraklic (approved):

genevb:
> I see there are changes to the TPC dE/dx codes here. Is this PR needed for dE/dx in Run 19 AuAu200?

Yuri stated at today's S&C meeting that this PR is needed for embedding, but not reconstruction. In other words, if this doesn't make it into SL22a (which will be used to produce Run 19 AuAu200) now, Yuri wants it in SL22a eventually. He also implied it will be needed for other embedding libraries, but I'm not 100% sure which ones (all used to produce iTPC data so far?).

-Gene

starsdong:
Yuri, two quick questions
1) I see this PR includes the commit [d6e6217](https://github.com/star-bnl/star-sw/pull/319/commits/d6e621768cbdf51e68e2ae22dbb39ca7de028c7e) which should be the same as in PR 317. So if this gets merged, we can drop out 317.
2) On the TPC masking part, which datasets will need this? Or this is only for Run19 AuAu 200 GeV?

starsdong:
Here are the comments from Yuri: 
"tpcRDOMask correction (PR319) is affect only simulation for Run XIX. It is not affected embedding. The game is with time stamp. The problem has been appeared when it is used simulation time stamp (2018-12-10). In embedding time stamp is coming from real data and this problem does not showed up."
I think we can go ahead with merging this PR. Since this includes PR 317, we can then ignore 317.

starsdong (approved):
See comments.

starsdong:
Dmitry or Dmitri, please take a review. If no object, please approve and merge this PR. Thanks.

starsdong:
Dmitri, just to clarify, this PR is needed for the SL22a lib tag which will be used for the upcoming Run19 AuAu200 GeV production.

dmarkh (approved):

starsdong:
Thanks, Dmitry. I am going to merge this to main.

PR #320, opened 2022-03-09
Keep short tracks pointing to ETOF

genevb:
ETOF has not been removed after Run 21 and will remain with STAR, so short tracks pointing to ETOF should continue being kept.


starsdong (approved):
Sure

klendathu2k (approved):
Approved.

PR #322, opened 2022-03-10
Restore GetCVS() function in StTriggerSimuMaker

genevb:
While the CVS tags don't work any longer with the code now sitting in git, the date and time of compilation are still useful pieces of information from this function that are displayed in the log files. Otherwise the date and time of StMaker's compilation are shown for this maker from the inherited function. Additionally, the log files are polluted by a warning message if GetCVS() is not overloaded from StMaker.

### DIFF
--- a/StRoot/StTriggerUtilities/StTriggerSimuMaker.h
+++ b/StRoot/StTriggerUtilities/StTriggerSimuMaker.h
@@ -128,8 +128,8 @@ public:
   void useOfflineDB() { mUseOfflineDB = 1; }
 
 
-  //virtual const char *GetCVS() const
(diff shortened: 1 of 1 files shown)


zlchang (approved):

starsdong (approved):

plexoos (approved):
It is far more common to identify libraries by the version of the source code rather than the time when they are built. I don't see much value in this but I guess it does not hurt either.
It is worth noting that `__DATE__` and `__TIME__` are updated only when the file is recompiled. If the maker relies on another source file its functionality can still change without a proper update in GetCVS()

genevb:
@plexoos , I agree that the date and time of compilation are of quite limited usefulness, but there are (rare) occasions when looking through the nightly test log files it helps us spot a problem. As you noted, other information is generally more relevant; this is just additional information.

-Gene

PR #330, opened 2022-03-17
Star6

perevbnlgov:
Upstream::Star6 iz expected to be exactly the same as origin::Star6
So this request is needed only to convince internal structures of GitHub that they are the same.


klendathu2k (approved):
I don't see anything problematic here, especially since this is being committed to the Star6 branch on star-bnl.

iraklic:
Should I look over all 34 files that have changes in this commit? or are they already approved (as some comments seem to suggest). If you can point me to those that need review that will save me some time. Thanks.

perevbnlgov:
Irakli, as I wrote these files upstream::Star6 and origin::Star6 are 
exactly the same.
So there is no a reason to to commit. But I have made full cleanup of 
origin,
removing some garbage. Then I filled it by upstream::Star6 using some 
workaround,
which I found. In Github direct methods usually does not work or does 
not exist.
Bit in spite of these files are the same, I afraid that internally 
GitHub could find
something illegal. So I do this PR just technically. Unfortunately I can 
not do without
agreement with others. So you need only agree.
Victor

iraklic:
Thank you for the clarification Victor.

iraklic (approved):

akioogawa (approved):

PR #336, opened 2022-03-30
Avoid onldb master usage when possible

genevb:
This is a general sweep of code in the repository to avoid accessing the online master DB unless writing, and possibly use offline replicas if not using code from within the online enclave. People tend to copy existing code when they write new codes and it's best if everyone, old and new, makes a habit of reading from replicas (@dmarkh may wish to comment).

@akioogawa , I also updated some logic you use in StRoot/StFcsFastSimulatorMaker/macro/thresh.pl for determining which backup port to access, so please have a look. You were relying on a Jan. 1st rollover from one database to another, but Oct. 1st is more appropriate.

-Gene

### DIFF
--- a/StRoot/StFcsFastSimulatorMaker/macro/thresh.pl
+++ b/StRoot/StFcsFastSimulatorMaker/macro/thresh.pl
@@ -1,8 +1,8 @@
 #!/usr/bin/perl
 #
 #  Server and Port depend on run year:
-#   Year : ongoing  onldb:3501
-#          20xx     onlsun1:3400+x-1
+#   Year : ongoing  heston:3501
+#          20xx     db04:3400+year-1
 
 if (@ARGV != 1) {
     printf("thresh.pl run\n");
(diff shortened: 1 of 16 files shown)


dmarkh:
Thank you, Gene! I totally support your changes. 

Yes, using master DBs for read-only operations should be avoided in most cases. STAR has plenty of replica DB servers available - these will generally work faster than masters, and will not prevent/block STAR services from inserting data into master DBs.

plexoos (commented):
There are scripts in /StRoot/StFgtPool and /StRoot/StSpinPool that also use `onldb.starp.bnl.gov`. Why leave them out?

plexoos (approved):

veprbl (approved):

zlchang (approved):

starsdong:
Akio and Daniel, if you can take a review on the relevant part that you are the owner, comment/approve the PR? Thanks

plexoos:
Does it make sense to wait for other approvals any longer?

starsdong:
Dmitri, please proceed and merge this PR. Thanks

PR #337, opened 2022-04-10
Add StBadRunChecker

liuxiaoyuyuyu:
StBadRunChecker is used to check bad runs based on different STAR sub-systems for BESII data. BadRunSubSys.h is used to store the bad runs based on subsystems. It now has run19 19.6 GeV injection runs and TPC bad runs.


StRoot/StRefMultCorr/BadRunSubSys.h
```diff
@@ -0,0 +1,38 @@
...
...1,20071016,20071019,20068058,20068055,20068051,20074029,20074027,20074026,20074023,20074021,20074020,20070016,20070013,20070010,20092057,20092054,20092053,20070019,20075011,20075010,20076051,20076054,20076059,20082059,20082056,20082053,20082050,20076025,20076021,20081018,20060012,20081014,20081015,20081012,20076028,20091020,20072047,20072046,20072045,20072041,20086056,20071050,20076031,20068026,20069001,20069007,20069004,20071059,20078019,20078016,20071056,20078013,20071053,20083019,20062009,20062007,20092018,20093016,20067038,20093010,20092012,20067030,20092015,20077017,20077014,20077011,20064040,20077018,20058003,20071029,20058001,20090048,20075054,20075057,20058004,20071024,20071027,20070002,20088009,20088005,20072055,20086015,20086012,20086011,20082024,20082029,20091011,20076017,20076010,20076013,20069042,20078051,20078057,20078054,20087021,20072036,20072037,20072034,20072035,20072038,20072039,20089028,20089003,20084002,20084001,20084005,20084009,20080016,20075015,20075014,20064008,20064009,20075013,20090005,20090002,20071063,20090008,20068060,20068064,20074012,20074010,20074011,20074016,20074017,20074014,20058002,20093025,20074018,20074019,20070005,20058005,20075060,20075066};
```
marrbnl, line 54: Judging from the way this array is used in the cxx file, I think there should not be any duplicated entries in the array. If this is true, please add a comment such that future users do not prepare the wrong format. At the moment, the injection runs and TPC bad runs do not overlap by design, as the TPC bad runs are determined after removing the injection runs. However, the bad run lists for different subsystems could in principle overlap, and one would need a dedicated macro or script to prepare the array from the lists of bad runs.

StRoot/StRefMultCorr/BadRunSubSys.h
```diff
@@ -0,0 +1,38 @@
...
...0},{1,0,0,0,0,0,0,0,0,0,0,0},{1,0,0,0,0,0,0,0,0,0,0,0},{1,0,0,0,0,0,0,0,0,0,0,0},{1,0,0,0,0,0,0,0,0,0,0,0},{1,0,0,0,0,0,0,0,0,0,0,0},{1,0,0,0,0,0,0,0,0,0,0,0},{1,0,0,0,0,0,0,0,0,0,0,0},{1,0,0,0,0,0,0,0,0,0,0,0},{0,1,0,0,0,0,0,0,0,0,0,0},{1,0,0,0,0,0,0,0,0,0,0,0},{0,1,0,0,0,0,0,0,0,0,0,0},{0,1,0,0,0,0,0,0,0,0,0,0},{1,0,0,0,0,0,0,0,0,0,0,0},{0,1,0,0,0,0,0,0,0,0,0,0},{1,0,0,0,0,0,0,0,0,0,0,0},{1,0,0,0,0,0,0,0,0,0,0,0},{1,0,0,0,0,0,0,0,0,0,0,0},{1,0,0,0,0,0,0,0,0,0,0,0},{1,0,0,0,0,0,0,0,0,0,0,0},{1,0,0,0,0,0,0,0,0,0,0,0},{1,0,0,0,0,0,0,0,0,0,0,0},{1,0,0,0,0,0,0,0,0,0,0,0},{1,0,0,0,0,0,0,0,0,0,0,0},{0,1,0,0,0,0,0,0,0,0,0,0},{0,1,0,0,0,0,0,0,0,0,0,0},{0,1,0,0,0,0,0,0,0,0,0,0},{1,0,0,0,0,0,0,0,0,0,0,0},{1,0,0,0,0,0,0,0,0,0,0,0},{1,0,0,0,0,0,0,0,0,0,0,0},{1,0,0,0,0,0,0,0,0,0,0,0},{1,0,0,0,0,0,0,0,0,0,0,0},{1,0,0,0,0,0,0,0,0,0,0,0},{0,1,0,0,0,0,0,0,0,0,0,0},{1,0,0,0,0,0,0,0,0,0,0,0},{1,0,0,0,0,0,0,0,0,0,0,0},{1,0,0,0,0,0,0,0,0,0,0,0},{0,1,0,0,0,0,0,0,0,0,0,0},{0,1,0,0,0,0,0,0,0,0,0,0},{0,1,0,0,0,0,0,0,0,0,0,0},{1,0,0,0,0,0,0,0,0,0,0,0},{0,1,0,0,0,0,0,0,0,0,0,0},{1,0,0,0,0,0,0,0,0,0,0,0},{1,0,0,0,0,0,0,0,0,0,0,0},{0,1,0,0,0,0,0,0,0,0,0,0},{1,0,0,0,0,0,0,0,0,0,0,0},{1,0,0,0,0,0,0,0,0,0,0,0}};
```
marrbnl: Similar to the comment for line 31, a dedicated macro/script is needed to prepare this list. We need to upload those macros/scripts as well in another PR

Maybe you can use "isbadrun" instead of "badrun" to make it clear that the numbers in this array indicate whether a particular run from the "badrun_*_auau_all" array is bad or not for a particular subsystem. It took me a while to figure it out. Would be good to add a comment about this. You can also consider using Bool_t instead of Int_t

StRoot/StRefMultCorr/StBadRunChecker.cxx
```diff
@@ -0,0 +1,317 @@
...
+        exit(0);
+    }
+    mS.ToLower(); 
+    vector<Int_t>::iterator iter = std::find(mBadRun_all.begin(), mBadRun_all.end(), RunId);
+    if(iter == mBadRun_all.end()){
+        return 0;//RunId is not a bad run 
+    }
+    //return ( iter != mBadRun.end() ) ;
+    else{
+        int index = std::distance(mBadRun_all.begin(), iter);//index starts at zero
```
marrbnl, line 108: For this function to work, one needs to make sure there are no duplicated entries in "badrun_*_all" array

StRoot/StRefMultCorr/StBadRunChecker.cxx
```diff
@@ -0,0 +1,317 @@
...
+            if(mS.Contains(mSubSysName[iname])){
+                cout<<"Check if this run is bad for: "<<mSubSysName[iname]<<endl;
+                isbad*=(1-mBadRun_sub[index][iname]);
+            }
+        }
+        return (isbad==0);
+    }
+}
+
+Bool_t StBadRunChecker::isInjection(const Int_t RunId){
```
marrbnl, line 122: Instead of implementing new codes, could you just call "isBadRun(RunId, Injection)"? Same comments for all other subsystem-specific functions.

StRoot/StRefMultCorr/StBadRunChecker.cxx
```diff
@@ -0,0 +1,317 @@
...
+    }
+    mS.ToLower(); 
+    vector<Int_t>::iterator iter = std::find(mBadRun_all.begin(), mBadRun_all.end(), RunId);
+    if(iter == mBadRun_all.end()){
+        return 0;//RunId is not a bad run 
+    }
+    //return ( iter != mBadRun.end() ) ;
+    else{
+        int index = std::distance(mBadRun_all.begin(), iter);//index starts at zero
+        int isbad=1;
```
marrbnl: If I understand correctly, you use "isbad=0" as the run being bad and "isbad=1" being good, which is a bit counter-intuitive. This is also opposite to the convention used in "badrun_*_sub".

StRoot/StRefMultCorr/StBadRunChecker.h
```diff
@@ -0,0 +1,50 @@
...
+  Bool_t isBadRunbTOFPID(const Int_t RunId) ;
+  Bool_t isBadRuneTOF(const Int_t RunId) ;
+  Bool_t isBadRunEPD(const Int_t RunId) ;
+  Bool_t isBadRunVPD(const Int_t RunId) ;
+  Bool_t isBadRunBEMCStatus(const Int_t RunId) ;
+  Bool_t isBadRunBEMCPID(const Int_t RunId) ;
+  Bool_t isBadRunBEMCTrigger(const Int_t RunId) ;
+  Bool_t isBadRunMTD(const Int_t RunId) ;
+  Bool_t isBadRunAnalysis(const Int_t RunId) ; 
+  Bool_t isBadRun(const Int_t RunId,TString mSys) ;
```
marrbnl, line 34: Would be good to add a comment that "mSys" could be for a single sub-system or a combination of different sub-systems, such as "injection" or "tpc_etof_vpd", etc.

StRoot/StRefMultCorr/StBadRunChecker.h
```diff
@@ -0,0 +1,50 @@
...
+  //std::vector<Int_t> mStop_runId        ; /// Stop run id
+
+  //std::multimap<std::pair<Double_t, Int_t>, Int_t> mBeginRun ; /// Begin run number for a given (energy, year)
+  //std::multimap<std::pair<Double_t, Int_t>, Int_t> mEndRun   ; /// End run number for a given (energy, year)
+  std::vector<Int_t> mBadRun_all; /// Bad run number list
+  std::vector<std::vector<Int_t>> mBadRun_sub;
+  std::vector<Int_t> mRunRange;
+  //const TString mSubSysName[12]={"Injection","TPC","bTOFStatus","bTOFPID","eTOF","EPD","VPD","BEMCStatus","BEMCPID","BEMCTrigger","MTD","Analysis"};
+  TString mSubSysName[12]={"Injection","TPC","bTOFStatus","bTOFPID","eTOF","EPD","VPD","BEMCStatus","BEMCPID","BEMCTrigger","MTD","Analysis"};
+  TString mRun;
```
marrbnl, line 49: Add comments to mRun, mEnergy, mColMode, mSpecies about the code expects them to be look like, such as "run15, run19", "200, 19.6", "col, fxt", "auau, pau", etc. In this way, users know how to set them up.

StRoot/StRefMultCorr/BadRunSubSys.h
liuxiaoyuyuyu: I have a python script for preparing the correct format, I will add it to the next pull request.

StRoot/StRefMultCorr/BadRunSubSys.h
liuxiaoyuyuyu: Will do

StRoot/StRefMultCorr/StBadRunChecker.cxx
liuxiaoyuyuyu: Yes, the python script can make sure there is no duplicated entries in "badrun_*_all" array

StRoot/StRefMultCorr/StBadRunChecker.cxx
liuxiaoyuyuyu: Yes, you can just call isBadRun(RunId,injection). The subsystem-specific functions are indeed optional. I was thinking maybe some users prefer calling methods with explicit names instead of passing arguments to a method themselves, so they don't need to worry about typos? It doesn't hurt to keep both but I am also okay with removing all the subsystem-specific functions if you think it will cause confusion.

StRoot/StRefMultCorr/StBadRunChecker.cxx
liuxiaoyuyuyu: You are right. I want to implement a "or" logic in this method, so it is easier to set isbad=0 for bad runs. I can change the name to "isgood" to avoid confusion.

StRoot/StRefMultCorr/StBadRunChecker.h
liuxiaoyuyuyu: Will do

StRoot/StRefMultCorr/StBadRunChecker.cxx
marrbnl: I think it is a good idea to have those subsystem-specific functions. My comment was that you can use the master isBadRun() function for implementation, instead of writing somewhat duplicated codes. Something like: 
Bool_t StBadRunChecker::isInjection(const Int_t RunId){
return isBadRun(RunId,injection);
}

liuxiaoyuyuyu: I see! Good idea!

nigmatkulov:
Hi @liuxiaoyuyuyu , please create a new branch (from main) and commit modifications related to StBadRunChecker there.
When it will be ready please let me know. I'll close this commit.

PR #282, opened 2022-01-17
Fst QA and Calibration Makers

jdbrice:
This pull request builds on PR#266 and is needed for offline QA.

PR #338, opened 2022-04-22
Fix the issues causing FST codes not running

techuan-huang:
There are two issues in the current DEV causing FST codes not running.
The first one is StFstUtil not loaded before StFstRawHitMaker in the chain options. The second is the incorrect range of FST sensor id in StarDb geometry.

In order to run FST codes, I made following changes:
1. Clean up duplicate chain options for FST.
2. Add a chain option for StFstUtil.
3. Change the range of sensor id from 1-108 to 1000-1107 in StarDb/Geometry/fst/fstSensorOnWedge.20211110.000001.C.


StarDb/Geometry/fst/fstSensorOnWedge.20211110.000001.C
```diff
@@ -5,7 +5,7 @@ St_Survey *tableSet = new St_Survey("fstSensorOnWedge",108);
 //
 for(int i = 0; i < 108; i++){
     memset(&row,0,tableSet->GetRowSize());
-        row.Id   = i+1;
+        row.Id   = i+1000;
```
plexoos, line 8: Could you point to the code where it is used so we can see the problem?

StRoot/StBFChain/BigFullChain.h
```diff
@@ -1580,25 +1581,12 @@ Bfc_st BFC[] = { // standard chains
...
-  {"fstFastSim","fstChain","","StMcEvent,StEvent","StFstFastSimMaker","StFstSimMaker","FST fast simulator", 
-                                                                                                    kFALSE},
-  {"fstRawHit", "", "fstChain", "fstUtil,fstDb","StFstRawHitMaker", "StFstRawHitMaker","FST raw hit maker", 
-                                                                                                    kFALSE},
-  {"fstCluster", "", "fstChain", "fstRawHit","StFstClusterMaker",  "StFstClusterMaker","FST Cluster maker", 
-                                                                                                    kFALSE},
-  {"fstHit",     "", "fstChain", "event,fstCluster", "StFstHitMaker",       "StFstHitMaker","FST Hit Maker",
-                                                                                                    kFALSE},
-  {"fstUtil"  , "", "fstChain", "",          "",       "StFstUtil",    "Fst Utilities",                     
-                                                                                                    kFALSE},
```
plexoos, line 1601: Indeed these lines appear to be identical to the ones above them. These were approved twice in #266 and #282 about 3 and 2 months ago. So much for the huge PRs

genevb (approved):
I certainly approve of removing the duplication in BigFullChain.h.

StarDb/Geometry/fst/fstSensorOnWedge.20211110.000001.C
techuan-huang: I'm not sure if somewhere else will be affected, but for what I tested, the problem is in StRoot/StFstHitMaker/StFstHitMaker.cxx:
The database is loaded with StFstDb and the table is assigned to a variable on line 44.
This variable is then used on line 192 for transforming the coordinates of the hit positions.
As defined on line 191, the sensorId should begin with 1000.

starsdong: Thank you, Techuan. I added Xu and Zhenyu here to make sure everyone is on the same page.
On a separate note, Xu and Zhenyu, as the PoC of FST are not listed as the code maintainers for StarDb/xxx/fst directory. I feel we may want to specify this, also for other subsystems, so the PRs can automatically pick up the reviewers.

sunxuhit (approved):
These changes are already discussed in FST group and verified locally.

StarDb/Geometry/fst/fstSensorOnWedge.20211110.000001.C
plexoos: > the sensorId should begin with 1000

Ok, I see that in the corresponding database table these sensor ids start from 1000, so this macro is supposed to follow the same numbering.

techuan-huang: Dmitri, thanks for checking.
Since there are already two reviewers approved, should we go ahead to merge this PR or we need to wait for your approval?

plexoos: The formal requirements are met, so merge it anytime

techuan-huang: Thanks, I will click the merge button.

PR #344, opened 2022-05-05
eTOF Geometry changes based on technical drawings

PhilippWeidenkaff:
a series of small geometry adjustments that could be seen in data and have been confirmed by technical drawings


plexoos (approved):
If it is possible to reference the technical drawings, please do so

klendathu2k (approved):
Updated positions for eTOF planes, modules.   No changes to structure, hits, etc...  Approved.

PR #314, opened 2022-02-28
TriggerSimu + picoDst for BEMC data in daq->picoDst production

starsdong:
- PicoDst EMC collection pointer taken from StEvent directly in case of daq production
- TriggerSimuMaker added to BFC chain
- EmcRawMaker used to obtain BEMC data in case of daq production

PR #346, opened 2022-05-06
Enable shared linking of mysqlclient lib to root4stat

plexoos:
An alternative solution to the issue described in PR #343

### DIFF
--- a/asps/rexe/Conscript
+++ b/asps/rexe/Conscript
@@ -79,9 +79,7 @@ if ($FC =~ /^pgf/) {
 }
 $LIBS .= " -lgeant321 -lgcalor"; 
 $LIBS  .= " " . $env->{CERNLIBS};
-$LIBS  .= $env->{LDALL};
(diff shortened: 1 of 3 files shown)


plexoos:
Probably I should have sent a reminder earlier... but let's merge and see if the nightly build and tests on the farm can digest this change. If not we'll roll back.

plexoos:
@amolhj I don't see anything wrong with the nightly tests after this merge. If you see otherwise please let me know

genevb:
This is a minor observation that requires no action, but just to document what I observed: This PR appears to have added ~4.5 MB of memory usage to all nightly test jobs after the 2022-05-11 AutoBuild of DEV.

plexoos:
Another observation is that the root4star executable size (on /afs) reduced from 33 MB to 31 MB. It is peculiar that root4star's size is 19 MB in the container.

There seems to be another small increase after 2022-05-16

The large step on or after 2022-03-06 seems to correspond to the changes in #314

PR #357, opened 2022-05-23
Help cons figure out location of dependencies

plexoos:


klendathu2k (approved):
Looks reasonable to me.  I presume that, if anyone needs to compile / test with a specific fast jet or gsl or ... library, they can just set it up in their environment, and then the config scripts will do the rest.

plexoos:
Yes, when additional versions are available one can switch between them by adding a corresponding config scripts to PATH

genevb (commented):
No objections.

PR #358, opened 2022-05-30
genDst adaption for eTOF

PhilippWeidenkaff:
Contains:

-Implementation of eTOF makers in genDst script
-Adjustments to eTOF calib-, hit-, and matchmaker to check for muDsts when empty StEvents are found (needed to run in genDst.C due to EMC makers)
-Database interface for alignment parameters in Geometry class and Matchmaker


StRoot/StETofCalibMaker/StETofCalibMaker.cxx
```diff
@@ -1100,15 +1100,32 @@ StETofCalibMaker::Make()
 {
     LOG_DEBUG << "StETofCalibMaker::Make(): starting ..." << endm;
 
-    mEvent = ( StEvent* ) GetInputDS( "StEvent" );
+    //mEvent = ( StEvent* ) GetInputDS( "StEvent" );
+    mEvent = NULL; //don't check for StEvent for genDst.C testing. PW
```
genevb: I expect the commented line should be switched here for the commit.

genevb (changes requested):
The changes to genDst.C look fine to me.

-Gene

StRoot/StETofUtil/StETofGeometry.cxx
```diff
@@ -578,18 +643,18 @@ void
 StETofGeometry::init( TGeoManager* geoManager )
 {
     double safetyMargins[ 2 ] = { 0., 0. };
-    init( geoManager, safetyMargins, false );
+    init( NULL, geoManager, safetyMargins, false );
```
plexoos: In C++ we use `nullptr`

StRoot/StETofUtil/StETofGeometry.cxx
```diff
@@ -578,18 +643,18 @@ void
...
 }
 
 void
 StETofGeometry::init( TGeoManager* geoManager, const double* safetyMargins )
 {
-    init( geoManager, safetyMargins, false );
+    init( NULL, geoManager, safetyMargins, false );
```
plexoos: In C++ we use `nullptr`

plexoos: In C++ we use `nullptr`

StRoot/StETofUtil/StETofGeometry.cxx
```diff
@@ -600,23 +665,30 @@ StETofGeometry::init( TGeoManager* geoManager, const double* safetyMargins, cons
 
     mNValidModules = 0;
 
+
+	 if( mFileNameAlignParam != ""){
```
plexoos:
```suggestion
    if ( !mFileNameAlignParam.empty() ) {
```

StRoot/StETofUtil/StETofGeometry.cxx
```diff
@@ -600,23 +665,30 @@ StETofGeometry::init( TGeoManager* geoManager, const double* safetyMargins, cons
...
             std::string geoPath( formTGeoPath( geoManager, plane, sector ) );
 
             if( geoPath.empty() ) {
-                LOG_DEBUG << "StETofGeometry::Init(...) - cannot find path to ETOF module "
+                LOG_INFO << "StETofGeometry::Init(...) - cannot find path to ETOF module "
                             "(id " << plane << sector << "). Skipping..." << endm;
                 continue;
             }
             mNValidModules++;
-
```
plexoos: Please don't make so many changes in white space. This is completely irrelevant to the primary goal of this PR

StRoot/StETofUtil/StETofGeometry.cxx
```diff
@@ -832,6 +903,21 @@ StETofGeometry::findETofNode( const int moduleId, const int counterId )
...
 
+void
+StETofGeometry::pointMaster2local( const int moduleId, const int counterId, const double* master,  double* local )
+{
+    local[ 0 ] = 0;
+    local[ 1 ] = 0;
+    local[ 2 ] = 0;
+
+    if( !findETofNode( moduleId, counterId ) ) {
+        LOG_ERROR << "ETOF volume of a hit is not loaded in the geometry" << endl;
```
plexoos:
```suggestion
        LOG_ERROR << "ETOF volume of a hit is not loaded in the geometry" << endm;
```

StRoot/StETofUtil/StETofGeometry.cxx
```diff
@@ -1488,4 +1572,82 @@ StETofGeometry::getFieldZ( const double& x, const double& y, const double& z ) {
     mStarBField->BField( X, B );
 
     return B[ 2 ];
-}
\ No newline at end of file
+}
+
+void 
+StETofGeometry::readAlignmentParameters(){
+      LOG_INFO << "etofAlignParam: filename provided --> use parameter file: " << mFileNameAlignParam.c_str() << endm;
```
plexoos: std::string can be streamed. Remove c_str()

StRoot/StETofUtil/StETofGeometry.cxx
```diff
@@ -1488,4 +1572,82 @@ StETofGeometry::getFieldZ( const double& x, const double& y, const double& z ) {
...
+        
+		  //add check for no alignment set here.
+	
+		  ifstream paramFile;		
+        paramFile.open( mFileNameAlignParam.c_str() );
```
plexoos: std::ifstream.open accepts std::string. Remove c_str

StRoot/StETofUtil/StETofGeometry.cxx
```diff
@@ -150,7 +192,7 @@ StETofNode::calcXYPlaneNormal()
     // calculate the normal vector to the local XY-plane
     // i.e. the global representation of the local unit vector (0,0,1)
 
-    double xl[ 3 ] = { 0, 0, 1 };
+    double xl[ 3 ] = { 0, 0, 1 }; //change to x = 0.0227. PW.
```
plexoos: Confusing comment. Elaborate or remove completely

StRoot/StETofUtil/StETofGeometry.cxx
```diff
@@ -60,16 +69,7 @@ StETofNode::StETofNode( const TGeoPhysicalNode& gpNode )
...
-    /*
-    double* trans  = mGeoMatrix->GetTranslation();
-    double* rot    = mGeoMatrix->GetRotationMatrix();
 
-    LOG_INFO << trans[0] << "  " << trans[1] << "  " << trans[2] << endm;
-    
-    LOG_INFO << rot[0] << "  " << rot[1] << "  " << rot[2] << endm;
-    LOG_INFO << rot[3] << "  " << rot[4] << "  " << rot[5] << endm;
-    LOG_INFO << rot[6] << "  " << rot[7] << "  " << rot[8] << endm;
-    */
```
plexoos, line 72: Remove the following commented code as well https://github.com/star-bnl/star-sw/pull/358/files#diff-fa0ea941010f867b3171cb872336b971cf270aaa392bf777c57d206147fdc9d0R84-R93

StRoot/StETofUtil/StETofGeometry.cxx
```diff
@@ -30,12 +30,19 @@
...
 
 #include "TGeoManager.h"
+#include "TGeoNavigator.h"
 #include "TGeoVolume.h" 
 #include "TGeoPhysicalNode.h"
 #include "TGeoMatrix.h"
 #include "TGeoBBox.h"
+#include "TRotation.h"
+#include "TMath.h"
+#include "TFile.h"
```
plexoos: Not used. Remove

StRoot/StETofUtil/StETofGeometry.cxx
```diff
@@ -30,12 +30,19 @@
  *
  ***************************************************************************/
 #include <algorithm>
+#include <fstream>
+#include <sstream>
+#include <string>
 
 #include "TGeoManager.h"
+#include "TGeoNavigator.h"
```
plexoos: Not used. Remove

StRoot/StETofUtil/StETofGeometry.cxx
```diff
@@ -30,12 +30,19 @@
  *
  ***************************************************************************/
 #include <algorithm>
+#include <fstream>
+#include <sstream>
+#include <string>
```
plexoos: Are these used? Remove if not

StRoot/StETofUtil/StETofGeometry.cxx
```diff
@@ -95,11 +95,53 @@ StETofNode::StETofNode( const TGeoPhysicalNode& gpNode, const float& dx, const f
...
+  mSafetyMarginY( 0. ), 
+  mDebug( false )
+{
+
+    mGeoMatrix = static_cast< TGeoHMatrix* > ( gpNode.GetMatrix() );
+
+    mBox = static_cast< TGeoBBox* > ( gpNode.GetShape() );
+
+
+
```
plexoos: Sometimes you remove existing empty lines, sometimes you add new ones...

StRoot/StETofUtil/StETofGeometry.cxx
```diff
@@ -95,11 +95,53 @@ StETofNode::StETofNode( const TGeoPhysicalNode& gpNode, const float& dx, const f
...
+    mEtaMin = calcEta( -1. );
+    mEtaMax = calcEta(  1. );
+
+    mPhiMin = calcPhi( -1, -1. );
+    mPhiMax = calcPhi( -1,  1. );
+
+    if( mDebug ) print();
+
+//    buildMembers();
```
plexoos:
```suggestion
```

StRoot/StETofUtil/StETofGeometry.cxx
```diff
@@ -600,23 +665,30 @@ StETofGeometry::init( TGeoManager* geoManager, const double* safetyMargins, cons
...
     mNValidModules = 0;
 
+
+	 if( mFileNameAlignParam != ""){
+	 	readAlignmentParameters();
+	 }else{
+		readAlignmentDatabase(maker);
+	 }
+
+	 geoManager->AddNavigator();
```
plexoos: What is the purpose of this line?

StRoot/StETofUtil/StETofGeometry.cxx
```diff
@@ -1488,4 +1572,82 @@ StETofGeometry::getFieldZ( const double& x, const double& y, const double& z ) {
...
+            mAlignmentParameters.push_back( counterAlignmentParameter );
+        }
+
+        if( mAlignmentParameters.size() != 108 ) {
+            LOG_WARN << "parameter vector for alignments has not the right amount of entries: ";
+            LOG_WARN << mAlignmentParameters.size() << " instead of 108 !!!!" << endm;
+            return;
+        }else{
+				return;
+		 }
```
plexoos: mAlignmentParameters is a private member, you just filled it so, why check the vector size? Don't you trust yourself? Also, is eTofConst::nCounters equal to 108? Why use the literal then?

StRoot/StETofUtil/StETofGeometry.cxx
```diff
@@ -1488,4 +1572,82 @@ StETofGeometry::getFieldZ( const double& x, const double& y, const double& z ) {
...
+		  //add check for no alignment set here.
+	
+    	  TDataSet* dbDataSet = nullptr;
+
+		 if(!maker){
+			LOG_WARN << "No valid pointer to any StMaker given. Do not load alignment from database" << endm;
+			return;
+		 }
+
+        dbDataSet = maker->GetDataBase( "Geometry / etof / etofAlign" );
```
plexoos: I don't think you need a instance of a maker to access a data set. See if there is a static method

plexoos (changes requested):
Changes in StarVMC/Geometry/EtofGeo/EtofGeo0.xml should be submitted separately

StRoot/StETofCalibMaker/StETofCalibMaker.cxx
PhilippWeidenkaff: Corrected by the commit below

StRoot/StETofUtil/StETofGeometry.cxx
PhilippWeidenkaff: Latest commit should address all your comments. Thanks for the review.

plexoos: I don't see any new commits on this branch. Please make sure you actually push your changes. Thanks

StRoot/StETofUtil/StETofGeometry.cxx
PhilippWeidenkaff: fixed in latest commit

PhilippWeidenkaff: fixed in latest commit

StRoot/StETofUtil/StETofGeometry.cxx
PhilippWeidenkaff: fixed in latest commit

StRoot/StETofCalibMaker/StETofCalibMaker.cxx
PhilippWeidenkaff: Note, that this change is from an intermediate commit. Final version has the proper line update

PhilippWeidenkaff: Sorry, my mistake. didn't see it correctly. Now solved with latest commit

StRoot/StETofUtil/StETofGeometry.cxx
PhilippWeidenkaff: Noted :). Will try to avoid in the future.

StRoot/StETofMatchMaker/StETofMatchMaker.cxx
```diff
@@ -394,13 +394,31 @@ StETofMatchMaker::Make()
...
+
+		     if( mMuDst ) {
+             LOG_DEBUG << "Make() - running on MuDsts" << endm;
+
+             mIsMuDstIn = true;
+
+             fillIndexToPrimaryMap();
+
+             cleanUpTraits();
+		     }
```
plexoos: You don't have an else statement. What is supposed to happen when there is neither `etofCollection` nor `mMuDst`?

StRoot/StETofUtil/StETofGeometry.cxx
```diff
@@ -81,25 +75,54 @@ StETofNode::StETofNode( const TGeoPhysicalNode& gpNode, const float& dx, const f
...
-    LOG_INFO << rot[0] << "  " << rot[1] << "  " << rot[2] << endm;
-    LOG_INFO << rot[3] << "  " << rot[4] << "  " << rot[5] << endm;
-    LOG_INFO << rot[6] << "  " << rot[7] << "  " << rot[8] << endm;
-    */
     mBox = static_cast< TGeoBBox* > ( gpNode.GetShape() );
 
     // resize mBox with dx and dy
     float dz = mBox->GetDZ();
-    mBox->SetBoxDimensions( dx, dy, dz );
+    mBox->SetBoxDimensions( dx, dy, dz );
```
plexoos:
```suggestion
    mBox->SetBoxDimensions( dx, dy, dz );
```

StRoot/StETofUtil/StETofGeometry.cxx
```diff
@@ -81,25 +75,54 @@ StETofNode::StETofNode( const TGeoPhysicalNode& gpNode, const float& dx, const f
...
+    mBox = static_cast< TGeoBBox* > ( gpNode.GetShape() );
+
+    // resize mBox with dx and dy
+    float dz = mBox->GetDZ();
+    mBox->SetBoxDimensions( dx, dy, dz ); 
+
+    //  modified  buildMembers() method. PW;
+    // build member variables: mMinEta, mMaxEta, mMinPhi, mMaxPhi, mCenter, mNormal 
+    double xl[3] = { -alignment.x(), -alignment.y(), -alignment.z() };
+//z alignment is positive to reflect the correct direction in which z has to be moved in global coordinates. x and y are negative to match the numbers from hit-intersection
```
plexoos:
```suggestion
    //z alignment is positive to reflect the correct direction in which z has to be moved in global coordinates. x and y are negative to match the numbers from hit-intersection 
```

StRoot/StETofUtil/StETofGeometry.cxx
plexoos: If it was fixed the diff wouldn't show the original line. To accept the suggestion just click on the "Commit suggestion" button

StRoot/StETofUtil/StETofGeometry.cxx
```diff
@@ -1488,4 +1544,76 @@ StETofGeometry::getFieldZ( const double& x, const double& y, const double& z ) {
...
+
+		  for( int iCounter = 0; iCounter < eTofConst::nCounters; iCounter++){
+       		tempX = table[iCounter].detectorAlignX[0];
+       		tempY = table[iCounter].detectorAlignY[0];
+       		tempZ = table[iCounter].detectorAlignZ[0];
+		  		StThreeVectorD counterAlignmentParameter = StThreeVectorD( tempX, tempY, tempZ );
+            mAlignmentParameters.push_back( counterAlignmentParameter );
+        }
+
+		return;
```
plexoos:
```suggestion
```

StRoot/StETofUtil/StETofGeometry.cxx
```diff
@@ -600,23 +637,30 @@ StETofGeometry::init( TGeoManager* geoManager, const double* safetyMargins, cons
...
     mNValidModules = 0;
 
+
+	 if( mFileNameAlignParam != ""){
+	 	readAlignmentParameters();
+	 }else{
+		readAlignmentDatabase(maker);
+	 }
+
+	 //geoManager->AddNavigator();
```
plexoos:
```suggestion
```

StRoot/StETofUtil/StETofGeometry.cxx
```diff
@@ -1488,4 +1544,76 @@ StETofGeometry::getFieldZ( const double& x, const double& y, const double& z ) {
...
+		  //add check for no alignment set here.
+	
+    	  TDataSet* dbDataSet = nullptr;
+
+		 if(!maker){
+			LOG_WARN << "No valid pointer to any StMaker given. Do not load alignment from database" << endm;
+			return;
+		 }
+
+        dbDataSet = maker->GetDataBase( "Geometry / etof / etofAlign" );
```
plexoos: As I mentioned earlier you don't need to pass around the pointer to the Maker. The following should work

```suggestion
        dbDataSet = StMaker::GetChain()->GetDataBase("Geometry/etof/etofAlign");
```

plexoos (commented):
Thanks for addressing my previous comments Philipp. It looks like you were able to get rid of the unintended change in xml file by merging in the main branch.

StRoot/StETofMatchMaker/StETofMatchMaker.cxx
PhilippWeidenkaff: Nothing. If there is neither a (filled) StEvent nor an MuDst, Make() should not do anything.

StRoot/StETofUtil/StETofGeometry.cxx
```diff
@@ -1488,4 +1542,74 @@ StETofGeometry::getFieldZ( const double& x, const double& y, const double& z ) {
...
+        if( mAlignmentParameters.size() != 108 ) {
+            LOG_INFO << "parameter file for alignments has not the right amount of entries: ";
+            LOG_INFO << mAlignmentParameters.size() << " instead of 108 !!!!" << endm;
+            return;
+        }
+
+}
+
+void 
+StETofGeometry::readAlignmentDatabase(StMaker* maker){
```
plexoos: It seems the argument is not used by this function anymore

StRoot/StETofUtil/StETofGeometry.cxx
```diff
@@ -100,6 +84,45 @@ StETofNode::StETofNode( const TGeoPhysicalNode& gpNode, const float& dx, const f
     buildMembers();
 }
 
+StETofNode::StETofNode( const TGeoPhysicalNode& gpNode, const float& dx, const float& dy, const StThreeVectorD alignment )
```
plexoos: It does not make much sense to pass dx and dy by reference

StRoot/StETofUtil/StETofGeometry.cxx
```diff
@@ -483,6 +506,19 @@ StETofGeomCounter::StETofGeomCounter( const TGeoPhysicalNode& gpNode, const floa
...
+  mCounterIndex( counterId ),
+  mDebug( false )
+{
+    mSector = calcSector( moduleId  );
+    mPlane  = calcPlane( moduleId );
+
+    createGeomStrips();
+
+    if( mDebug ) print();
+}
```
plexoos: This new constructor is practically identical to the one above it. Would it make sense to avoid the boilerplate code and modify the existing one instead by adding the alignment parameter with default values {0, 0, 0}?

StRoot/StETofUtil/StETofGeometry.cxx
PhilippWeidenkaff: Removed with the removal of the StMaker-Pointer to load database

StRoot/StETofUtil/StETofGeometry.cxx
PhilippWeidenkaff: Removed with the removal of the StMaker-Pointer to load database

StRoot/StETofUtil/StETofGeometry.cxx
PhilippWeidenkaff: Fixed. Thanks for finding the proper way of doing it

StRoot/StETofUtil/StETofGeometry.cxx
PhilippWeidenkaff: I removed this comment since it is not up to date anymore. All alignment numbers are passed as negatives, so the comment about z being positive doesn't make sense any more.

StRoot/StETofUtil/StETofGeometry.cxx
```diff
@@ -1488,4 +1542,69 @@ StETofGeometry::getFieldZ( const double& x, const double& y, const double& z ) {
...
+
+void 
+StETofGeometry::readAlignmentDatabase(){
+        LOG_INFO << "No etof alignment file provided --> use database: " << endm;
+        
+		  //add check for no alignment set here.
+	
+    	  TDataSet* dbDataSet = nullptr;
+
+        dbDataSet = StMaker::GetChain()->GetDataBase("Geometry/etof/etofAlign");
```
plexoos:
```suggestion
    	  TDataSet* dbDataSet = StMaker::GetChain()->GetDataBase("Geometry/etof/etofAlign");
```

StRoot/StETofUtil/StETofGeometry.h
```diff
@@ -47,6 +47,8 @@
 
 #include "StETofUtil/StETofConstants.h"
 
+#include "StMaker.h"
+
```
plexoos:
```suggestion
```

StRoot/StETofUtil/StETofGeometry.h
```diff
@@ -387,7 +399,7 @@ inline void  StETofGeometry::debugOn()  { mDebug = true;  }
 inline void  StETofGeometry::debugOff() { mDebug = false; }
 
 inline bool  StETofGeometry::isDebugOn() const { return mDebug; }
+inline void  StETofGeometry::setFileNameAlignParam(std::string FileNameAlignParam) {mFileNameAlignParam = FileNameAlignParam;}
```
plexoos:
```suggestion
inline void  StETofGeometry::setFileNameAlignParam(std::string fileNameAlignParam) {mFileNameAlignParam = fileNameAlignParam;}
```

https://star-bnl.github.io/star-coding-guidelines/formatguide.xml#Variable_Names

StRoot/StETofUtil/StETofGeometry.cxx
```diff
@@ -1488,4 +1542,69 @@ StETofGeometry::getFieldZ( const double& x, const double& y, const double& z ) {
...
+        }
+        
+        paramFile.close();
+
+        if( mAlignmentParameters.size() != 108 ) {
+            LOG_INFO << "parameter file for alignments has not the right amount of entries: ";
+            LOG_INFO << mAlignmentParameters.size() << " instead of 108 !!!!" << endm;
+            return;
+        }
+
```
plexoos: If the number of values read from the file is not 108 as expected it'd better be a warning

```suggestion
        if( mAlignmentParameters.size() != 108 ) {
            LOG_WARN << "parameter file for alignments has not the right amount of entries: "
                     << mAlignmentParameters.size() << " instead of 108 !!!!" << endm;
        }
```

StRoot/StETofCalibMaker/StETofCalibMaker.cxx
```diff
@@ -1343,11 +1360,11 @@ StETofCalibMaker::processMuDst()
 	}*/
 
     size_t nDigis = mMuDst->numberOfETofDigi();
-    //LOG_INFO << "processMuDst() - # fired eTOF digis : " << nDigis << endm;
+    LOG_INFO << "processMuDst() - # fired eTOF digis : " << nDigis << endm;
```
plexoos: You uncommented or added a bunch of printout statements. Are you suggesting to print all of them for every event? Our log files are already bloated. Preferably only warnings and errors should be logged.

StRoot/StETofUtil/StETofGeometry.cxx
```diff
@@ -395,6 +409,16 @@ StETofGeomModule::addCounter( const TGeoPhysicalNode& gpNode, const float& dx, c
...
+void
+StETofGeomModule::addCounter( const TGeoPhysicalNode& gpNode, const float& dx, const float& dy, const int moduleId, const int counterId, const double* safetyMargins, const StThreeVectorD alignment )
+{
+    StETofGeomCounter* counter = new StETofGeomCounter( gpNode, dx, dy, moduleId, counterId, alignment );
+
+    counter->setSafetyMargins( safetyMargins );
+
+    mETofCounter.push_back( counter );
+}
+
```
plexoos: This seems like a case where introducing a default argument is more appropriate. 

From C++ Style Guide https://google.github.io/styleguide/cppguide.html#Default_Arguments:

"Compared to overloading the function, default arguments have a cleaner syntax, with less boilerplate and a clearer distinction between 'required' and 'optional' arguments."

StRoot/StETofUtil/StETofGeometry.h
PhilippWeidenkaff: The include is needed for StMaker::GetChain(), which we now decided to use to call the database.

StRoot/StETofCalibMaker/StETofCalibMaker.cxx
PhilippWeidenkaff: No, i accidently pushed the wrong commit. All the debugging messages should be removed now. This exact line that you commented on though has historically always been in.

plexoos: The particular diff indicates that this line is currently commented out on the main branch but you intend to uncomment it. Can you elaborate on the value of having "# fired eTOF digis : " printed in the log file for every event? How is it going to be used? Can the same information be obtained from the output ROOT file?

StRoot/StETofUtil/StETofGeometry.cxx
PhilippWeidenkaff: Agreed. This is part of the core for this class that was written by Florian and it seems his logic seemed to be to pass just about everything as references. I also do not see it making a difference (even if someone were to change dx/dy in this method accidently, it would not matter, as these parameters are passed once and then never used again. ). I could go through the full file and change that consistently, but i see little benefit from it. What do you think?

StRoot/StETofUtil/StETofGeometry.cxx
PhilippWeidenkaff: Ok. Makes sense. Changed in latest commit

StRoot/StETofCalibMaker/StETofCalibMaker.cxx
PhilippWeidenkaff: I probably already deactived this line in a former print-out clean-up. "Historically", the idea was to check that the same number of data members are seen by different makers. This information could not be obtained for single events from the output file. On the other hand, probably nobody is ever going through production logs and check for single event level issues. That is a case for local debugging. If bloated log files are becoming a problem, then all event-by-event print-outs should go. I removed them from all relevant eTOF makers in the latest commit unless they are hidden behind doQA flags or debug flags.

StRoot/StETofUtil/StETofGeometry.cxx
```diff
@@ -586,10 +625,10 @@ StETofGeometry::init( TGeoManager* geoManager, const double* safetyMargins )
 {
     init( geoManager, safetyMargins, false );
 }
-
+*/
 
 void
-StETofGeometry::init( TGeoManager* geoManager, const double* safetyMargins, const bool& useHelixSwimmer )
+StETofGeometry::init( TGeoManager* geoManager, const double safetyMargins[2] = (double[2]){ 0, 0 }, const bool& useHelixSwimmer = false )
```
plexoos: This C-cast is quite ugly and I am not even sure if it is valid to use an initializer like this in C++. Why not simply use a `nullptr` or `std::array<double, 2>`?

StRoot/StETofUtil/StETofGeometry.cxx
plexoos: Although it is considered to be a good practice, I don't insist because practically this is unlikely to result in any noticeable difference. But if you do decide to make these changes please do that in a separate PR.

StRoot/StETofUtil/StETofGeometry.cxx
PhilippWeidenkaff: Agree on the uglyness, but it was the only way i got it working with a default :) . I want to keep the c-style array notation as that is what is provided with by all classes and macros that use this class. I could provide a null pointer default, but then i would have to check for the nullptr and set the default (0,0) in the method. I considered that even more ugly.

plexoos: You can keep the pointer to pass a C-array but in C++ you can't take an address of a temporary array. It likely compiles for you because you use gcc with GNU extensions. If you add the `-pedantic -Werror` options to gcc your code won't compile. On the other hand, taking an address of a temporary object is OK in this context. So, I already mentioned two options to deal with this, another one would be to use an `initializer_list<double>{0, 0}` to create a proxy object. Here is a link if you want to play with this code https://godbolt.org/z/czPrqnf9e

By the way, this looks ugly `const bool& useHelixSwimmer = false` 😄  Anyway, if it works and nobody else cares, keep it. I feel like I already exhausted the time I want to spend reviewing this PR. It would be nice to hear from at least another reviewer but if not let's merge it already.

PhilippWeidenkaff: Thanks for the example :)

YannickSoehngen (approved):

plexoos (approved):
Thank you Phillipp for addressing the comments!

genevb:
Unfortunately, merging this PR appears to have caused numerous nightly test job failures for year 2019+ datasets (and tagging SL22b may have been premature). Example from a nightly test log file /star/rcf/test/dev/daq_sl302.ittf/Thu/year_2020/production_11p5GeV_2020/st_physics_20351078_raw_5000001.log:

StETofHitMaker:INFO  - *** StETofHitMaker::Make() == StOK(0) ***
StETofMatchMaker:INFO  - StETofMatchMaker::InitRun()
StETofMatchMaker:INFO  - etofMatchParam: no filename provided --> load database table
StETofMatchMaker:INFO  -  match radius: 2
StETofMatchMaker:INFO  -  track cut (nHitsFit): 5
StETofMatchMaker:INFO  -  track cut (nHitsRatio): 0
StETofMatchMaker:INFO  -  track cut (low pt): 0.045
StETofMatchMaker:INFO  -  creating a new eTOF geometry . . .
StETofMatchMaker:INFO  -  eTOF geometry initialization ...
StETofMatchMaker:INFO  - No etof alignment file provided --> use database:
StChain:WARN  - StDbSql::QueryDb(table,time) line=439 etofAlign  has No data for query
StETofMatchMaker:FATAL - TUnixSystem::DispatchSignals : segmentation violation

How to reproduce in DEV:

    root4star -b -q -l 'bfc.C(2,"P2020a,btof,mtd,etofA,BEmcChkStat,picoVtxDefault,PicoCovMtxWrite,-hitfilt,QAalltrigs","/star/rcf/test/daq/2020/351/st_physics_20351078_raw_5000001.daq")'

Please have a look.

Thanks,
-Gene

genevb:
gdb says:

Program received signal SIGSEGV, Segmentation fault.
0xeb6cd819 in StETofGeometry::readAlignmentDatabase (this=0x15966d50) at .sl73_gcc485/OBJ/StRoot/StETofUtil/StETofGeometry.cxx:1530
1530	    St_etofAlign* etofAlign = static_cast< St_etofAlign * > ( dbDataSet->Find( "etofAlign" ) );
(gdb) p dbDataSet
$1 = (TDataSet *) 0x0

This appears to be the common case of not having a database table entry applicable for data for which the table isn't needed. The likely solutions are:

1. Place a default (ideal) entry in the database with beginTime from before all data for which this code may be run. We usually use "1996-01-01" as a default. It may also be necessary for the entryTime to be set (with the help of @dmarkh ) to "1996-01-01" to ensure old BFC chains for this data with this maker, but with a DbV, do not crash. This may not require a code update depending on what is supposed to happen for this case in the code.

2. Modify the code not to crash if no valid entry for the table is found.

However, there were other jobs that crashed too without the etofAlign error. Example:
/star/rcf/test/dev/daq_sl302.stica/Wed/year_2022/production_pp500_2022/st_physics_23010027_raw_1000015.log
which appears to have still crashed in the StETofMatchMaker:
StETofMatchMaker:FATAL - TUnixSystem::DispatchSignals : bus error

-Gene

PhilippWeidenkaff:
Hi Gene,

I'm uploaded a default alignment table. I also added an error-hand-down to the MatchMaker in a new pull request. This new pull request also contains a change in the default start time handling, which caused a 150 ps shift in the timing. It should still be included in the data reproduction!

I'm rather confused by the second error. It seems like StMuDstMaker throws this bus error first:

QAInfo: doPs for strangeMuDst: EndMaker total =1772.691406 heap =773.125565 and 175.093246( +5.760719)
StStrangeMuDstMaker:INFO - *** StStrangeMuDstMaker::Make() == StOK(0) ***
StMuDstMaker:INFO - StMuDSTMaker filling StMuFcsCollection from StEvent
StMuDstMaker:INFO - fillMuFtt
StMuDstMaker:INFO - fillMuFttRawHits
StMuDstMaker:INFO - fillMuFttPoints
StMuDstMaker:INFO - fillMuFttClusters
StMuDstMaker:FATAL - TUnixSystem::DispatchSignals : bus error
Function bfc() busy flag cleared
Function bfc() busy flag cleared

Then the chain starts shutting down and StEtofMatchMaker seems to crash after the MuDst is already written:

StMuDstMaker:INFO - ##### virtual void StMuDstMaker::closeWrite()
StMuDstMaker:INFO - ##### File=./st_physics_23010027_raw_1000015.MuDst.root NumberOfEvents= 661 #####
StStrangeMuDstMaker:INFO - StV0Controller: Total kept: 0 V0 candidates
StStrangeMuDstMaker:INFO - StXiController: Total kept: 0 Xi candidates
StStrangeMuDstMaker:INFO - StKinkController: Total kept: 0 Kink candidates
StETofMatchMaker:FATAL - TUnixSystem::DispatchSignals : bus error

The line directly above seems to come from StStrangeMuDstMaker::Finish(), so the chain should be calling the ::Finish() method of all makers now. StEtofMatchMaker::Finish() looks like this:

Int_t
StETofMatchMaker::Finish()
{
LOG_DEBUG << "StETofMatchMaker::Finish()" << endm;

if( mDoQA ) {
LOG_INFO << "Finish() - writing *.etofMatch.root ..." << endm;
setHistFileName();
writeHistograms();
}

if( mETofGeom ) {
delete mETofGeom;
mETofGeom = nullptr;
}

return kStOk;
}

Since mDoQa is false in this chain, literally the only thing that is happening is the deletion of the Geometry class. The destructor of the Geometry class should delete all relevant members correctly, but just to be sure i also added a clear() for the only new vector in there. However, any incorrectly deleted pointer in the destructor should thow an Segmentation fault, not a bus error. To be honest, i have no idea what a bus error is supposed to tell me.

Best Regards
Philipp
June 23, 2022 11:37 PM, "Gene Van Buren" ***@***.*** ***@***.******@***.***>)> wrote:
	gdb says: 

	Program received signal SIGSEGV, Segmentation fault.
0xeb6cd819 in StETofGeometry::readAlignmentDatabase (this=0x15966d50) at .sl73_gcc485/OBJ/StRoot/StETofUtil/StETofGeometry.cxx:1530
1530 St_etofAlign* etofAlign = static_cast< St_etofAlign * > ( dbDataSet->Find( "etofAlign" ) );
(gdb) p dbDataSet
$1 = (TDataSet *) 0x0 

	This appears to be the common case of not having a database table entry applicable for data for which the table isn't needed. The likely solutions are: 
	* 

	Place a default (ideal) entry in the database with beginTime from before all data for which this code may be run. We usually use "1996-01-01" as a default. It may also be necessary for the entryTime to be set (with the help of @dmarkh (https://github.com/dmarkh) ) to "1996-01-01" to ensure old BFC chains for this data with this maker, but with a DbV, do not crash. This may not require a code update depending on what is supposed to happen for this case in the code. 
	* 

	Modify the code not to crash if no valid entry for the table is found. 
	However, there were other jobs that crashed too without the etofAlign error. Example:
/star/rcf/test/dev/daq_sl302.stica/Wed/year_2022/production_pp500_2022/st_physics_23010027_raw_1000015.log
which appears to have still crashed in the StETofMatchMaker:
StETofMatchMaker:FATAL - TUnixSystem::DispatchSignals : bus error 

	-Gene

plexoos:
> Example from a nightly test log file /star/rcf/test/dev/daq_sl302.ittf/Thu/year_2020/production_11p5GeV_2020/st_physics_20351078_raw_5000001.log:
> 
> ```
> root4star -b -q -l 'bfc.C(2,"P2020a,btof,mtd,etofA,BEmcChkStat,picoVtxDefault,PicoCovMtxWrite,-hitfilt,QAalltrigs","/star/rcf/test/daq/2020/351/st_physics_20351078_raw_5000001.daq")'
> ```

I added the same test in https://github.com/star-bnl/star-sw/pull/371. It passed without crashing

Screen Shot 2022-06-25 at 10 25 58 AM

Here are the same lines from the log file (search for `StETofMatchMaker::InitRun`)

```
StETofMatchMaker:INFO  - StETofMatchMaker::InitRun()
StETofMatchMaker:INFO  - etofMatchParam: no filename provided --> load database table
StETofMatchMaker:INFO  -  match radius: 2
StETofMatchMaker:INFO  -  track cut (nHitsFit): 5
StETofMatchMaker:INFO  -  track cut (nHitsRatio): 0
StETofMatchMaker:INFO  -  track cut (low pt): 0.045
StETofMatchMaker:INFO  -  creating a new eTOF geometry . . . 
StETofMatchMaker:INFO  -  eTOF geometry initialization ... 
StETofMatchMaker:INFO  - No etof alignment file provided --> use database: 
StETofMatchMaker:INFO  - amount of valid modules: 36
StETofMatchMaker:INFO  - *** StETofMatchMaker::Make() == StOK(0) ***
```

The `etofAlign` table seems to be in the DB, and it contains some entries.
Maybe a full rebuild of the libraries can help...  Not sure what else to suggest

genevb:
Hi, Philipp

I see the default database entry is probably working now. However, I noticed that the default entry will be used for the first month of Run 20, which started in early December of 2019,  because the entry you had in there before had a timestamp of 2020-01-01. If the default entry has the same content as the 2020-01-01 entry, then that is probably of no consequence. But if they are different, that may not be as desired.

Thanks,
-Gene

PhilippWeidenkaff:
Hi Gene,

Just to be sure, i set up a new database entry on 25.11.2019, in case we ever change the run19 alginment again. I also put up a new pull request, which i would like to still see included in the production library.

Best regards
Philipp

June 25, 2022 10:06 PM, "Gene Van Buren" ***@***.*** ***@***.******@***.***>)> wrote:
	Hi, Philipp 

	I see the default database entry is probably working now. However, I noticed that the default entry will be used for the first month of Run 20, which started in early December of 2019, because the entry you had in there before had a timestamp of 2020-01-01. If the default entry has the same content as the 2020-01-01 entry, then that is probably of no consequence. But if they are different, that may not be as desired. 

	Thanks,
-Gene

PR #361, opened 2022-06-08
Refactor CI workflow files 

plexoos:
The idea behind this is to easily aggregate additional parameters (e.g. compiler) defining the environment

PR #362, opened 2022-06-08
Bump starenv to v0.1.6 with compiler as build argument

plexoos:
Needs #361

### DIFF
--- a/.github/workflows/build-containers.yml
+++ b/.github/workflows/build-containers.yml
@@ -15,8 +15,9 @@ jobs:
     strategy:
       matrix:
         starenv: [root5, root6]
+        compiler: [gcc485]
     env:
-      STARENV: ${{ matrix.starenv }}
+      STARENV: ${{ matrix.starenv }}-${{ matrix.compiler }}
     steps:
       - name: Get branch name
         id: branch-name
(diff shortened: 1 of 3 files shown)


.github/workflows/build-containers.yml
```diff
@@ -37,7 +37,7 @@ jobs:
           cache-to:   type=registry,ref=ghcr.io/${{ github.repository_owner }}/star-sw-${{ matrix.starenv }}:cache,mode=max
           build-args: starenv=${{ matrix.starenv }}
           push: true
-          tags: ghcr.io/${{ github.repository_owner }}/star-sw-${{ matrix.starenv }}:${{ steps.branch-name.outputs.current_branch }}_latest
+          tags: ghcr.io/${{ github.repository_owner }}/star-sw-${{ matrix.starenv }}:${{ steps.branch-name.outputs.current_branch }}
```
veprbl: I can't remember why the `_latest` suffix came around, but I'd guess it is there for a reason. Could be there because we used to create tag matching the branch name `SLXXy` instead of a different name `SLXXy_0`. Also it's nice to be able to easily distinguish tags that are "permanent" references and the branches for which things might be in flux.

plexoos: You remember it right. The `_latest` suffix indeed helped to distinguish between conflicting branch and tag names. But we are not using the same name for branches and tags anymore because it is very confusing. The tags now have an underscore in their name just like you mentioned. It is better to fix this notation now rather than continue to support the bad choices from the CVS times.

veprbl: Right, if the names don't clash, there should not be any technical issue. I think, the `_latest` is quite fine, but it's not a strong preference.

plexoos: As you may have noticed, my intention is to move the label from the package name to the image tag. In the future we may introduce a more conventional `:latest` tag

PR #363, opened 2022-06-08
ci: don't push containers for :SLYYx, only :SLYYx_z

veprbl:
This is a follow up to discussion in https://github.com/star-bnl/star-sw/pull/362#discussion_r892492034.

The SLYYx branches may be modified without a notice. Users should use SLYYx_z tags to get reproducible results. The :main label is left for those folks who like to use stardev.


veprbl:
Let's table this for now

PR #364, opened 2022-06-08
ci: avoid a duplicate step in build-containers.yml

veprbl:
Originally, the "latest" containers for a branch were supposed to
provide a cache for the build environment for the PRs and new
tags. Since the switch to the new star-spack infrastructure the
environment is now built elsewhere, the only target left in the local
dockerfile is to build the star-sw itself.

### DIFF
--- a/.github/workflows/build-containers.yml
+++ b/.github/workflows/build-containers.yml
@@ -32,20 +32,9 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Build latest container for the branch
+      - name: Build a release container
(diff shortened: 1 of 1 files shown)


veprbl:
I suppose the `${{ steps.branch-name.outputs.current_branch }}` should return a tag name for tag pushes. The https://github.com/tj-actions/branch-names/blob/main/README.md is a bit unclear on this, but it must have worked for us somehow.

plexoos (approved):
I confirmed in my fork that `${{ steps.branch-name.outputs.current_branch }}` produces a proper image tag even when the workflow is triggered by a push of a Git tag

PR #308, opened 2022-02-17
[g4star] Miscellaneous updates to StarAgmlLib

klendathu2k:
Merging StarAgmlLib which are independent of changes to
the StGeant4Maker development.

Brings code up to commit
71921c98f22fd5f8378ee394b1e546875743a3b2

The changes represented in the next commit appears to have already been integrated.
7dcbd58301d1e9a4eb9611256491c5b981069070

Beyond that commit (starting with 4c5edb1d9891627c6f0f56e6a1d1d4f52cc27646),
changes to StarAgmlLib are coupled to changes introduced in the StGeant4Maker/
directory.

- Agml modules extended with hit scoring

PR #366, opened 2022-06-11
Fix errrors reported by GCC 11.2.1

plexoos:


genevb (commented):
I understand memcpy to have no defined error return values, just returning the destination pointer argument that is passed to it. So it doesn't sound consequential to me whether the return value is tested to be negative vs. NULL, and whatever satisfies the computer is fine.

plexoos:
Sorry, I should have marked this PR as draft, of course, to avoid noisy notifications before proposing any final changes... I am just trying to see if the newer compiler can be satisfied with quick fixes.

StRoot/StFgtPool/StFgtClusterTools/StFgtGeneralBase.cxx
```diff
@@ -565,7 +565,7 @@ Int_t StFgtGeneralBase::fillFromMuDst(StFgtCollection& fgtCollection)
 		  {
 		    //construct new cluster
 		    oldClusIdx=clusIdx;
-		    if(pFgtHit>0)
+		    if(!pFgtHit)
```
klendathu2k: Logic is inverted here... the intent is to do...

if ( pFgtHit ) {
  fill the collection
}

StRoot/StFgtPool/StFgtClusterTools/StFgtGeneralBase.cxx
```diff
@@ -636,7 +636,7 @@ Int_t StFgtGeneralBase::fillFromMuDst(StFgtCollection& fgtCollection)
 	      }
 	  }
 	//put last cluster in...
-	if(pFgtHit>0)
+	if(!pFgtHit)
```
klendathu2k: Logic inverted

mgr/ConsDefs.pm
```diff
@@ -156,7 +156,7 @@
 	$G77FLAGS .= " -std=legacy -fno-second-underscore -w -fno-automatic -Wall -W -Wsurprising -fPIC";
 	
 	$FFLAGS    = $G77FLAGS;     # will be overwritten below, ignore
-	$FLIBS     = "-lgfortran";
+	$FLIBS     = "-lgfortran -lquadmath";
```
klendathu2k, line 159: Could you provide more context for including the libquadmath in the FORtran linker?  Is there some warning during FORtran compilation / linking that this solves?  

My concern is that this could, depending on the underlying math libraries, change precision for floating point computations.  It is unlikely to have a negative effect... but if so, it would be helpful to have the context as a starting point for debugging.

klendathu2k (changes requested):
Generally everything looks good.  With the FGT code, I think the logic is inverted in a couple of places. 

I would like to ask for some additional information on why the libquadmath library needs ti be linked on the FORtran side.  I'm nervous with any compiler option that could change the floating point precision (for better or worse)... we stumbled upon a 30+ year old GEANT bug that took months to track down over such a change.

mgr/ConsDefs.pm
plexoos: I think there is nothing to worry about, at least right now. The change is triggered by this error message during linking of `root4star` which clearly suggests to add `libquadmath` to the command line:

    /opt/rh/devtoolset-11/root/usr/libexec/gcc/x86_64-redhat-linux/11/ld: /opt/rh/devtoolset-11/root/usr/lib/gcc/x86_64-redhat-linux/11/libgfortran_nonshared.a(nonshared-trigd.o): undefined reference to symbol 'fmaq@@QUADMATH_1.0'
    /opt/rh/devtoolset-11/root/usr/libexec/gcc/x86_64-redhat-linux/11/ld: /lib64/libquadmath.so.0: error adding symbols: DSO missing from command line
    collect2: error: ld returned 1 exit status
    cons: *** [.sl79_gcc11/OBJ/asps/rexe/root4star] Error 1
    cons: errors constructing .sl79_gcc11/OBJ/asps/rexe/root4star

`libquadmath` is not new to us. It is already being used by geant3 and is linked to `root4star`. I believe gcc 4.8.5 is linking it without an explicit request but this is not uncommon with standard libraries shipped with compilers.

StRoot/StFgtPool/StFgtClusterTools/StFgtGeneralBase.cxx
plexoos: Makes perfect sense. Thanks!

StarVMC/StarAgmlLib/AgMath.h
```diff
@@ -39,7 +39,6 @@ const Double_t keV    = 0.001 * MeV;
 #define dup _dup_
 
 #define tanf _tanf_
-#define fpos _fpos_
```
klendathu2k, line 42: The point of this definition is to hide the "fpos" function, so that it doesn't interfere with the declaration of the FPOS volume in the FMS geometry.  It was always a bit of a hack.

If the compiler doesn't see a conflict building the geometry, then I am happy with this change.

plexoos: Hopefully, it won't cause a problem during runtime when the geometry is loaded. Although, what you propose in #308 is still worrisome and may not play well with modern compilers
https://github.com/star-bnl/star-sw/pull/308/files/2fb56841b5cafae4a95e587a5f65acaf01428d3e#r882221155

veprbl:
Also, I think we probably will need to plug a `-fcommon` to the CFLAGS if there are runtime issues, as the `-fno-common` is the default starting gcc 10 (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=85678).

klendathu2k:
I don't see why StBiTree was ever introduced.  Going back to the last fix to StBiTree, there were no codes which depended on it.  Going forward to SL10k, SL11a, SL12a... still no codes depended on it.  So.... safe to remove.

StarVMC/StarAgmlLib/AgMath.h
klendathu2k: The FMS geometry contains a data structure named FPOS.  Under earlier versions of AgML, this resulted in an object named "fpos" being declared... causing a compiler error.  Later releases of AgML move this object into the namespace of the geometry module.  So... if it compiles without the define, we will be good.

plexoos:
Rebased this PR branch before final merge. No new changes since previous reviews except incorporating changes suggested by @klendathu2k in #370

PR #370, opened 2022-06-23
ci: Enable tests for ROOT5 GCC11 environment

plexoos:
This PR branch is based on and includes changes from currently unmerged #366


jml985 (approved):
As discussed these changes are fine with me and do not cause issues.

StRoot/StarRoot/StBiTree.cxx
```diff
@@ -1,141 +0,0 @@
-#include "StBiTree.h"
```
klendathu2k: Removing StBiTree should be fine.  A quick grep does not show it being used in our current software stack.  It was first introduced around 2010, and I cannot find any codes in that time frame which depended on it.

StarVMC/StarAgmlLib/AgMath.h
```diff
@@ -39,7 +39,6 @@ const Double_t keV    = 0.001 * MeV;
 #define dup _dup_
 
 #define tanf _tanf_
-#define fpos _fpos_
```
klendathu2k: Adding historical context:

The FMS geometry contains a data structure named "FPOS".   Under the original implementation of AgML, this resulted in an object named "fpos" being created inside of FpdmGeo4.cxx, which conflicted with a cmath function.  Later revisions of AgML would move the objects which handle the data structures inside of a namespace, removing the conflict.

asps/Simulation/starsim/Conscript
```diff
@@ -90,7 +90,7 @@ $CPPPATH .= $main::PATH_SEPARATOR . "#asps/Simulation/starsim/include";
 $CPPPATH .= $main::PATH_SEPARATOR . $env->{Packages}->{MYSQL}->{INCDIR}; 
 $CPPPATH .= $main::PATH_SEPARATOR . $env->{ENV}->{CERN_ROOT} . "/include";
 $CPPPATH .= $main::PATH_SEPARATOR . $env->{ENV}->{CERN_ROOT} . "/src/pawlib/paw/ntuple";
-$CPPFLAGS  = "-DCERNLIB_DZDOC -DCERNLIB_NONEWL -DCERNLIB_SHL -DCERNLIB_HADRON -DCERNLIB_BSLASH -DCERNLIB_HIGZ -DCERNLIB_CG";
+$CPPFLAGS  = "-fcommon -DCERNLIB_DZDOC -DCERNLIB_NONEWL -DCERNLIB_SHL -DCERNLIB_HADRON -DCERNLIB_BSLASH -DCERNLIB_HIGZ -DCERNLIB_CG";
```
veprbl: `CPP` is the C Pre-Processor. This has to go to `CFLAGS` and, possibly, also `FFLAGS`.

StarVMC/StarAgmlLib/AgMath.h
```diff
@@ -54,7 +53,6 @@ Int_t nint(Float_t x);
 // Fortran abs is fabs
 //
 //#define abs(x) TMath::Abs(x)
-inline double abs (double x) {return fabs(x);}
```
klendathu2k: This line should be replaced with `using std::abs;`. 

---

TMath.h (which is included) includes <cmath>.  The version of abs that we want to be called here (which has overloads for FP arguments) is in namespace std.

The version which we absolutely don't want... (https://cplusplus.com/reference/cstdlib/abs/) will happily coerce this into a integer evaluation, resulting in abs(-1.23) == 1...

klendathu2k (changes requested):
I don't have any issues with the changes, save for one.  Add the line `using std::abs` in AgMath.h.  This will ensure that a floating point version of the abs function is used, rather than rely on whatever version gets included through TMath.h.    (Worst case, it uses the cstdlib version, which will happily cast to integer and break geometries.)

StRoot/StBFChain/BigFullChain.h
```diff
@@ -1352,7 +1352,7 @@ Bfc_st BFC[] = { // standard chains
   {"PmdUtil"     ,""  ,"","","",                                       "StPmdUtil","Load StPmdUtil",kFALSE},
 
   {"QUtils"      ,""  ,"","PmdUtil,EmcUtil","",                      "","Load QA Libs dependencies",kFALSE},
-  {"Stu"         ,""  ,"","","",                         "StEventUtilities","Load StEventUtilities",kFALSE},
+  {"Stu"         ,""  ,"","","",                 "StTpcDb,StEventUtilities","Load StEventUtilities",kFALSE},
```
genevb, line 1355: Why is this change neededr? I found this dependence unnecessary a year ago:
https://www.star.bnl.gov/cgi-bin/protected/viewvc.cgi/cvsroot/StRoot/StBFChain/BigFullChain.h?r1=1.309&r2=1.310

plexoos: Yes, I saw your commits claiming the removal of the dependency, but here is the run-time error that prompted this fix:

```
...
QA :INFO  - Library libStEEmcUtil          [       EEmcUtil] (/star-sw/.sl79_gcc11/LIB/libStEEmcUtil.so)	is loaded
QA :INFO  - Library libStPmdUtil           [        PmdUtil] (/star-sw/.sl79_gcc11/LIB/libStPmdUtil.so)	is loaded
dlopen error: /star-sw/.sl79_gcc11/LIB/libStEventUtilities.so: undefined symbol: gStTpcDb
Load Error: Failed to load Dynamic link library /star-sw/.sl79_gcc11/LIB/libStEventUtilities.so
BFC:FATAL - problem with loading of libStEventUtilities
BFC:FATAL - Stu is switched off 	!!!!
root4star: .sl79_gcc11/OBJ/StRoot/StBFChain/StBFChain.cxx:177: virtual Int_t StBFChain::Load(): Assertion `libraryload!=kStErr' failed.
```

I checked that the `libStEventUtilities.so` library did contain an unresolved symbol for `gStTpcDb` which is defined in `libStTpcDb.so`. It could be that the dependency is established via the following inclusion of `StTpcDb/StTpcDb.h` in `StRoot/StEventUtilities/StRedoTracks.cxx` or something related...

https://github.com/star-bnl/star-sw/blob/330b97707cc661275abdac94d90fb43f3e5b420e/StRoot/StEventUtilities/StRedoTracks.cxx#L11

If there is another fix that would break the unnecessary dependency between these two libraries I'd definitely prefer it. I haven't tried anything other than this fix.

asps/Simulation/starsim/Conscript
plexoos: See https://github.com/star-bnl/star-sw/pull/370/commits/7d5fd7f3ed2a7e6b3d48d140334e6d4af5a0f2a0

klendathu2k (approved):
Looks good...

StarVMC/StarAgmlLib/AgMath.h
plexoos: In general, I would oppose such change in a header file... I don't understand why can't you call `std::abs` directly where it is supposed to be called. Perhaps I am missing the real context.. but here you go https://github.com/star-bnl/star-sw/pull/370/commits/4eeac85f0fed17deca83434200159cd97e6d330c

StRoot/StBFChain/BigFullChain.h
plexoos: So, it looks to me that StEventUtilities does depend on StTpcDb via this call in StRedoTracks.cxx:
https://github.com/star-bnl/star-sw/blob/330b97707cc661275abdac94d90fb43f3e5b420e/StRoot/StEventUtilities/StRedoTracks.cxx#L45-L47

Here are the relevant symbols in the library built with Gcc 11 ⬇️ 
```
.sl79_gcc11/OBJ/StRoot/StEventUtilities/StRedoTracks.o:00000000000007b0 T StRedoTracks::StRedoTracks(char const*, StTpcDbMaker*)
.sl79_gcc11/OBJ/StRoot/StEventUtilities/StRedoTracks.o:00000000000007b0 T StRedoTracks::StRedoTracks(char const*, StTpcDbMaker*)
.sl79_gcc11/OBJ/StRoot/StEventUtilities/StRedoTracks.o:                 U StTpcDbMaker::Class()
.sl79_gcc11/OBJ/StRoot/StEventUtilities/StRedoTracks.o:                 U StMagUtilities::StMagUtilities(StTpcDb*, int)
.sl79_gcc11/OBJ/StRoot/StEventUtilities/StRedoTracks.o:                 U StTpcDb::StTpcDb()
.sl79_gcc11/OBJ/StRoot/StEventUtilities/StRedoTracks.o:0000000000000000 W StTpcDbMaker::tpcDbInterface() const
.sl79_gcc11/OBJ/StRoot/StEventUtilities/StRedoTracks.o:                 U gStTpcDb
```

and with Gcc 4.8.5 ⬇️ 

```
.sl79_gcc485/OBJ/StRoot/StEventUtilities/StRedoTracks.o:0000000000000830 T StRedoTracks::StRedoTracks(char const*, StTpcDbMaker*)
.sl79_gcc485/OBJ/StRoot/StEventUtilities/StRedoTracks.o:0000000000000830 T StRedoTracks::StRedoTracks(char const*, StTpcDbMaker*)
.sl79_gcc485/OBJ/StRoot/StEventUtilities/StRedoTracks.o:                 U StTpcDbMaker::Class()
.sl79_gcc485/OBJ/StRoot/StEventUtilities/StRedoTracks.o:                 U StMagUtilities::StMagUtilities(StTpcDb*, int)
```

Not sure why Gcc 4.8.5 throws away `gStTpcDb`

Do we expect any side effects from loading libStTpcDb.so in BFC as proposed?

veprbl: The output for gcc11 makes sense. The `StTpcDbMaker::tpcDbInterface() const` gets copied over as a weak symbol to `StRedoTracks.o`, so an undefined reference to `gStTpcDb` appears.

plexoos: Unless I see more comments in this conversation, I'll assume that the proposed change in BigFullChain.h is acceptable and can be committed. Thanks

genevb: Have you considered the macros that don't use BFC?

genevb: (I don't mind the BFC change; it appears to be appropriate.)

plexoos:
> Have you considered the macros that don't use BFC?

Sure, let's consider other macros. Do you have anything specific in mind?

genevb:
On Jun 29, 2022, at 5:08 PM, Dmitri Smirnov ***@***.******@***.***>> wrote:

Sure, let's consider other macros. Do you have anything specific in mind?

> grep StEventUtilities $STAR/StRoot/macros/*.C $STAR/StRoot/macros/*/*.C
/afs/rhic.bnl.gov/star/packages/SL21a/StRoot/macros/Load.C<http://rhic.bnl.gov/star/packages/SL21a/StRoot/macros/Load.C>:void Load(Char_t *loadList="St_base,St_Tables,StChain,StDetectorDbMaker,StBichsel,StEvent,StTpcDb,StUtilities,StDbLib,StDbBroker,St_db_Maker,StTriggerDataMaker,StEventUtilities,StBFChain"){
/afs/rhic.bnl.gov/star/packages/SL21a/StRoot/macros/loadMuDst.C<http://rhic.bnl.gov/star/packages/SL21a/StRoot/macros/loadMuDst.C>:  gSystem->Load("StEventUtilities");
/afs/rhic.bnl.gov/star/packages/SL21a/StRoot/macros/makePicoDst.C<http://rhic.bnl.gov/star/packages/SL21a/StRoot/macros/makePicoDst.C>:  gSystem->Load("StEventUtilities");
/afs/rhic.bnl.gov/star/packages/SL21a/StRoot/macros/RedoSpaceCharge.C<http://rhic.bnl.gov/star/packages/SL21a/StRoot/macros/RedoSpaceCharge.C>:  gSystem->Load("StEventUtilities");
/afs/rhic.bnl.gov/star/packages/SL21a/StRoot/macros/StMuDstMakerYear1.C<http://rhic.bnl.gov/star/packages/SL21a/StRoot/macros/StMuDstMakerYear1.C>:  gSystem->Load("StEventUtilities");
/afs/rhic.bnl.gov/star/packages/SL21a/StRoot/macros/analysis/doEvents.C<http://rhic.bnl.gov/star/packages/SL21a/StRoot/macros/analysis/doEvents.C>:    gSystem->Load("StEventUtilities");
/afs/rhic.bnl.gov/star/packages/SL21a/StRoot/macros/analysis/doEvents.C<http://rhic.bnl.gov/star/packages/SL21a/StRoot/macros/analysis/doEvents.C>:// Load StEventUtilities
/afs/rhic.bnl.gov/star/packages/SL21a/StRoot/macros/analysis/find_vertex.C<http://rhic.bnl.gov/star/packages/SL21a/StRoot/macros/analysis/find_vertex.C>:    gSystem->Load("StEventUtilities");
/afs/rhic.bnl.gov/star/packages/SL21a/StRoot/macros/graphics/Draw3DDoc.C<http://rhic.bnl.gov/star/packages/SL21a/StRoot/macros/graphics/Draw3DDoc.C>:   doc.SetSourceDir(".:src:inc:StRoot:StRoot/StarRoot:StRoot/StEventUtilitie");
/afs/rhic.bnl.gov/star/packages/SL21a/StRoot/macros/graphics/GeomBrowse.C<http://rhic.bnl.gov/star/packages/SL21a/StRoot/macros/graphics/GeomBrowse.C>:   gSystem->Load("StEventUtilities");
>

genevb:
Gotta love github's automatic formatting. Trying again from the gui:

> grep StEventUtilities $STAR/StRoot/macros/*.C $STAR/StRoot/macros/*/*.C 
/afs/rhic.bnl.gov/star/packages/SL21a/StRoot/macros/Load.C:void Load(Char_t *loadList="St_base,St_Tables,StChain,StDetectorDbMaker,StBichsel,StEvent,StTpcDb,StUtilities,StDbLib,StDbBroker,St_db_Maker,StTriggerDataMaker,StEventUtilities,StBFChain"){
/afs/rhic.bnl.gov/star/packages/SL21a/StRoot/macros/loadMuDst.C:  gSystem->Load("StEventUtilities");
/afs/rhic.bnl.gov/star/packages/SL21a/StRoot/macros/makePicoDst.C:  gSystem->Load("StEventUtilities");
/afs/rhic.bnl.gov/star/packages/SL21a/StRoot/macros/RedoSpaceCharge.C:  gSystem->Load("StEventUtilities");
/afs/rhic.bnl.gov/star/packages/SL21a/StRoot/macros/StMuDstMakerYear1.C:  gSystem->Load("StEventUtilities");
/afs/rhic.bnl.gov/star/packages/SL21a/StRoot/macros/analysis/doEvents.C:    gSystem->Load("StEventUtilities");
/afs/rhic.bnl.gov/star/packages/SL21a/StRoot/macros/analysis/doEvents.C:// Load StEventUtilities
/afs/rhic.bnl.gov/star/packages/SL21a/StRoot/macros/analysis/find_vertex.C:    gSystem->Load("StEventUtilities");
/afs/rhic.bnl.gov/star/packages/SL21a/StRoot/macros/graphics/Draw3DDoc.C:   doc.SetSourceDir(".:src:inc:StRoot:StRoot/StarRoot:StRoot/StEventUtilitie");
/afs/rhic.bnl.gov/star/packages/SL21a/StRoot/macros/graphics/GeomBrowse.C:   gSystem->Load("StEventUtilities");

plexoos:
Have you considered all of them when you removed the dependency on Stu?

genevb:
I considered them when I added the dependency in BFC, and that drove me to avoid adding the dependency if it wasn't necessary. Thus I removed it shortly after adding it.

-Gene

On Jun 29, 2022, at 5:37 PM, Dmitri Smirnov ***@***.******@***.***>> wrote:

Have you considered all of them when you removed the dependency on Stu?

plexoos:
Did you run all of these macros with default parameters?
And how did you define the criteria to pass/fail the tests?

genevb:
Nope.

plexoos:
Would you then care to share how you run the tests?

genevb:
> On Jun 29, 2022, at 10:48 PM, Dmitri Smirnov ***@***.***> wrote:
> Would you then care to share how you run the tests?

I did not say I ran tests. But I think you've determined that StTpcDb.so needs to be loaded before StEventUtilities.so. I expect that a consequence of this is that all macros which load StEventUtilities.so must load StTpcDb.so beforehand. If those macros are not relevant for, not impacted by this PR, then you should just ignore my question about considering these macros.

-Gene

plexoos:
I see the following files modified in 79e3801453d69237491e49183f62db22aa024b69 "Remove StTpcDb dependence for StEventUtilities"

```
 StRoot/StBFChain/BigFullChain.h                         |  2 +-
 StRoot/StEventUtilities/StEbyET0.cxx                    | 25 +++++++++++++++----------
 StRoot/StMuDSTMaker/COMMON/macros/loadSharedLibraries.C |  1 -
 StRoot/macros/analysis/doEvents.C                       |  7 +++++--
 StRoot/macros/analysis/find_vertex.C                    |  6 ++++--
 StRoot/macros/graphics/GeomBrowse.C                     |  3 +--
 StRoot/macros/loadMuDst.C                               |  1 -
 7 files changed, 26 insertions(+), 19 deletions(-)
```

- StRoot/StBFChain/BigFullChain.h  is what I propose to revert
- Not sure how  StRoot/StEventUtilities/StEbyET0.cxx  is relevant, it is not a macro
- Interpreting StRoot/StMuDSTMaker/COMMON/macros/loadSharedLibraries.C and  StRoot/macros/loadMuDst.C in root won't reveal any problems with undefined symbols because the loader (unlike the linker) would not try to resolve them. If there is a problem it needs to be seen in runtime.
- Running StRoot/macros/analysis/doEvents.C on a .event.root file works fine in ROOT5 gcc4.8.5 environment and with gcc11 it does require a revert due to the missing symbols from libStTpcDb
- When I run StRoot/macros/analysis/find_vertex.C I see errors not related to StTpcDb even in ROOT5/gcc4.8.5 environment

UPDATE. The error from `find_vertex.C` (see test job log file https://github.com/star-bnl/star-sw/runs/7171131802)
```
  env:
    STARENV: root5-gcc485
 *** Start at Date : Sun Jul  3 20:13:49 2022
Processing StRoot/macros/analysis/find_vertex.C("/star/rcf/test/daq/2020/351/st_physics_20351078_raw_5000001.event.root")...
Error: Syntax error StRoot/macros/analysis/find_vertex.C:31:
Error: Missing one of ',;' expected at or after line 31.
Error: Unexpected end of file (G__fignorestream():3) StRoot/macros/analysis/find_vertex.C:261:
This is the end of STAR ROOT -- Goodbye
```

plexoos:
In my previous comment I forgot to mention that running `root4star -l -q 'StRoot/macros/graphics/GeomBrowse.C("y2022a")` also gives me an error:

```
Error in <TUnixSystem::DynamicPathName>: St_geom_Maker[.so | .dll | .dylib | .sl | .dl | .a] does not exist in ...
```
St_geom_Maker does not exist since 2019: 173d6b42a5 - (3 years, 3 months ago) Moved to the attic - Jerome Lauret

`doEvents.C` can be used as a test perhaps. `find_vertex.C` and `GeomBrowse.C` do not seem to be maintained. What's the point of changing them?

So, Gene, I am not sure where we want to go from here. Revert the entire https://github.com/star-bnl/star-sw/commit/79e3801453d69237491e49183f62db22aa024b69 "Remove StTpcDb dependence for StEventUtilities" ?

genevb:
If a macro requires retired code, I would consider the macro retired and remove it.
-Gene

klendathu2k:
Thanks!

The context is that AgML produces both C++ and FORtran sources from the 
xml inputs.
The token "abs" is a FORtran keyword, and a C/C++ function.  So both the 
FORtran
and C++ codes need to compile.

In principle I could have added these definitions in the C++ geometry
implementation files... the AbcdGeo.cxx files produced by the agml 
parser.
I decided to collect these "hacks" inside of a single header instead, so 
that
they wouldn't be forgotten.  AgMath should not appear in any codes 
outside of
the geometry area.

Cheers,
Jason

PR #373, opened 2022-06-28
A minor bug fix in StFcsDb::getDetFromName

akioogawa:
A small bug fix in getDetFromName by David

### DIFF
--- a/StRoot/StFcsDbMaker/StFcsDb.cxx
+++ b/StRoot/StFcsDbMaker/StFcsDb.cxx
@@ -515,9 +515,8 @@ int StFcsDb::getDetFromName(const std::string& detname){
     else if( detname=="ES") {return 1;}
     else if( detname=="HN") {return 2;}
     else if( detname=="HS") {return 3;}
-    else if( detname=="HN") {return 4;}
-    else if( detname=="PN") {return 5;}
-    else if( detname=="PS") {return 6;}
+    else if( detname=="PN") {return 4;}
+    else if( detname=="PS") {return 5;}
     else {
 	LOG_ERROR << "ERROR:Invalid name for detector;Input:"<<detname<<endm;
 	return -1;


jdbrice (approved):

StRoot/StFcsDbMaker/StFcsDb.cxx
```diff
@@ -515,9 +515,8 @@ int StFcsDb::getDetFromName(const std::string& detname){
     else if( detname=="ES") {return 1;}
     else if( detname=="HN") {return 2;}
     else if( detname=="HS") {return 3;}
-    else if( detname=="HN") {return 4;}
-    else if( detname=="PN") {return 5;}
-    else if( detname=="PS") {return 6;}
+    else if( detname=="PN") {return 4;}
+    else if( detname=="PS") {return 5;}
```
plexoos, line 519: You are breaking the public API with this change. Just removing the duplicated line `else if( detname=="HN") {return 4;}` seems to be safer

akioogawa: Sorry but I don't get your comment. Elaborate please? What public API? All lines added & removed were needed change.

plexoos: If you call `getDetFromName("PN")` it would return 5, now it returns 4 for the same argument. Similarly, for "PS". Since the function is public, some code may rely on the specific returned values and can break after the change.

akioogawa: "PN" should return 4, not 5. That was a part of the cut & paste & mis-edited bug...

plexoos: That's the problem with public interfaces when bugs cannot be fixed easily and sometimes have to be carried around... This is of course nowhere close to where it can be relevant.

plexoos (approved):

StRoot/StFcsDbMaker/StFcsDb.cxx
akioogawa: Sooo.... PN should keep returning 5 even though it is wrong? I guess David should have used local (unnamed namespace) const char* DET[6] for this.... since that is used original getName() function, and this is the partial reverse.

Navagyan:
Hi Akio,
Just looking at the part of the code posted above, should not “EN” return 2 instead of “HN”, is this a typo??

Best,
Navagyan

From: akioogawa ***@***.***>
Date: Tuesday, June 28, 2022 at 5:37 PM
Cc: Navagyan Ghimire ***@***.***>, Review requested ***@***.***>
Subject: [External] Re: [star-bnl/star-sw] A minor bug fix in StFcsDb::getDetFromName (PR #373)

@akioogawa commented on this pull request.

In StRoot/StFcsDbMaker/StFcsDb.cxx<https://nam10.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fstar-bnl%2Fstar-sw%2Fpull%2F373%23discussion_r908991947&data=05%7C01%7Cnavagyan.ghimire%40temple.edu%7C3f9f140f44aa45042f0208da594e596c%7C716e81efb52244738e3110bd02ccf6e5%7C0%7C0%7C637920490285155804%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000%7C%7C%7C&sdata=WQsWtZKckB4AjaqYSrn48i5OdlRzpRMBRvSoRGLSfAM%3D&reserved=0>:

> @@ -515,9 +515,8 @@ int StFcsDb::getDetFromName(const std::string& detname){

     else if( detname=="ES") {return 1;}

     else if( detname=="HN") {return 2;}

     else if( detname=="HS") {return 3;}

-    else if( detname=="HN") {return 4;}

-    else if( detname=="PN") {return 5;}

-    else if( detname=="PS") {return 6;}

+    else if( detname=="PN") {return 4;}

+    else if( detname=="PS") {return 5;}

Sooo.... PN should keep returning 5 even though it is wrong? I guess David should have used local (unnamed namespace) const char* DET[6] for this.... since that is used original getName() function, and this is the partial reverse.

akioogawa:
See 
  const char* DET[6]={"EN","ES","HN","HS","PN","PS"};
EN=0 ES=1 HN=2 HS=3 PN=4 PS=5

genevb:
> See
> const char* DET[6]={"EN","ES","HN","HS","PN","PS"};
> EN=0 ES=1 HN=2 HS=3 PN=4 PS=5

Looks like an appropriate situation for an enumeration.

-Gene

akioogawa:
Hummm.... enum kFcsDet {EN,ES...} will not work. In fact we already have

enum StFcsDetectorId{
    kFcsEcalNorthDetId=0,
    kFcsEcalSouthDetId=1,
    ...
};

in StEnunerations.h, which won't work neither.

Original char* getName(int det, int id) was to give standardized channel name (including det&Id, geometry=row/column, readout=DEP/channel and slowControl=SC i2c addr/etc) like "EN023_r02c02_Dep00Ch00_F20/0/00/3", to be used for histogram names and text outputs, so human can easily understand which channel it is. David wants to get back to det(int) from char name created by getName(det,id). I guess I can use enum.name if we are writing Java, or something similar in other languages. I don't know C++ had something like this... Maybe added with recent c++ standards?

genevb:
As long as there's no confusion, I'm not worried about what you've done. With Navagyan 's comment, I thought there seemed to be some confusion. Less compact, but less confusion might be:

TString DET[6];
DET[kFcsEcalNorthDetId] = "EN";
DET[kFcsEcalSouthDetId] = "ES";
DET[kFcsHcalNorthDetId] = "HN";
DET[kFcsHcalSouthDetId] = "HS";
DET[kFcsPresNorthDetId] = "PN";
DET[kFcsPresSouthDetId] = "PS";

...and then access to the DET elements would be via the StFcsDetectorId enum.

No big deal though.
-Gene

akioogawa:
Navagyan >  should not “EN” return 2 instead of “HN”, is this a typo??

I think his confusion is coming from GitHub webpage did not show the first line 
    if( detname=="EN") {return 0;}
and otherwise there is no confusion.

I proposed following to David:

int StFcsDb::getDetFromName(const std::string& detname){
  for(int i=0; i<kFcsNDet; i++) if(detname==DET[i]) return i;
  LOG_ERROR << "ERROR:Invalid name for detector;Input:"<<detname<<endm;
  return -1;
}

I'll see what he says.

akioogawa:
David says getDetFromName() need to be static and cannot use DET which is in anonymous namespace...

PR #376, opened 2022-07-08
[SL22b] Backport #372: Fix misconfigured alignment correction in #358, ...

plexoos:
- Changing default start time modus in StEtofMatchMaker to use bTOF only start time instead of eTOF hybrid start time. 
- Added error hand-down from eTOF geometry to matchmaker if no alignment database table is found.

Additional notes:

This patch (PR #372) fixes a bug missed while introducing misalignment parameters for ETOF counters in PR #156
The code was reading a wrong number of entries from the database thus disabling the correction effect for all channels


plexoos:
If you can think of any other reviewed changes we should patch SL22b with, please let me know.

PhilippWeidenkaff (approved):

PR #378, opened 2022-07-18
Add FST raw hits in StEvent and StMuDst

techuan-huang:
FST group would like to have FST raw hits information for the detector QA in future calibration productions. So we would like to add this information into StEvent and StMuDst.
In StEvent part, StFstRawHits.{h,cxx} are moved from StFstUtil to StEvent. A new collection (StFstEvtCollection) is created to keep the backward compatibility.
In StMuDst part, 8 files were modified (StMuArrays.cxx/h, StMuDstMaker.cxx, StMuFstCollection.cxx/h, StMuFstUtil.cxx/h, and StMuTypes.hh) and 2 files were added (StMuFstRawHit.cxx/h).
Two chain options (fstEvtRawHit and fstMuRawHit) are added to turn this feature on and off.


starsdong:
Hi Techuan, Thank you. I wonder whether you have a comparison on the MuDst file size difference between before and after this change? Thanks

techuan-huang:
> Hi Techuan, Thank you. I wonder whether you have a comparison on the MuDst file size difference between before and after this change? Thanks

Hi Xin, I think you pointed out the key.
I tried with a DAQ file with 7907 events and with only FST and sTGC information.
The size before and after were ~28M and ~41M. Its ~48% larger.

plexoos:
> FST group would like to have FST raw hits information for the detector QA in future calibration productions.

Sounds like the inclusion of raw hits is required only for special processing, if so this PR should include a new BFC option to turn this feature on when needed but it should be off by default.

starsdong:
Hi Te-chuan, it is possible that you can incorporate the suggestion from Dmitri to include a chain option for turning on the FstRawHits in the production?

techuan-huang:
Hi Xin and Dmitri,
Thanks a lot for your suggestions.
I have added a new chain option "fstMuRawHit", and now FST raw hits are NOT stored in MuDst by default.
They will be stored only when the new option is called during the production.

starsdong:
Thank you, Techuan. I would like to hear the comments/approval from Daniel B., particularly for the MuDst related changes. Other owners, please feel free to comment as well.

StRoot/StMuDSTMaker/COMMON/StMuFstRawHit.h
```diff
@@ -0,0 +1,77 @@
...
+ * Author: tchuang 2022
+ **************************************************************************
+ *
+ * Description: Data class for FST hit in StMuDst
+ *
+ **************************************************************************/
+#ifndef StMuFstRawHit_h
+#define StMuFstRawHit_h
+
+#include <TVector3.h>
```
klendathu2k: TVector3 is not used in the header file (that I can see...) so this include should be removed.

StRoot/StMuDSTMaker/COMMON/StMuFstRawHit.h
```diff
@@ -0,0 +1,77 @@
...
+    Int_t       mChannelId;                 ///< channel Id, numbering from 0 to 36863
+    Int_t       mGeoId;                     ///< geometry Id, numbering from 0 to 36863
+    Int_t       mSeedhitflag;               ///< seed hit flag, 0 not a seed hit & 1 mean a seed hit
+    Float_t     mCharge[kFstNumTimeBins];   ///< pedestal non-subtracted ADC value saved in calibration mode;
+    ///< pedestal and CMN subtracted in physics mode
+    Float_t     mChargeErr[kFstNumTimeBins];///< charge error in all time bins
+    UChar_t     mMaxTimeBin;                ///< the max ADC time bin index of the raw hit
+    UShort_t    mIdTruth;                   ///< for embedding, 0 as background
+
+    static UChar_t mDefaultTimeBin;
```
klendathu2k: Why do we need a static data member, which will be common across all hits?

StRoot/StMuDSTMaker/COMMON/StMuFstRawHit.cxx
```diff
@@ -0,0 +1,160 @@
...
+ ***************************************************************************
+ *
+ * Description: Implementation of StMuFstRawHit, the StEvent hit structure
+ *
+ ***************************************************************************/
+#include "StMuFstRawHit.h"
+#include "St_base/StMessMgr.h"
+#include "StEvent/StEnumerations.h"
+#include "StRoot/StFstUtil/StFstRawHit.h"
+#include "StThreeVector.hh"
```
klendathu2k: ... also I don't see a StThreeVector being used... again, we can remove this include...

StRoot/StMuDSTMaker/COMMON/StMuFstUtil.cxx
```diff
@@ -64,10 +68,32 @@ void StMuFstUtil::fillFst(StFstHitCollection* fstcol,StMuFstCollection* muFst)
...
 void StMuFstUtil::fillMuFstHits(StMuFstCollection* muFst,
-        StFstHitCollection* fstcol)
+        StFstHitCollection* fstcol,
+        StFstCollection *fstCollectionPtr)
 {
-    LOG_INFO << "fillMuFstHits" << endm;
-
+    //fill FST Raw Hits
+    if(fstCollectionPtr) {
+        LOG_INFO << "fillMuFstHits (raw hits)" << endm;
```
klendathu2k: No need to print this out on every event.  LOG_DEBUG would be more than sufficient...

StRoot/StMuDSTMaker/COMMON/StMuFstRawHit.h
techuan-huang: Done.

StRoot/StMuDSTMaker/COMMON/StMuFstRawHit.cxx
techuan-huang: Done.

StRoot/StMuDSTMaker/COMMON/StMuFstUtil.cxx
techuan-huang: I changed back to print out only for FST hits to be consistent with other forward detectors and used LOG_DEBUG for FST raw hits.

StRoot/StMuDSTMaker/COMMON/StMuFstRawHit.h
techuan-huang: I just simply copied all the variables we have in StFstRawHit. I think it can be used to change the number of time bins for FST raw hit in some special case in the future.

StRoot/StBFChain/BigFullChain.h
```diff
@@ -1587,6 +1587,7 @@ Bfc_st BFC[] = { // standard chains
                                                                                                     kFALSE},
   {"fstHit",     "", "fstChain", "event,fstCluster", "StFstHitMaker",       "StFstHitMaker","FST Hit Maker",
                                                                                                     kFALSE},
+  {"fstMuRawHit", "", "fstChain", "CMuDST",                     "","","Store FST Raw Hit in MuDst", kFALSE},
```
genevb: Hm....this probably doesn't need to be included in `fstChain` because it implies that the makers invoked by `CMuDST` would be part of the `fstChain`. I haven't tested this, but I'm guessing you can have an empty string for the third element.

StRoot/StMuDSTMaker/COMMON/StMuDstMaker.cxx
```diff
@@ -1208,7 +1209,22 @@ void StMuDstMaker::fillFst(StEvent* ev) {
...
+
+  //raw hit data input
+  StFstCollection* fstCollectionPtr = 0;
+  if (IAttr("fstMuRawHit")){//True for storing FST Raw hits
+    TObjectSet* fstDataSet = (TObjectSet*)GetDataSet("fstRawHitAndCluster");
+    if (!fstDataSet) {
+        LOG_WARN << "StMuDstMaker::Make() - there is no fstDataSet (raw hit) " << endm;
+    }        
+
+    fstCollectionPtr = (StFstCollection*)fstDataSet->GetObject();
```
genevb: Although a warning is printed, a null `fstDataSet` pointer will be used here and cause possible problems. I think a more serious action is needed here to avoid using a null pointer. Perhaps return from the function if the `if (!fstDataSet)` clause, or replace this line with:

    fstCollectionPtr = (fstDataSet ? (StFstCollection*)fstDataSet->GetObject() : 0);
    
And then some similar logic should be done for when `fstCollectionPtr` is null.

StRoot/StBFChain/BigFullChain.h
techuan-huang: Done.

StRoot/StMuDSTMaker/COMMON/StMuDstMaker.cxx
techuan-huang: Thanks for the suggestion and I have done it for `fstDataSet`. For `fstCollectionPtr`, I have:
```
if(fstCollectionPtr) {
  ...
}
```
including all the cases the `fstCollectionPtr` is being used. So, I think it should be fine.

jdbrice:
We discussed this in our FWD software meeting. We do not normally add info directly to StMuDst that does not exists in StEvent. We had some discussion about the problems this *might* cause depending on our future plans with embedding. Because of these I am not ready to approve this, though we may in the  future, especially if we reach parity with StEvent first.

techuan-huang:
I have implemented FST raw hits also in StEvent. In order to keep the backward compatibility, I created StFstEvtCollection in StEvent to store the raw hits instead of moving the original StFstCollection to StEvent. A chain option (fstEvtRawHit) similar to the one for StMuDst is also added.

klendathu2k (commented):
A few comments / questions regarding StFstEvtCollection:

First, you are storing a flat data structure... all hits from all disks will be stored in a single array.   This will be slow to search.  So is there a reason for holding all hits in a single array?

Typically in StEvent we create a hierarchy of collections which reflects the structure of the detector...  e.g. an FST collection holds a collection of disks, which holds a collection of wedges, which holds a vector of hits. So a track which intersects any given wedge, you can quickly limit the possible hits to query.  

I am a little bit concerned that we didn't start with the StEvent integration...  Because this implies that there are production codes which do not follow the event data model.  For example, StFstClusterMaker.  That needs to be refactored to read data from StEvent.  (And probably other makers as well).

plexoos:
One test job failed with the following message:

```
StChain:INFO  - MysqlDb::Connect: Load balancer took 0 ms, will use dbx.star.bnl.gov:3316
StChain:INFO  - Ssl_cipher = AES128-SHA
StChain:ERROR - MysqlDb::ExecQuery() line=662  Query Failed  Lost connection to MySQL server during query
```

Will re-run it and the skipped jobs

techuan-huang:
> Typically in StEvent we create a hierarchy of collections which reflects the structure of the detector... e.g. an FST collection holds a collection of disks, which holds a collection of wedges, which holds a vector of hits. So a track which intersects any given wedge, you can quickly limit the possible hits to query.

 I understand your concern, and the structure you mention is already used for FST hits (the space positions used in tracking). However, the FST RAW hits are not used in tracking and only used for QA. This is why we made chain options to turn it on only when needed. When it is off, everything should be the same as before.

StRoot/StBFChain/BigFullChain.h
```diff
@@ -1587,6 +1587,8 @@ Bfc_st BFC[] = { // standard chains
                                                                                                     kFALSE},
   {"fstHit",     "", "fstChain", "event,fstCluster", "StFstHitMaker",       "StFstHitMaker","FST Hit Maker",
                                                                                                     kFALSE},
+  {"fstEvtRawHit", "", "", "fstRawHit",                       "","","Store FST Raw Hit in StEvent", kFALSE},
+  {"fstMuRawHit", "", "", "CMuDST",                             "","","Store FST Raw Hit in MuDst", kFALSE},
```
klendathu2k: I would assume (hope) that MuDst maker is reading FST raw hits from StEvent (our event data model) rather than a direct read from elsewhere.  So the fstMuRawHit option _implies_ the fstEvtRawHit option.  If so, I suggest...

` {"fstMuRawHit", "", "", "CMuDST,fstEvtRawHit",                             "","","Store FST Raw Hit in MuDst", kFALSE},`

If not... I may have other comments.

StRoot/StMuDSTMaker/COMMON/StMuDstMaker.cxx
```diff
@@ -1208,7 +1209,22 @@ void StMuDstMaker::fillFst(StEvent* ev) {
     connectFstCollection();
     mStMuDst->set(this);
   }
-  mFstUtil->fillMuFst(mFstCollection,fstcol);
+
+  //raw hit data input
+  StFstCollection* fstCollectionPtr = 0;
+  if (IAttr("fstMuRawHit")){//True for storing FST Raw hits
+    TObjectSet* fstDataSet = (TObjectSet*)GetDataSet("fstRawHitAndCluster");
```
klendathu2k: The FST raw hits are taken directly from the raw hit maker, rather than StEvent.  This sets up a situation where the information in the MuDst is no longer a subset of the information in StEvent.

Please obtain the hits from StEvent.

klendathu2k (changes requested):
The MuDst should be a subset of StEvent.  By getting raw hits directly from their producer, the MuDst can contain information not stored in StEvent.   You could not (for instance) create a MuDst containing raw hits from an event.root file which holds them.

Please change the MuDst maker to obtain FST raw hits from the appropriate StEvent collection, and adjust BFC options as noted.

StRoot/StBFChain/BigFullChain.h
techuan-huang: Thanks for the comments. I agree with you that MuDst should be a subset of StEvent. This is the part I forgot to change when I include FST raw hits in StEvent. It is now implemented.

StRoot/StMuDSTMaker/COMMON/StMuDstMaker.cxx
techuan-huang: Done.

klendathu2k (approved):
Looks like all of my concerns have been addressed.

StRoot/StFstRawHitMaker/StFstRawHitMaker.cxx
```diff
@@ -390,6 +391,60 @@ Int_t StFstRawHitMaker::Make()
...
+                        newRawHit->setChannelId( rawHit->getChannelId() );
+                        newRawHit->setGeoId( rawHit->getGeoId() );
+                        newRawHit->setSeedhitflag( rawHit->getSeedhitflag() );
+                        for(int itb=0; itb<kFstNumTimeBins; itb++){
+                            newRawHit->setCharge( rawHit->getCharge(itb), itb );
+                            newRawHit->setChargeErr( rawHit->getChargeErr(itb), itb );
+                        }
+                        newRawHit->setMaxTimeBin( (int)rawHit->getMaxTimeBin() );
+                        newRawHit->setDefaultTimeBin( (int)rawHit->getDefaultTimeBin() );
+                        newRawHit->setIdTruth( rawHit->getIdTruth() );
```
plexoos: Please write a copy constructor for StFstRawHit and use it here in place of these lines

StRoot/StMuDSTMaker/COMMON/StMuFstRawHit.cxx
```diff
@@ -0,0 +1,159 @@
...
+{
+    cout << " elecId=" << getChannelId() << " Charge=(" ;
+    for(int i=0; i<nTimeBins; i++)
+        cout << getCharge(i) << " " ;
+    cout << ") ChargeErr=(" ;
+
+    for(int i=0; i<nTimeBins; i++)
+        cout << getChargeErr(i) << " " ;
+    cout << ") decode0: at disk=" << (short) getDisk() << " wedge=" << (short) getWedge() << " sensor=" << (short) getSensor()
+        << " rstrip=" << (short) getRStrip() << " phistrip=" << (short) getPhiStrip() << endl;
```
plexoos: Why do you convert the types explicitly?

StRoot/StMuDSTMaker/COMMON/StMuFstRawHit.cxx
```diff
@@ -0,0 +1,159 @@
...
+
+void StMuFstRawHit::setMaxTimeBin(int tb)
+{
+    mMaxTimeBin = ((tb < 0 || tb >= kFstNumTimeBins) ? mDefaultTimeBin : tb);
+};
+
+
+void StMuFstRawHit::print(int nTimeBins)
+{
+    cout << " elecId=" << getChannelId() << " Charge=(" ;
```
plexoos: The STAR coding style suggests to use `LOG_...` streamers instead of std::cout

StRoot/StMuDSTMaker/COMMON/StMuFstRawHit.h
plexoos: While static variables should not cause a significant overhead with ROOT files, it is still better to avoid them in persistent classes. StRoot/StEvent/StFstConsts.h is probably a better place for such variables

techuan-huang: I add a new constant in StRoot/StEvent/StFstConsts.h for this and initialize the variable using this constant in the constructor.

StRoot/StFstRawHitMaker/StFstRawHitMaker.cxx
techuan-huang: Done.

StRoot/StMuDSTMaker/COMMON/StMuFstRawHit.cxx
techuan-huang: These functions return "char"s, so need to convert them into integers in order to display correct values.

StRoot/StMuDSTMaker/COMMON/StMuFstRawHit.cxx
techuan-huang: Changed to use LOG_DEBUG instead.

StRoot/StEvent/StFstRawHit.h
```diff
@@ -22,6 +22,8 @@ class StFstRawHit : public StObject
       const Container &charges, const Container &chargeErrs = Container{},
       UChar_t maxTimeBin = 1, UShort_t idTruth = 0);
 
+   StFstRawHit(const StFstRawHit *rawHit);
```
plexoos: This is not a copy constructor. The signature for the copy constructor is: `StFstRawHit(const StFstRawHit& rawHit);`

StRoot/StFstRawHitMaker/StFstRawHitMaker.cxx
```diff
@@ -425,19 +425,7 @@ Int_t StFstRawHitMaker::Make()
...
-                        newRawHit->setSeedhitflag( rawHit->getSeedhitflag() );
-                        for(int itb=0; itb<kFstNumTimeBins; itb++){
-                            newRawHit->setCharge( rawHit->getCharge(itb), itb );
-                            newRawHit->setChargeErr( rawHit->getChargeErr(itb), itb );
-                        }
-                        newRawHit->setMaxTimeBin( (int)rawHit->getMaxTimeBin() );
-                        newRawHit->setDefaultTimeBin( (int)rawHit->getDefaultTimeBin() );
-                        newRawHit->setIdTruth( rawHit->getIdTruth() );
-
+                        StFstRawHit* newRawHit = new StFstRawHit( rawHit );
```
plexoos: Are you actually sure you need to create copies of StFstRawHit objects with the `new` operator? Can't you just copy pointers from `rawHitCollectionPtr->getRawHitVec()` to `mFstEvtCollection->rawHits()`?

StRoot/StMuDSTMaker/COMMON/StMuFstRawHit.h
```diff
@@ -67,7 +67,7 @@ class StMuFstRawHit : public TObject {
     UChar_t     mMaxTimeBin;                ///< the max ADC time bin index of the raw hit
     UShort_t    mIdTruth;                   ///< for embedding, 0 as background
 
-    static UChar_t mDefaultTimeBin;
+    UChar_t     mDefaultTimeBin;
```
plexoos: When I commented about the static member in a persistent class I actually meant to remove it completely or at least remove it from the streamer. Here is how you can tell ROOT not to stream a data member:

```diff
-    static UChar_t mDefaultTimeBin;
+    static UChar_t mDefaultTimeBin;   //!
```

StRoot/StEvent/StFstRawHit.h
techuan-huang: Thanks for correction. Changed it.

StRoot/StFstRawHitMaker/StFstRawHitMaker.cxx
techuan-huang: I'm not sure if I got your point, but I had tried to directly put `rawHit`, the pointer from the iterator of `rawHitCollectionPtr->getRawHitVec()` before as following:
```
mFstEvtCollection->addRawHit(rawHit);
```
The compiler won't complain but I got some error like:
```
StEventMaker:FATAL - TUnixSystem::DispatchSignals : segmentation violation
```
when I ran the codes.

StRoot/StMuDSTMaker/COMMON/StMuFstRawHit.h
techuan-huang: Thanks for clarification. I have changed back the codes and add the `//!` as you suggested.

starsdong:
Daniel, could you please review this PR, comment or approve it so we can proceed soon? Thanks

StRoot/StEvent/StFstRawHit.cxx
```diff
@@ -33,6 +33,19 @@ StFstRawHit::StFstRawHit(int channelId, int geoId,
...
+    mChannelId = rawHit.getChannelId();
+    mGeoId = rawHit.getGeoId();
+    mSeedhitflag = rawHit.getSeedhitflag();
+    for(int itb=0; itb<kFstNumTimeBins; itb++){
+        mCharge[itb] = rawHit.getCharge(itb);
+        mChargeErr[itb] = rawHit.getChargeErr(itb);
+    }
+    mMaxTimeBin = (int)rawHit.getMaxTimeBin();
+    mIdTruth = (int)rawHit.getDefaultTimeBin();
+    mDefaultTimeBin = rawHit.getIdTruth();
```
plexoos: This does not look right

StRoot/StEvent/StFstRawHit.cxx
```diff
@@ -33,6 +33,19 @@ StFstRawHit::StFstRawHit(int channelId, int geoId,
    std::copy(std::begin(chargeErrs), std::end(chargeErrs), mChargeErr);
 }
 
+StFstRawHit::StFstRawHit(const StFstRawHit &rawHit)
+{
+    mChannelId = rawHit.getChannelId();
+    mGeoId = rawHit.getGeoId();
+    mSeedhitflag = rawHit.getSeedhitflag();
```
plexoos: Even though the data members are private, you can still access them directly without calling the getters in a copy constructor. I think the following is less error-prone:

```suggestion
    mChannelId = rawHit.mChannelId;
    mGeoId = rawHit.mGeoId;
    mSeedhitflag = rawHit.mSeedhitflag;
```

StRoot/StEvent/StFstRawHit.cxx
techuan-huang: Thanks for the carefully checking. It is implemented with your next comment.

StRoot/StEvent/StFstRawHit.cxx
techuan-huang: I changed all the member variables accordingly.

StRoot/StEvent/StFstEvtCollection.cxx
```diff
@@ -0,0 +1,52 @@
...
+
+StFstEvtCollection::StFstEvtCollection() {/* no operation*/}
+StFstEvtCollection::~StFstEvtCollection()
+{
+    // Usually this wouldn't be necessary but mRawHits is a polymorphic container and StFstRawHit provides
+    // its own new/delete operator
+    for (unsigned int i=0; i<mRawHits.size(); i++) {
+        delete mRawHits[i];
+        mRawHits[i] = 0;
+    }
```
plexoos: The statement in the comment is incorrect

> StFstRawHit provides its own new/delete operator

As far as I can tell, StFstRawHit does not implement custom new/delete operators and even if it did the statement is still incorrect. The comment in the header file is actually valid:

> ```
> StSPtrVecFstRawHit mRawHits; ///< Inherits from StStrArray which takes care of deleting the objects
>                              ///< pointed by the pointers in this array.
>```

This class does not need a destructor. See other similar classes in StEvent/St*RawHit.h

StRoot/StMuDSTMaker/COMMON/StMuFstRawHit.cxx
plexoos: Actually, it is still `cout` here and in other places. Please change to LOG_...

StRoot/StFstRawHitMaker/StFstRawHitMaker.cxx
```diff
@@ -390,6 +391,48 @@ Int_t StFstRawHitMaker::Make()
...
+                StFstRawHitCollection *rawHitCollectionPtr = mFstCollectionPtr->getRawHitCollection( wedgeIdx );
+                if( rawHitCollectionPtr ){
+                    std::vector<StFstRawHit*>& rawHitVec = rawHitCollectionPtr->getRawHitVec();
+                    std::vector< StFstRawHit* >::iterator rawHitIter;
+
+                    for( rawHitIter = rawHitVec.begin(); rawHitIter != rawHitVec.end(); ++rawHitIter ){
+                        StFstRawHit* rawHit = *rawHitIter;
+                        StFstRawHit* newRawHit = new StFstRawHit( *rawHit );
+                        mFstEvtCollection->addRawHit(newRawHit);
+                    }
```
plexoos, line 430: This copying of allocated StFstRawHit objects from one vector to another is very questionable... but at least do it in one line like so:

```suggestion
                    //#include <algorithm>
                    std::transform(begin(rawHitVec), end(rawHitVec), back_inserter(mFstEvtCollection->rawHits()), [](const StFstRawHit* h) { return new StFstRawHit(*h);} );
```

StRoot/StEvent/StFstEvtCollection.cxx
techuan-huang: I removed the destructor for this class and also implemented your next comment about `cout`.

StRoot/StFstRawHitMaker/StFstRawHitMaker.cxx
techuan-huang: I have tried your suggestion, however, I got errors as follow:
```
/usr/include/c++/4.8.2/bits/stl_iterator.h:436:7: error: no type named 'value_type' in 'class StSPtrVecFstRawHit'
       operator=(const typename _Container::value_type& __value)
       ^
/usr/include/c++/4.8.2/bits/stl_iterator.h:443:7: error: no type named 'value_type' in 'class StSPtrVecFstRawHit'
       operator=(typename _Container::value_type&& __value)
       ^
In file included from /usr/include/c++/4.8.2/algorithm:62:0,
                 from /afs/rhic.bnl.gov/star/ROOT/5.34.38/.sl73_gcc485/rootdeb/include/TMath.h:34,
                 from /afs/rhic.bnl.gov/star/packages/DEV/StRoot/StChain/StMaker.h:16,
                 from /afs/rhic.bnl.gov/star/packages/DEV/StRoot/StChain/StRTSBaseMaker.h:12,
                 from .sl73_gcc485/obj/StRoot/StFstRawHitMaker/StFstRawHitMaker.h:12,
                 from .sl73_gcc485/obj/StRoot/StFstRawHitMaker/StFstRawHitMaker.cxx:1:
/usr/include/c++/4.8.2/bits/stl_algo.h: In instantiation of '_OIter std::transform(_IIter, _IIter, _OIter, _UnaryOperation) [with _IIter = __gnu_cxx::__normal_iterator<StFstRawHit**, std::vector<StFstRawHit*> >; _OIter = std::back_insert_iterator<StSPtrVecFstRawHit>; _UnaryOperation = StFstRawHitMaker::Make()::__lambda0]':
.sl73_gcc485/obj/StRoot/StFstRawHitMaker/StFstRawHitMaker.cxx:432:171:   required from here
/usr/include/c++/4.8.2/bits/stl_algo.h:4926:12: error: no match for 'operator=' (operand types are 'std::back_insert_iterator<StSPtrVecFstRawHit>' and 'StFstRawHit*')
  *__result = __unary_op(*__first);
            ^
/usr/include/c++/4.8.2/bits/stl_algo.h:4926:12: note: candidates are:
In file included from /usr/include/c++/4.8.2/bits/stl_algobase.h:67:0,
                 from /usr/include/c++/4.8.2/bits/char_traits.h:39,
                 from /usr/include/c++/4.8.2/string:40,
                 from /usr/include/c++/4.8.2/stdexcept:39,
                 from /usr/include/c++/4.8.2/array:38,
                 from .sl73_gcc485/obj/StRoot/StFstRawHitMaker/StFstRawHitMaker.h:5,
                 from .sl73_gcc485/obj/StRoot/StFstRawHitMaker/StFstRawHitMaker.cxx:1:
/usr/include/c++/4.8.2/bits/stl_iterator.h:402:11: note: std::back_insert_iterator<StSPtrVecFstRawHit>& std::back_insert_iterator<StSPtrVecFstRawHit>::operator=(const std::back_insert_iterator<StSPtrVecFstRawHit>&)
     class back_insert_iterator
           ^
/usr/include/c++/4.8.2/bits/stl_iterator.h:402:11: note:   no known conversion for argument 1 from 'StFstRawHit*' to 'const std::back_insert_iterator<StSPtrVecFstRawHit>&'
/usr/include/c++/4.8.2/bits/stl_iterator.h:402:11: note: std::back_insert_iterator<StSPtrVecFstRawHit>& std::back_insert_iterator<StSPtrVecFstRawHit>::operator=(std::back_insert_iterator<StSPtrVecFstRawHit>&&)
/usr/include/c++/4.8.2/bits/stl_iterator.h:402:11: note:   no known conversion for argument 1 from 'StFstRawHit*' to 'std::back_insert_iterator<StSPtrVecFstRawHit>&&'
```
Did I miss anything?

plexoos: Right, seems like the `StSPtrVec...` types wrapping `std::vector` are not very suitable for the standard algorithms

plexoos (approved):

jdbrice (approved):

PR #379, opened 2022-07-27
Add Geant track info (g2t) to StFcsHit

akioogawa:
Adding parent g2t track info to StEvent/StFcsHit.
Update StFcsFastSimulatorMaker to push these info from g2t table to StFcsHit.
Adding utility functions to trace back to immediate parent and primary g2t track for a given StFcsHit and StFcsCluster in StFcsDb.


StRoot/StEvent/StFcsHit.cxx
```diff
@@ -139,3 +141,20 @@ void StFcsHit::print(Option_t *option) const {
     }
     cout << endl;
 }
+
+void StFcsHit::addGeantTrack(unsigned int id, float e){    
+    unsigned int n=mGeantTracks.size();
+    for(unsigned int i=0; i<n; i++){
+	if(mGeantTracks[i].first == id) {mGeantTracks[i].second += e; return;}
+    }
```
plexoos: Is there any reason why you don't want to use a function from the standard library? `std::find_if()` seems appropriate for this case. The advantage of using iterator-based functions is that they let you switch to a different container class.

StRoot/StEvent/StFcsHit.h
```diff
@@ -100,9 +104,10 @@ class StFcsHit : public StObject {
     Float_t  mEnergy=0.0;     // corrected energy
     StFcsCluster* mCluster=0; // pointer to cluster this hit belongs
     TArrayS* mData=0;         // 12bit ADC values + flag at highest 4 bits, array of timebin
-    
-    ClassDef(StFcsHit,5)
 
+    vector<pair<unsigned int, float>> mGeantTracks; // parent G2T track id and dE
```
plexoos, line 107: What are the reasons for choosing the `vector` class for this member? It looks as you want to store sorted and unique elements. Why not use `set`?

StRoot/StFcsClusterMaker/StFcsClusterMaker.cxx
```diff
@@ -217,6 +218,48 @@ int StFcsClusterMaker::makeCluster(int det) {
...
+	}
+      }
+    }
+    if(g2ttrk){
+      float frc=0;
+      int nh = hits.size();
+      for(int i=0; i<nh; i++){
+	StFcsHit* hit=hits[i];
+	g2t_track_st* trk = mDb->getParentG2tTrack(hit,g2ttrk,frc);
+	printf("Det=%1d Id=%3d E=%8.3f Parent  Id=%4d Pid=%4d E=%8.3f Frc=%6.3f\n",
```
plexoos: Shouldn't you be using LOG_DEBUG instead of printf for consistency? For string formatting we have `Form()` and `FormString()` functions returning C and C++ srtings respectively

StRoot/StFcsDbMaker/StFcsDb.h
```diff
@@ -256,6 +257,13 @@ class StFcsDb : public TDataSet {
   void setPedestal(int ehp, int ns, int dep, int ch, float ped); //! setting pedestal
   void readPedFromText(const char* file="fcsped.txt"); //! reading pedestal from text
   
+  //g2t track info
+  unsigned int backTraceG2tTrack(unsigned int id, g2t_track_st* g2ttrk);
+  g2t_track_st* getParentG2tTrack(StFcsHit* h,      g2t_track_st* g2ttrk, float& fraction, unsigned int order=0);
+  g2t_track_st* getParentG2tTrack(StFcsCluster* c,  g2t_track_st* g2ttrk, float& fraction, unsigned int order=0);
+  g2t_track_st* getPrimaryG2tTrack(StFcsHit* h,     g2t_track_st* g2ttrk, float& fraction, unsigned int order=0);
+  g2t_track_st* getPrimaryG2tTrack(StFcsCluster* c, g2t_track_st* g2ttrk, float& fraction, unsigned int order=0);
```
plexoos: You should use `std::tuple` to return multiple values. Passing inputs as function arguments is more natural and is inline with standard math notation.
Please use `const` for immutable arguments. These are public methods, so you need to explain the meaning of the arguments with doxygen comments.

StRoot/StEvent/StFcsHit.cxx
akioogawa: auto trk = std::find_if(mGeantTracks.begin(), mGeantTracks.end(),
                            [&id](const std::pair<unsigned int, float>& t) {
                                return t.first == id;});
    if(trk!=mGeantTracks.end()) {trk->second += e; return;}
This seems to work...and while I do use lambda expressions elsewhere... I try to avoid it when possible. Maybe my inexperience, but it makes much harder to understand the code. In this case I don't think it made code any better nor more generic nor easier to maintain. If you insist, I can switch...

StRoot/StEvent/StFcsHit.h
akioogawa: I don't know how to use set<pair<unsigned int, float>> and later modify (add) to float in the pair.... nor keep the set unique just fist unsigned int regardless of float part. Please let me know.

StRoot/StFcsClusterMaker/StFcsClusterMaker.cxx
akioogawa: Yes update coming

StRoot/StEvent/StFcsHit.cxx
plexoos: I mentioned one of the reasons above. 
It might be easier to read if you assign the comparison function to a variable. C++20 provides an even terser syntax

akioogawa: Are we using c++11? 17? 20?

StRoot/StFcsDbMaker/StFcsDb.h
akioogawa: At
  /star/u/akio/fcsmc/star-sw/StRoot/StFcsDbMaker/StFcsDb.h (and .cxx)
I'm having issue with CINT:

.sl73_gcc485/obj/StRoot/StFcsDbMaker/StFcsDbMaker_Cint.cxx: In function 'int G__StFcsDbMaker_Cint_646_0_93(G__value*, const char*, G__param*, int)':
.sl73_gcc485/obj/StRoot/StFcsDbMaker/StFcsDbMaker_Cint.cxx:1338:73: error: invalid use of 'class std::tuple<g2t_track_st*, float, int>'
       G__letint(result7, 105, (long) ((StFcsDb*) G__getstructoffset())->std::tuple<g2t_track_st*,float,int> getParentG2tTrack((StFcsHit*) G__int(libp->para[0]), (g2t_track_st*) G__int(libp->para[1])
                                                                         ^
.sl73_gcc485/obj/StRoot/StFcsDbMaker/StFcsDbMaker_Cint.cxx:1342:73: error: invalid use of 'class std::tuple<g2t_track_st*, float, int>'
       G__letint(result7, 105, (long) ((StFcsDb*) G__getstructoffset())->std::tuple<g2t_track_st*,float,int> getParentG2tTrack((StFcsHit*) G__int(libp->para[0]), (g2t_track_st*) G__int(libp->para[1])));

StRoot/StEvent/StFcsHit.cxx
plexoos: Here is how the body of this function may look like

```c++
  auto cmp = [](decltype(mGeantTracks)::value_type t){ return t.first == id; };
  auto trk = find_if(begin(mGeantTracks), end(mGeantTracks), cmp);

  if (trk == end(mGeantTracks))
    mGeantTracks.insert(trk, make_pair(id, e));
  else
    trk->second += e
```

It does not make much sense to pass by reference, and the empty return statement is not necessary. You don't have to capture `id` if you declare the argument `const` as you should anyway. In C++20 the `cmp` lambda simply becomes

```c++
  auto cmp = [](auto t){ return t.first == id; };
```

I mention C++20 because we could switch to GCC11 supporting C++20 any time if our test suite was better developed and easier to interpret. This is one of the main reasons why we still use gcc 4.8.5 and 32-bit libraries in production...

StRoot/StEvent/StFcsHit.h
plexoos: If you change your `StFcsHit::addGeantTrack(unsigned int id, float e)` to work with iterators as suggested then it will work for the both `vector` and `set` types. Please note that I am not suggesting to switch to `set` of pairs right away. I have no idea how the performance may change because I don't know how many inserts or look-ups you expect. I have no real argument for a `set` other than I see your intention to store sorted and unique elements, hence, a `set` would be my first choice but a `vector` may work just as well.

In case you do decide to investigate the `set`, all you need to do is to redefine the default comparator for `std::pair<unsigned, float>` to something like this:

```c++
auto cmp = [](pair<unsigned, float> a, pair<unsigned, float> b) { return a.first < b.first; };

std::set< pair<unsigned, float>, decltype(cmp)> s(cmp);
```

StRoot/StFcsDbMaker/StFcsDb.h
plexoos: Ah, yes. I forgot about our ROOT5/rootcint not being c++11 compatible. Scratch the comment about the `std::tuple`...

StRoot/StEvent/StFcsHit.h
veprbl: Isn't this calling for `std::map`?

klendathu2k: For the most part, the hits from a single track which initiates an EM or hadronic cascade will be presented to the slow simulator before the next track is processed.  So the vector is the natural choice here.  For any geant hit added to the FCS hit, the pair which needs to be updated is likely to be at the back of the vector.  Once all hits have been processed, we can loop over the FCS hits and ensure that they are sorted descending in energy.

StRoot/StEvent/StFcsHit.cxx
```diff
@@ -139,3 +141,26 @@ void StFcsHit::print(Option_t *option) const {
     }
     cout << endl;
 }
+
+void StFcsHit::addGeantTrack(const unsigned int id, float e){    
+    // auto cmp = [](auto t){ return t.first == id; }; //c++20
+    auto cmp = [id](decltype(mGeantTracks)::value_type t){ return t.first == id; };
+    auto trk = find_if(begin(mGeantTracks), end(mGeantTracks), cmp);
```
klendathu2k: A reverse iteration through the vector should be optimal, because (mostly) the hits from a single track will be presented in sequence before the hits from the next track are processed.

akioogawa: auto trk = find_if(rbegin(mGeantTracks), rend(mGeantTracks), cmp);
    if(trk == rend(mGeantTracks)){...}

Will this make the auto to be reverse iterator automatically? Or do I need do something like 

  std::vector<pair<unsigned int,float>>::reverse_iterator trk= find_if(...)

klendathu2k:
I believe that will work.

StRoot/StEvent/StFcsHit.cxx
akioogawa: Looks like

    auto trk = find_if(mGeantTracks.rbegin(), mGeantTracks.rend(), cmp);
    if(trk == mGeantTracks.rend()){
      mGeantTracks.push_back(make_pair(id, e));
    }else{
      trk->second += e;
    }

works. I don't know why but rbegin(x) did not compile, while x.rbegin() worked??? I changed insert->push_back since I guess rend() is pointing at element -1, not size()?

akioogawa: Google says begin(x) & end(x) were added by c++11, while rbegin(x), rend(x) etc were added at c++14.

plexoos:
@akioogawa can you elaborate on how the new data member `StFcsHit::mGeantTracks` will be used? It looks like you want to make it persistent in `event.root` files but I don't think we normally save them in simulation jobs.

akioogawa:
A student is working on cluster finder for charged hadrons at StEvent file right now. Once this is in the place, Daniel will be adding this to Mudst, and we move on to Mudst.

StRoot/StEvent/StFcsHit.cxx
```diff
@@ -139,3 +141,26 @@ void StFcsHit::print(Option_t *option) const {
     }
     cout << endl;
 }
+
+void StFcsHit::addGeantTrack(const unsigned int id, float e){
```
klendathu2k: I'd mark "e" as const here as well to avoid confusion (looks like you intend to modify "e" but didn't).

StRoot/StFcsDbMaker/StFcsDb.cxx
```diff
@@ -1979,3 +1979,125 @@ void StFcsDb::readGainCorrFromText(){
...
+    i = g2ttrk[i].next_parent_p - 1;
+  }
+  if(mDebug>3) LOG_INFO<<Form("  BackTrace from=%3d id=%3d Epid=%4d Gpid=%3d Vtx=%3d Parent=%3d E=%6.2f Primary!!",
+			      id,g2ttrk[i].id,g2ttrk[i].eg_pid,g2ttrk[i].ge_pid,g2ttrk[i].start_vertex_p,
+			      g2ttrk[i].next_parent_p,g2ttrk[i].e)<<endm;
+  return i + 1;
+}
+    
+const g2t_track_st* StFcsDb::getParentG2tTrack(StFcsHit* h, g2t_track_st* g2ttrk, float& fraction, int& ntrk, unsigned int order){
+  const vector<pair<unsigned int, float>>& gt = h->getGeantTracks();
```
klendathu2k: I don't see where "using namespace std;" is declared?  If I didn't miss it, it means that it is included through a header file... which is not the best practice.

I would add "using namespace std;" near the top of StFcsDb.cxx.

StRoot/StFcsDbMaker/StFcsDb.h
```diff
@@ -256,6 +257,19 @@ class StFcsDb : public TDataSet {
   void setPedestal(int ehp, int ns, int dep, int ch, float ped); //! setting pedestal
   void readPedFromText(const char* file="fcsped.txt"); //! reading pedestal from text
   
+  // Tracing back from a given g2t_track_id to primary track's g2t_track_id
+  unsigned int backTraceG2tTrack(unsigned int id, g2t_track_st* g2ttrk);
+
+  // Getting pointer to parent & primary g2t_track from StFcsHit & StFcsCluster
```
klendathu2k: You should use doxygen-style comments for your methods -- https://doxygen.nl/manual/docblocks.html.

klendathu2k (changes requested):
A couple of small changes requested, but otherwise this looks good to me.

StRoot/StFcsDbMaker/StFcsDb.cxx
```diff
@@ -1979,3 +1979,125 @@ void StFcsDb::readGainCorrFromText(){
...
+  float detot=0;
+  for(int itrk=0; itrk<n; itrk++){
+    float de=gt[itrk].second;
+    unsigned int id=backTraceG2tTrack(gt[itrk].first, g2ttrk);
+    int found=0;
+    unsigned int np=primary.size();
+    for(unsigned int jtrk=0; jtrk<np; jtrk++){
+      if(primary[jtrk].first == id) {primary[jtrk].second += de; found=1; break;}
+    }
+    if(found==0) primary.push_back(make_pair(id,de));
```
plexoos: If I understand correctly what you are trying to accomplish here, the logic can be replaced by a single line of code such as:

```c++
primary[id] += de;
```

if `primary` is defined as `unordered_map<unsigned int, float> primary;`. The access to the elements of an `unordered_map` is very efficient O(1)

StRoot/StFcsDbMaker/StFcsDb.cxx
```diff
@@ -1979,3 +1979,125 @@ void StFcsDb::readGainCorrFromText(){
...
+    detot+=de;
+  }
+  ntrk = primary.size();
+  if(order > ntrk) {fraction=0; return 0;}
+  std::sort(primary.begin(), primary.end(),
+	    [](const pair<unsigned int,float>&a, const pair<unsigned int,float>&b){
+	      return b.second < a.second;
+	    });
```
plexoos: I don't think you need to sort all of the elements to get the value of the Nth highest element. In fact, there is a function in the standard library `std::nth_element` that may do what you want.

StRoot/StFcsFastSimulatorMaker/StFcsFastSimulatorMaker.cxx
```diff
@@ -318,6 +322,7 @@ void StFcsFastSimulatorMaker::fillStEvent(StEvent* event) {
       int ehp = mFcsDb->ecalHcalPres(det);
       hits[i]->setAdc(0,adc);
       hits[i]->setEnergy(digi_energy);
+      hits[i]->sortGeantTracks();
```
plexoos: Do we really need to sort all tracks contributing to every hit before they persist? If you look at only a small fraction of hits, wouldn't it be somehow more efficient to delay this until the analysis stage?

StRoot/StFcsDbMaker/StFcsDb.cxx
veprbl: Yes, I don't understand @klendathu2k's answer above. This does not participate in the simulation data flow. The rest is just a premature optimization.

klendathu2k: @veprbl not sure which answer are you referring to?  (this discussion thread is getting a bit long).

veprbl: It was your original response to suggestion to use a map type from STL:
https://github.com/star-bnl/star-sw/pull/379#discussion_r933201592
> For the most part, the hits from a single track which initiates an EM or hadronic cascade will be presented to the slow simulator before the next track is processed. So the vector is the natural choice here. For any geant hit added to the FCS hit, the pair which needs to be updated is likely to be at the back of the vector. Once all hits have been processed, we can loop over the FCS hits and ensure that they are sorted descending in energy.

After reading this, at least I left with impression that the `vector` type is somehow justified by downstream code or somehow more performant. Hence my suggestion to write code as simple and short as possible so that people can understand it before we optimize anything.

klendathu2k: One of your impressions is correct.  The vector implementation will be more performant than the other options.  (Or at least I expect it to be given the order in which GEANT creates, and starsim stores, the hits).  

Storing the ID truth pair (ID track, dE contribution) in order of the highest contribution to the hit was chosen by Akio, and I agree that it is the most natural ordering.  This is what drives the choice of a vector rather than a set or a map.

StRoot/StEvent/StFcsHit.cxx
veprbl:
```suggestion
void StFcsHit::addGeantTrack(unsigned int id, float e){    
```

StRoot/StEvent/StFcsHit.h
```diff
@@ -87,9 +87,13 @@ class StFcsHit : public StObject {
     void setCluster(StFcsCluster* clu) {mCluster = clu;}
     StFcsCluster *cluster() {return mCluster;}
 
+    const vector<pair<unsigned int, float>>& getGeantTracks() const {return mGeantTracks;}
+    void addGeantTrack(const unsigned int id, float e);
+    void sortGeantTracks();
```
veprbl: Do we need the `sortGeantTracks()` to be a part of the public interface if it's supposedly called unconditionally anyway?

StRoot/StEvent/StFcsHit.h
```diff
@@ -87,9 +87,13 @@ class StFcsHit : public StObject {
     void setCluster(StFcsCluster* clu) {mCluster = clu;}
     StFcsCluster *cluster() {return mCluster;}
 
+    const vector<pair<unsigned int, float>>& getGeantTracks() const {return mGeantTracks;}
```
veprbl, line 90: Do we really need this information? It's not complete enough to learn something about the shower, but too complicated to learn about what caused the hit. I don't see why the analyzers couldn't get by with just `unsigned int idTruth()` and `float qaTruth()` (aka `fraction` in your current PR). This should be enough for two-particle studies (energy fraction that is not associated with the leading particle must be coming from the other one). I suppose, unlike for TPC, you also wanted to store the particle that arrived to the detector (`getParentG2tTrack`) in the current PR, but again, this is not equivalent to storing energy and id of every hit.

I'm mostly asking because I'm pretty sure the interface from StEvent is going to migrate to MuDst without a change.

StRoot/StFcsDbMaker/StFcsDb.cxx
veprbl: I'm not sure if it's worth discussing performance of the code that casually calls `std::sort`.

StRoot/StEvent/StFcsHit.h
klendathu2k: Yes, it is needed.  A single ID truth is insufficient to understand the relative contributions of electromagnetic and hadronic components to the cluster energy response  (You don't know what the "rest" of the energy is...)   

With the list of track IDs hitting each cluster, their contribution to the energy deposition in the cluster, and the MC track's kinematics from the MC event record, you have the complete information needed to understand the clustering performance.

veprbl: > Yes, it is needed. A single ID truth is insufficient to understand the relative contributions of electromagnetic and hadronic components to the cluster energy response (You don't know what the "rest" of the energy is...)

Okay, but if that is the case, I need to also save PDG id. If not, I need to look into geant.root, and at this point I might as well not save anything to event.root. Right?

> With the list of track IDs hitting each cluster, their contribution to the energy deposition in the cluster, and the MC track's kinematics from the MC event record, you have the complete information needed to understand the clustering performance.


Again, for any foreseeable study, there are just two particles that could merge into the same cluster, one fraction per cell describes that case it well enough.

klendathu2k: > Okay, but if that is the case, I need to also save PDG id. If not, I need to look into geant.root, and at this point I might as well not save anything to event.root. Right?

PDG ID is available in the MuDst MuMcTrack branch.

> Again, for any foreseeable study, there are just two particles that could merge into the same cluster, one fraction per cell describes that case it well enough.

Relative contribution of hadronic and electromagnetic particles to the energy of a cluster is one such study which requires this information.

StRoot/StEvent/StFcsHit.cxx
plexoos: Using `const` when passing by value is fine and, actually, should be encouraged. As a bonus one can skip explicit  capturing of such cons parameters in lambda expressions.

veprbl: > Using `const` when passing by value is fine and, actually, should be encouraged. As a bonus one can skip explicit capturing of such cons parameters in lambda expressions.

I've never seen a codebase that does that. I'd keep surprises to a minimum ;)

StRoot/StEvent/StFcsHit.h
veprbl: > PDG ID is available in the MuDst MuMcTrack branch.

I'm not very familiar with that facility. And this answer also implies that this structure will be propagated to MuDst.

> 
> > Again, for any foreseeable study, there are just two particles that could merge into the same cluster, one fraction per cell describes that case it well enough.
> 
> Relative contribution of hadronic and electromagnetic particles to the energy of a cluster is one such study which requires this information.

Okay, so that would be a study that looks at cluster merging and hadronic/electromagnetic fractions at the same time. Thank you for the explanations. I hope other experts/analyzers can chime in.

StRoot/StEvent/StFcsHit.cxx
plexoos: I think const is usually omitted in header files but in implementation files it serves as documentation

StRoot/StFcsDbMaker/StFcsDb.cxx
```diff
@@ -1979,3 +1979,125 @@ void StFcsDb::readGainCorrFromText(){
...
+  fraction = primary[order].second / detot;
+  return &g2ttrk[primary[order].first-1];
+}
+
+const g2t_track_st* StFcsDb::getParentG2tTrack(StFcsCluster* c, g2t_track_st* g2ttrk, float& fraction, int& ntrk, unsigned int order){
```
genevb, line 2011: I think it would be good to document somewhere what `order` means in these functions. It appears to me that invoking these functions with different values of `order` will be mean repeating all the code as nothing about the parents is stored/cached, so I'd really like to better understand the use cases of `order` to understand if this is the best way to encode these functions.

Thanks,
-Gene

StRoot/StFcsDbMaker/StFcsDb.cxx
akioogawa: Please suggest alternative

StRoot/StFcsDbMaker/StFcsDb.cxx
akioogawa: Done

StRoot/StFcsDbMaker/StFcsDb.h
akioogawa: Done

StRoot/StFcsDbMaker/StFcsDb.cxx
akioogawa: Done

StRoot/StEvent/StFcsHit.cxx
akioogawa: const taken out from passing by valie

StRoot/StEvent/StFcsHit.h
akioogawa: I don't know how to make sortGeantTracks() private and StFcsHit knows when to sort from inside. I guess I can move sorting to util (StFcsDb) as suggested above.

StRoot/StEvent/StFcsHit.h
akioogawa: - Yes, idTruth is not enough, especially when many towers contributing to a cluster, and hadron showers are fat and big.
  - Yes, when we get to mudst, since Mudst has StMcTrack, but not g2t_hits.
   - No, when reading event.root file and geant.root file in sync. But this means we essentially re-run StFcsFastSimulationMake and creating new StFcsHits. (for historical reason, we don't have StMcTrack in StEvent).

StRoot/StFcsDbMaker/StFcsDb.cxx
akioogawa: - It is in commet in StFcsDb.h file
   - With order=0 returns the highest contributing g2t_track_id
   - With order=1 returns the 2nd highest contributing g2t_track_id
   - Yes, I can make sorted vector available to user....But then user need to delete them.
   - Or cache sorted results somehow, but then delete when & where?
   - Any suggestions will be welcomed

StRoot/StFcsDbMaker/StFcsDb.cxx
```diff
@@ -1979,3 +1979,125 @@ void StFcsDb::readGainCorrFromText(){
...
+  if(order > ntrk) {fraction=0; return 0;}
+  std::sort(parents.begin(), parents.end(),
+	    [](const pair<unsigned int,float>&a, const pair<unsigned int,float>&b){
+	      return b.second < a.second;
+	    });
+  fraction = parents[order].second / detot;
+  return &g2ttrk[parents[order].first-1];
+}
+
+const g2t_track_st* StFcsDb::getPrimaryG2tTrack(StFcsCluster* c, g2t_track_st* g2ttrk, float& fraction, int& ntrk, unsigned int order){
```
plexoos: Akio, you have quite a few common lines of code shared between the three new functions you introduce, i.e.  `getPrimaryG2tTrack(StFcsHit* h, ...)`, `getPrimaryG2tTrack(StFcsCluster* c, ...)`, and `getParentG2tTrack(StFcsCluster* c, .. )`. Actually, the latter two are identical except one single line and the difference between the first and second is that there is a loop over hits in the cluster. This usually indicates that there might be a better way to organize the code without repeating the logic.

Have you considered splitting the functionality into a few smaller more focused functions?
E.g. we could have a function to create a mapping between the Geant track id and its deposited energy for a cluster by looping over all hits in the cluster. It could be a method of `StFcsCluster` doing something like this:

```c++
TtoEMap accumulateHits() {
  TtoEMap cluster;
  auto addit = [](TtoEMap m, TtoEMap::value_type p) { ... };

  for (auto hit : hits) {
    cluster = accumulate(begin(hit.getGeantTracks()), end(hit.getGeantTracks()), cluster, addit);
  }

  return cluster;
}
```

where `TtoEMap` is the type of `mGeantTracks`

Then, there can be a function to re-map energies from the final state tracks to primary ones.

```c++
TtoEMap remapPrimary(TtoEMap m) {
  TtoEMap m_primary;
  for (auto e : m) {
    auto id = backTraceG2tTrack(e.first, g2ttrk)
    m_primary.get_or_create( id ) += e.second;
  }
  return m_primary;
}
```

Once you construct the maps, calculating `detot` and selecting the n-th highest element is trivial by using the std functions.

akioogawa: Sorry for my ignorance. Maybe your pseudo code is not complete enough for me to understand. I guess I don't understand "TtoEMap is the type of mGeantTracks".

See what I did in the last push with getG2tTrack(StFcsCluster..., int mode=0) as a private function and calling it from getParentG2tTrack and getPrimaryG2tTrack with mode=0 & 1.

StRoot/StFcsDbMaker/StFcsDb.cxx
```diff
@@ -1979,3 +1979,107 @@ void StFcsDb::readGainCorrFromText(){
...
+  getG2tTrack(c,g2ttrk,fraction,ntrk,order,1);
+}
+
+const g2t_track_st* StFcsDb::getG2tTrack(StFcsCluster* c, g2t_track_st* g2ttrk, float& fraction, int& ntrk, unsigned int order, int mode){
+  float detot=0;
+  vector<pair<unsigned int, float>> parents;
+  StPtrVecFcsHit& hits = c->hits();  
+  int nhit = hits.size();
+  for(int ihit=0; ihit<nhit; ihit++){    
+    const vector<pair<unsigned int, float>>& gt = hits[ihit]->getGeantTracks();
```
plexoos: Is there a reason why you prefer to use indices over range for loops?

```c++
  for (const StFcsHit* hit : c->hits()) {
    const auto& gt = hit->getGeantTracks();
```

Looks less verbose

StRoot/StFcsDbMaker/StFcsDb.cxx
plexoos: > Maybe your pseudo code is not complete enough for me to understand. I guess I don't understand "TtoEMap is the type of mGeantTracks".

It means just what it says: `TtoEMap = delctype(StFcsHit::mGeantTracks)` which translates to `TtoEMap = vector<pair<unsigned, float>>` in your current code

> See what I did in the last push

Thanks. I think you will thank yourself later for that 😄  in case you need to make changes only in the private function rather than in a few places.

You can take it further from here and reuse `StFcsDb::getG2tTrack` in `StFcsDb::getParentG2tTrack(StFcsHit* h,` and `StFcsDb::getPrimaryG2tTrack(StFcsHit* h, ` by adding that single hit in a temporary "fake" cluster. I also see that there is a `StFcsCluster *StFcsHit::cluster()` method which perhaps can be used in case every hit is required to belong to a cluster.

StRoot/StFcsDbMaker/StFcsDb.cxx
akioogawa: I prefer indices over iterator & range since I often end up wanting to have random access & indexes later in the loop. But this case it may be ok with range. Change coming.

akioogawa:
I adopted your idea of fake cluster and also dropped auto sorting in StFcsFastSimulagtorMaker, leaving all sorting done in StFcsDb utility function at analysis stage.

StRoot/StFcsDbMaker/StFcsDb.cxx
```diff
@@ -1979,3 +1979,73 @@ void StFcsDb::readGainCorrFromText(){
...
+  if(mDebug>3) LOG_INFO<<Form("  BackTrace from=%3d id=%3d Epid=%4d Gpid=%3d Vtx=%3d Parent=%3d E=%6.2f Primary!!",
+			      id,g2ttrk[i].id,g2ttrk[i].eg_pid,g2ttrk[i].ge_pid,g2ttrk[i].start_vertex_p,
+			      g2ttrk[i].next_parent_p,g2ttrk[i].e)<<endm;
+  return i + 1;
+}
+    
+const g2t_track_st* StFcsDb::getParentG2tTrack(StFcsHit* h, g2t_track_st* g2ttrk, float& fraction, int& ntrk, unsigned int order){
+  StFcsCluster c;
+  c.hits().push_back(h); //dummy cluster with 1 hit
+  return getG2tTrack(&c,g2ttrk,fraction,ntrk,order,0);
```
plexoos, line 2002: As the author of StFcsHit and StFcsCluster you know better but why not the following:

```suggestion
  return getG2tTrack(h->cluster(),g2ttrk,fraction,ntrk,order,0);
```

plexoos (approved):
Thank you for addressing the comments. There is only one last suggestion from my side in the comments, I'll leave it up to you.

StRoot/StFcsDbMaker/StFcsDb.cxx
akioogawa: That will give totally different answer. A hit may not be belonging to a cluster. If It belongs to a cluster, this gives g2t track info for the cluster, not for the hit.

klendathu2k (approved):
It looks to me that all of my concerns were addressed.  Approving.

akioogawa:
Just did merge main into this branch. Any other comments?

plexoos:
Hi Daniel, looks like we need your approval to include this change in the next SL22c release

PR #380, opened 2022-07-27
docs: Add section about containers

plexoos:


veprbl (approved):
LGTM

plexoos:
> LGTM

🙇  Did you try to run the commands? One thing I don't understand is why `singularity shell` does not properly source the original `entrypoint.sh`. Maybe it is not supposed to... `singularity run ... bash -l` looks like a workaround to me.

veprbl:
> > LGTM
> 
> 🙇 Did you try to run the commands?

No. Mostly was looking at the spelling.

> One thing I don't understand is why `singularity shell` does not properly source the original `entrypoint.sh`. Maybe it is not supposed to... `singularity run ... bash -l` looks like a workaround to me.

Yeah, I believe this is what I've been observing when I tried it. I suppose one could create the `/.singularity.d/env/` [1] in Docker and it might be picked up by singularity.

[1] https://docs.sylabs.io/guides/3.7/user-guide/environment_and_metadata.html

PR #381, opened 2022-08-09
cons: Help agetof to find definition file

plexoos:
When upgrading Spack (used to build the STAR environment in containers) from v0.17.2 to v0.18.1, building with `cons` fails due to an issue with the `agetof` utility. It seems to have trouble finding the definition file `asps/Simulation/agetof/agetof.def` normally installed in `$STAR_BIN` along with the `agetof` executable.

The fix appears to be backward compatible and should work for all versions of `agetof` accepting the argument.

### DIFF
--- a/asps/Simulation/starsim/Conscript
+++ b/asps/Simulation/starsim/Conscript
@@ -127,7 +127,6 @@ $env2 = $env->clone('FC'           => $FC,
 		    'CPPFLAGS'     => $CPPFLAGS,
 		    'CPPPATH'      => $CPPPATH,
 		    'DEBUG'        => $DEBUG,
-		    'AGETOFLAGS'   => "",
 		    'LIBS'         => $LIBS,
 		    'ObjDir'       => $obj_dir,
 		    'Libraries'    => $Libraries
--- a/mgr/ConsDefs.pm
+++ b/mgr/ConsDefs.pm
@@ -210,7 +210,7 @@
     $STIC          = "stic";
     $STICFLAGS     = "";
     $AGETOF        = "agetof";
-    $AGETOFLAGS    = "-V 1";
+    $AGETOFLAGS    = "-V 1 -d $STAR_BIN/agetof.def";
     $LIBSTDC       = `$CC $CFLAGS -print-file-name=libstdc++.a | awk '{ if (\$1 != "libstdc++.a") print \$1}'`;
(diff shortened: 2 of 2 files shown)


veprbl:
> fails due to an issue with the `agetof` utility

Does that come with an error message?

plexoos:
No error message, `agetof` goes into an infinite loop and hangs.
These jobs https://github.com/star-bnl/star-spack/actions/runs/2828207965 should produce containers in ~1/2h to reproduce

klendathu2k (approved):
Agreed.  This should be completely backwards compatible.

plexoos:
> > fails due to an issue with the `agetof` utility
> 
> Does that come with an error message?

In case you want to investigate this further, use the new base images on the main branch:

```
cd star-sw
sed -i 's/v0.1.7/v0.2.0-rc1/' Dockerfile
docker build .
```

veprbl:
Before that, I was curious what you think might have caused this:
https://github.com/star-bnl/star-spack/compare/v0.1.7...v0.2.0-rc1

I'm not familiar with spack, but could it be this one
https://github.com/star-bnl/star-spack/commit/6013e547ba0720a5b951701e742fc8d7e0ee1e24
disabling the gcc 4?

plexoos:
> Before that, I was curious what you think might have caused this: [star-bnl/star-spack@v0.1.7...v0.2.0-rc1](https://github.com/star-bnl/star-spack/compare/v0.1.7...v0.2.0-rc1)

I think it is something in Spack after the upgrade from v0.17.2 to v0.18.1. Could be new versions of yacc/bison/flex? Not sure it is worth an investigation...

> I'm not familiar with spack, but could it be this one [star-bnl/star-spack@6013e54](https://github.com/star-bnl/star-spack/commit/6013e547ba0720a5b951701e742fc8d7e0ee1e24) disabling the gcc 4?

Highly unlikely. It is just a clean-up of something not used. The recent versions of STAR environment just make Spack rely on the compiler in the path rather than specifying it in Spack config files.

veprbl:
> Not sure it is worth an investigation...

I'm 100% on board with this. Only looking to lure out any details that you might already know. Just in case it may ring a bell to someone. Thanks for explaining this.

veprbl (approved):
Providing a file with a command argument instead of auto-discovery seems like a fine solution.

PR #382, opened 2022-08-09
cons: Pass PYTHONPATH to build environment

plexoos:
This variable is not utilized at the moment but is handy when python and
python modules are not installed in the standard location, e.g. when
installed via Spack.


klendathu2k (approved):
Shouldn't be an issue to support non-standard python paths...

veprbl:
This is needed to run the agml parser, right?

plexoos:
Yes, I believe it is the only reason we need python2 as a build-time dependency.

veprbl (approved):

PR #383, opened 2022-08-12
Add calculation dX in TPC using local curvature

fisyak:


starsdong:
Yuri, my understanding is that this request is to add a user function for dX calculation. Seem to be straightforward to me. Jason/Dmitri, any further comments?

klendathu2k:
> Yuri, my understanding is that this request is to add a user function for dX calculation. Seem to be straightforward to me. Jason/Dmitri, any further comments?

To first order this looks fine.  My only concern is whether there are other codes which write the dX value in StTpcHit?  If so, what is the reason for replacing the calculation?  (Or is the calculation just being moved into StiStEventFiller?)

fisyak:
As it was explained in message this method used local curvature for each hit instead of using curvature from the first and last hit. The only place where we can access local curvature (StKalanTrackNode parameters) is StiStEventFiller. This is essential for low pT tracks.
"

starsdong (approved):

starsdong:
If no other objections, I would suggest to merge this PR. Thanks

veprbl (approved):
Diff LGTM

genevb:
Is the time required for this extra pair of pathlength calculations for every TPC hit on a track negligible?

StRoot/StiMaker/StiStEventFiller.cxx
```diff
@@ -1470,6 +1471,32 @@ void StiStEventFiller::fillDca(StTrack* stTrack, StiKalmanTrack* track)
...
+  originD->setZ(node->z_g());
+
+  physicalHelix->setParameters(fabs(node->getCurvature()),
+			       node->getDipAngle(),
+			       node->getPhase(),
+			       *originD,
+			       node->getHelicity());
+  StThreeVectorD upper(tpcHit->positionU().x(),tpcHit->positionU().y(),tpcHit->positionU().z());
+  StThreeVectorD lower(tpcHit->positionL().x(),tpcHit->positionL().y(),tpcHit->positionL().z());
+  StThreeVectorD middle(tpcHit->position().x(),tpcHit->position().y(),tpcHit->position().z());
```
plexoos, line 1490: This variable is unused

```suggestion
```

klendathu2k:
A quick grep shows that StTpcHit::setdX is called from StdEdxY2Maker.  

I presume that dEdxY2 runs after StiStEventFiller.  So the dX values will be overwritten when it runs.  Don't we need to be able to shut this new calculation off in order to maintain the ability to reproduce old results?

starsdong:
Yuri, do you have any response to Jason's question? And regarding the time consumption raised by Gene, have you tested it also? Thanks

plexoos:
> Yuri, do you have any response to Jason's question? And regarding the time consumption raised by Gene, have you tested it also?

Was there any response to these questions?

There was also an unaddressed comment about a defined but unused variable in the submitted code. Why would we want to have it in the code? Doesn't it create a warning at compile time?

PR #387, opened 2022-08-27
Fix memory leaks when MuDst included with FWD data. Introduce a switch for FTT calibration mode (dump values to file)

jdbrice:
Akio found and fixed several memory leaks.
Fwd MuDst data structures (Fst, Ftt, Fcs) were causing memory leaks that could cause crash of BFC due to running out of memory. Memory leaks appeared to happen from StEvent structures, because MuDst code made a copy of StEvent containers.

Update uses const references to avoid any copy and subsequent leak.
Also includes one memory leak fix / small update on StFttHitCalibMaker to prepare for future calibration productions.


akioogawa (approved):

StRoot/StFttHitCalibMaker/StFttHitCalibMaker.cxx
```diff
@@ -22,9 +22,11 @@ StMaker("fttHitCalib",name), mHelper(nullptr)
...
 }
 //_____________________________________________________________                                                       
-// StFttHitCalibMaker::~StFttHitCalibMaker()
-// { 
-// }
+StFttHitCalibMaker::~StFttHitCalibMaker()
+{
+  if(mHelper) delete mHelper;
+  mHelper = nullptr;
+}
```
plexoos, line 29: Do I understand correctly that only this change in this commit (9b955f7637aad4795c27223825de25c9d4185a22) is related to the memory leak fix? The rest seems to be about "calib mode for StFttDb timeCut".

StRoot/StEvent/StFcsHit.h
```diff
@@ -99,7 +99,7 @@ class StFcsHit : public StObject {
     UInt_t   mNPeak=0;        // number of peaks found
     Float_t  mEnergy=0.0;     // corrected energy
     StFcsCluster* mCluster=0; // pointer to cluster this hit belongs
-    TArrayS* mData=0;         // 12bit ADC values + flag at highest 4 bits, array of timebin
+    TArrayS* mData=0;         //-> 12bit ADC values + flag at highest 4 bits, array of timebin
```
plexoos: Is this used/needed by rootcint?

starsdong (approved):

StRoot/StFttHitCalibMaker/StFttHitCalibMaker.cxx
jdbrice: [StRoot/StFttHitCalibMaker/HitCalibHelper.h](https://github.com/star-bnl/star-sw/pull/387/files#diff-498e5549bf293b766cf00cd0cfc0c5598244fad0a24e4f01fe1273abad0d0aa9)
Is a bug-fix, but not specifically memory related. The previous algorithm was incorrect. This algorithm does the correct thing, which is to count the number and offset of signals per channel in the detector since that is what we use to calibrate the timing. 

[StFttDb.cxx](https://github.com/star-bnl/star-sw/pull/387/files#diff-6968429fff05ed2873b738c6fd73f1e2192d7cbce9bf0eb387df5ffe98639905)
and
[StFttDb.h](https://github.com/star-bnl/star-sw/pull/387/files#diff-cd9d3c75e76ebd0b314286b71390c8640a9ebd77d8ebfbbd9e01ce754373a406)

fix the fact that this was unusable for calibration (generally without a pre-filled DB) since it was setup to read timing cuts from DB. So a switch was added for when no DB parameters exist, and we want to run in calibration mode with a user-defined timing cut.

StRoot/StEvent/StFcsHit.h
jdbrice: @akioogawa do you have any comment on this?

StRoot/StFttDbMaker/StFttDb.h
```diff
@@ -113,11 +115,29 @@ class StFttDb : public TDataSet {
...
+      h = mTimeCutHigh;
+      // mode = StFttDb::TimeCutMode::CalibratedBunchCrossingMode;
+      // l = -3;
+      // h = 3;
+      //mode = StFttDb::TimeCutMode::TimebinMode;
       // ideal; take +/- 3 bunch crossings
-      l = -3;
-      h = 3;
+      //l = -70;
+      //h = 30;
```
plexoos: This function `void getTimeCut( StFttRawHit * hit, int &mode, int &l, int &h )` takes a pointer to StFttRawHit. But I don't see it being used. If that is how the function is intended to work, shouldn't the argument be removed?

StRoot/StFttHitCalibMaker/StFttHitCalibMaker.cxx
plexoos: Thank you for the explanation. So, the title of this PR should be updated to reflect the second change unrelated to the memory fix.

StRoot/StFttHitCalibMaker/StFttHitCalibMaker.cxx
```diff
@@ -49,9 +51,31 @@ Int_t StFttHitCalibMaker::FinishRun(Int_t runnumber)
 Int_t StFttHitCalibMaker::Finish()
 { 
     LOG_INFO << "StFttHitCalibMaker::Finish()" << endm;
+
+    if (this->mCalibMode == StFttHitCalibMaker::CalibMode::Calibration) {
+        LOG_INFO << "Writing StFttHitCalib parameters to plaintext: " << endm;
+        // this->WriteCalibrationToPlainText();
+    }
+    this->WriteCalibrationToPlainText();
```
plexoos: This needs a confirmation that `WriteCalibrationToPlainText()` should be called regardless of `mCalibMode` value. If it is so, the log message is not relevant.

Also, we don't use `this->` for class members.

jdbrice: Fixed in 28da7017d2f9ed86b953205c21800711d2968e3e

StRoot/StFttDbMaker/StFttDb.h
jdbrice: You are right, somehow I lost the original DB based code that looks up a unique cut for each hit.
Fixed in 28da7017d2f9ed86b953205c21800711d2968e3e
In 34e97f28ccce8640e0823eadb939929b867a9b9a and 068ff9653e5458200175f7b790ecb85844eee436

I made sure that the default is production mode (read from DB) and that the flag mUserDefinedTimeCut is set whenever a time cut is applied by the user (via setTimeCut(...) )

StRoot/StEvent/StFcsHit.h
jdbrice: Removed in 28da7017d2f9ed86b953205c21800711d2968e3e

StRoot/StEvent/StFcsHit.cxx
```diff
@@ -93,7 +93,7 @@ void StFcsHit::setData(int ntimebin, const unsigned short* data) {
     }
     else {
         mData->Set(ntimebin,(const short*)data);
-  }
+    }
```
klendathu2k, line 96: I think this is the piece of code that I own that requires my approval....

StRoot/StMuDSTMaker/COMMON/StMuFcsUtil.cxx
```diff
@@ -72,7 +72,7 @@ void StMuFcsUtil::fillMuFcsHits(StMuFcsCollection* muFcs,
                                 StFcsCollection* fcscol) {
   
     for ( unsigned int idet = 0; idet < kFcsNDet+1; idet++ ){
-        StSPtrVecFcsHit vecHit = fcscol->hits(idet);
```
klendathu2k, line 75: This access pattern is important.  Accessing the hits by copy results in the memory leak.  Accessing by reference does not.

klendathu2k (approved):

plexoos (approved):
Thank you for addressing the comments. Looks good to me

PR #393, opened 2022-09-07
SpaceCharge distortion corrections for FXT

genevb:
Introduce distortion codes and steering routines to handle SpaceCharge distortions in BES-II FixedTarget data. Includes some very minor re-spacing of code text to keep it looking clean.

### DIFF
--- a/StRoot/StDbUtilities/StMagUtilities.cxx
+++ b/StRoot/StDbUtilities/StMagUtilities.cxx
@@ -2941,12 +2941,6 @@ void StMagUtilities::UndoSpaceChargeFXTDistortion( const Float_t x[], Float_t Xp
   Float_t   Er_integral, Ephi_integral ;
   Double_t  r, phi, z ;
 
-  // re-use these two radial E-field arrays from the R0 and R2 distributions
-  // since the R0 and R2 corrections cannot be used at the same time as FXT
-  // (someday better to replace these arrays with STL, but that's a lot to change)
-#define spaceFXTErE spaceEr
-#define spaceFXTErW spaceR2Er
-
   if (fSpaceChargeR2) { GetSpaceChargeR2();} // need to reset it; re-use the R2 database table and access
 
   if ( DoOnce )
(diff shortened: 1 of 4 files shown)


plexoos:
Is this intended for the upcoming SL22c release?

genevb:
> Is this intended for the upcoming SL22c release?

It is critical for processing the Run 21 FXT data. If not in SL22c now (which is my preference), then we would probably need either  an SL22d within a month, or we could have a second version of SL22c as this PR has no impact whatsoever on any existing nor soon-to-be-used chains .

Thanks,
-Gene

plexoos:
We can include this in SL22c. I recall we expected some updates for TPC dE/dx and forward tracking. Was it something beyond the merged #383 and #387?

StRoot/StDbUtilities/StMagUtilities.cxx
```diff
@@ -2906,6 +2925,123 @@ void StMagUtilities::UndoSpaceChargeR2Distortion( const Float_t x[], Float_t Xpr
...
+  const Int_t     ORDER       =    1 ;  // Linear interpolation = 1, Quadratic = 2
+
+  Float_t   Er_integral, Ephi_integral ;
+  Double_t  r, phi, z ;
+
+  // re-use these two radial E-field arrays from the R0 and R2 distributions
+  // since the R0 and R2 corrections cannot be used at the same time as FXT
+  // (someday better to replace these arrays with STL, but that's a lot to change)
+#define spaceFXTErE spaceEr
+#define spaceFXTErW spaceR2Er
```
plexoos: Unless I am missing something, I don't see how these two macros `spaceFXTErE` and `spaceFXTErW` are being used. Are they somehow used by a macro not in the repository?

StRoot/StDbUtilities/StMagUtilities.cxx
```diff
@@ -2906,6 +2925,123 @@ void StMagUtilities::UndoSpaceChargeR2Distortion( const Float_t x[], Float_t Xpr
...
+    for ( Int_t i = 0 ; i < ROWS ; i++ )
+      {
+        Double_t Radius = IFCRadius + i*GRIDSIZER ;
+        ArrayVE(i,j) = 0 ;
+        ChargeE(i,j) = 0 ;
+        ArrayVW(i,j) = 0 ;
+        ChargeW(i,j) = 0 ;
+        Rlist[i] = Radius ;
+      }
+  }
```
plexoos: You use a very strange indentation for the loop body here and below. Please fix.

StRoot/StDbUtilities/StMagUtilities.cxx
```diff
@@ -2664,8 +2679,10 @@ void StMagUtilities::UndoIFCShiftDistortion( const Float_t x[], Float_t Xprime[]
 void StMagUtilities::UndoSpaceChargeDistortion( const Float_t x[], Float_t Xprime[] , Int_t Sector )
 {
 
-  if ((mDistortionMode & kSpaceCharge) && (mDistortionMode & kSpaceChargeR2)) {
-      cout << "StMagUtilities ERROR **** Do not use kSpaceCharge and kSpaceChargeR2 at the same time" << endl ;
+  if (((mDistortionMode/kSpaceCharge)    & 1) +
+      ((mDistortionMode/kSpaceChargeR2)  & 1) +
+      ((mDistortionMode/kSpaceChargeFXT) & 1) > 1) {
```
plexoos: I would suggest to keep the bitwise & operator as before in order to check the bits against the bit masks. Instead of the logical && you can use the addition + operator to test the mutually exclusive bits

```suggestion
  if ((mDistortionMode & kSpaceCharge) +
      (mDistortionMode & kSpaceChargeR2) +
      (mDistortionMode & kSpaceChargeFXT) > 1) {
```

StRoot/StDbUtilities/StMagUtilities.cxx
genevb: Oops...didn't mean to leave those in there. I'll remove them.

StRoot/StDbUtilities/StMagUtilities.cxx
genevb: I copied from a similar previous function. I can fix it in both places.

StRoot/StDbUtilities/StMagUtilities.cxx
genevb: I don't think this works. Each of the kSpaceChargeXXX constants have a value greater than 1, so any of the three being active would result in a sum greater than 1. I'm testing for whether more than one of the three are active.

plexoos: That's right. The bitwise & results in an integer and the comparison of the sum with 1 does not make sense.

Since you expect only one of kSpaceChargeXXX flags to be set then their "union" should give only one bit set to 1, i.e. the number should be a power of 2. So, perhaps something like this can work

```suggestion
  int scopt = (mDistortionMode & kSpaceCharge) |
      (mDistortionMode & kSpaceChargeR2) |
      (mDistortionMode & kSpaceChargeFXT);
  if (scopt & (scopt - 1)) { // more than one specified
```

Just a suggestion.

plexoos (approved):

StRoot/StDbUtilities/StMagUtilities.cxx
genevb: Interesting suggestion, and I believe that works, though one can shorten it to use:

    int scopt = (mDistortionMode & (kSpaceCharge | kSpaceChargeR2 | kSpaceChargeFXT));

It's a subjective call whether that's prettier/clearer....unless....you don't trust `/` as a division operator to return a quotient that is guaranteed not to round up if the remainder is non-zero? As long as it does not round up, it certainly works for cases where the divisor has only a single non-zero bit.  So if you are worried about rounding up, here is an [alternative](https://en.cppreference.com/w/cpp/numeric/math/div) to `/` that must work because it is defined as the quotient (though it is moving towards a less pretty/clear solution):

  if (((std::div(mDistortionMode,kSpaceCharge   )).quot & 1) +
      ((std::div(mDistortionMode,kSpaceChargeR2) ).quot & 1) +
      ((std::div(mDistortionMode,kSpaceChargeFXT)).quot & 1) > 1) {
      ... }

Again, the only reason this could fail is if any of the kSpaceChargeXXX variables has more than one non-zero bit. But there's no reason to explore such cases for this situation.

Perhaps the best reason I can think of to use your solution is that it probably takes fewer CPU cycles, even though this is negligible here. But, I'll go ahead with your solution and be done with this.

-Gene

plexoos: > It's a subjective call whether that's prettier/clearer....unless....you don't trust `/` as a division operator to return a quotient that is guaranteed not to round up if the remainder is non-zero?

Using the division must be fine in this case too since we divide by a power of 2. This is simply equivalent to a certain number of binary shifts to the right. I guess if I saw the bitwise shift operator this piece of code wouldn't even trigger me :)

StRoot/StDbUtilities/StMagUtilities.cxx
```diff
@@ -2906,6 +2924,117 @@ void StMagUtilities::UndoSpaceChargeR2Distortion( const Float_t x[], Float_t Xpr
...
+/// FixedTarget 1/r SpaceCharge Distortion
+/*!
+  Similar to R2 distortion correction for SpaceCharge, but using distinct radiual distributions on
+  the east and west sides as determined by studies of URQMD simulations of fixed target
+  collisions conducted by Yue-Hang Leung in June 2022.
+*/
+void StMagUtilities::UndoSpaceChargeFXTDistortion( const Float_t x[], Float_t Xprime[] , Int_t Sector )
+{
+  
+  const Int_t     ORDER       =    1 ;  // Linear interpolation = 1, Quadratic = 2
```
klendathu2k, line 2938: per STAR naming conventions, prefer `int` to `Int_t`, `double` to `Double_t`, etc...

klendathu2k (commented):
After initial read through I don't see any major issues.  Will look through again tomorrow.  Would recommend a dedicated "clean up" PR to address STAR coding guildelines comment (Int_t --> int, etc...).

Issue #392, opened 2022-09-02
Get rid of compiler warnings


plexoos:
Our build system currently asks the compiler to print many useful warnings (-Wall). This sounds like a good idea since warnings may indicate problematic places in the code, obsolete features, or the use of nonstandard extension of C/C++. However, it appears that the reported warnings have been ignored for quite a long time. Currently there are about 3,000 warnings reported by gcc 4.8.5. A goal to address all the warnings should be recognized and a zero-warning policy should be enforced for all future code modifications.

A list of warnings sorted by their frequency in the build log (gcc 4.8.5, root 5):

```
                                                                      msg  count   freq
0         deprecated conversion from string constant to [-Wwrite-strings]    798  27.0%
1      may be used uninitialized in this function [-Wmaybe-uninitialized]    435  14.7%
2   deleting object of polymorphic class type which has non-virtual de...    403  13.6%
3                   variable set but not used [-Wunused-but-set-variable]    341  11.5%
4                                     unused variable [-Wunused-variable]    293   9.9%
5      format expects argument of type, but argument has type [-Wformat=]    210   7.1%
6   comparison between signed and unsigned integer expressions [-Wsign...     73   2.5%
7   iteration invokes undefined behavior [-Waggressive-loop-optimizati...     56   1.9%
8                        operation on may be undefined [-Wsequence-point]     54   1.8%
9                           declaration of shadows a member of [-Wshadow]     41   1.4%
10           converting to non-pointer type from NULL [-Wconversion-null]     30   1.0%
11                                  unused parameter [-Wunused-parameter]     25   0.8%
12                       ignoring #pragma omp section [-Wunknown-pragmas]     22   0.7%
13                                      was hidden [-Woverloaded-virtual]     20   0.7%
14                 array subscript is above array bounds [-Warray-bounds]     14   0.5%
15  dereferencing type-punned pointer will break strict-aliasing rules...     13   0.4%
16                               defined but not used [-Wunused-variable]     12   0.4%
17                                  will be initialized after [-Wreorder]     11   0.4%
18                               defined but not used [-Wunused-function]      9   0.3%
19                unknown conversion type character in format [-Wformat=]      8   0.3%
20               control reaches end of non-void function [-Wreturn-type]      7   0.2%
21  argument to in call is the same pointer type as the destination; e...      7   0.2%
22                    too many arguments for format [-Wformat-extra-args]      7   0.2%
[...]

PR #395, opened 2022-09-13
Fix GCC warning: unused variables

plexoos:
Getting rid of such warnings as

- variable set but not used [-Wunused-but-set-variable]
- unused variable [-Wunused-variable]
- unused parameter [-Wunused-parameter]
- etc.

See https://github.com/star-bnl/star-sw/issues/392 for motivation

### DIFF
--- a/StRoot/StBTofPool/alignment/analysis.cxx
+++ b/StRoot/StBTofPool/alignment/analysis.cxx
@@ -41,7 +41,7 @@ using namespace std;
 int main(int argc, char **argv)
 {
   if(argc !=2) exit(1);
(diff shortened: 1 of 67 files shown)


genevb:
What's the criteria for commenting out lines vs. deleting them?

plexoos:
> What's the criteria for commenting out lines vs. deleting them?

When I see that a variable is "used" in some other already commented lines then I rather comment it out as well, otherwise remove.

plexoos:
Progress so far. This PR vs SL22 branch

By warning:
fix_warns_warn

By package (40 with highest count selected, RTS and Jevp are simply disabled):
fix_warns_path

PR #396, opened 2022-09-13
BigFullChain, updates for FWD detectors and Tracking

jdbrice:
Adds a few BFC options for FWD chains.
Also, makes FTT options use consistent case.

PR #397, opened 2022-09-13
Addition of StFwdTrack and StFwdTrackCollection to StEvent

jdbrice:
Builds on FWD upgrade PR 396 (BFC additions)

cons +StRoot/StEvent succeeds but just cons fails with inability to find the StPtrVecFwdTrack based on new StFwdTrack class

### DIFF
--- a/StRoot/StEvent/StFwdTrack.cxx
+++ b/StRoot/StEvent/StFwdTrack.cxx
@@ -20,7 +20,7 @@ StFwdTrack::StFwdTrack( genfit::Track *t ) : mGenfitTrack(t), mProjections(0) {
 
 }
 
-const bool StFwdTrack::didFitConverge() const {
+bool StFwdTrack::didFitConverge() const {
     if (!mGenfitTrack)
         return false;
     // uses the default track rep
@@ -30,7 +30,7 @@ const bool StFwdTrack::didFitConverge() const {
     return fitStatus->isFitConverged();
 }
 
-const bool StFwdTrack::didFitConvergeFully() const {
+bool StFwdTrack::didFitConvergeFully() const {
     if (!mGenfitTrack)
         return false;
     // uses the default track rep
@@ -40,7 +40,7 @@ const bool StFwdTrack::didFitConvergeFully() const {
     return fitStatus->isFitConvergedFully();
 }
 
-const int StFwdTrack::numberOfFailedPoints() const {
+int StFwdTrack::numberOfFailedPoints() const {
     if (!mGenfitTrack)
         return false;
     // uses the default track rep
@@ -50,7 +50,7 @@ const int StFwdTrack::numberOfFailedPoints() const {
     return fitStatus->getNFailedPoints();
 }
 
-const double StFwdTrack::chi2() const {
+double StFwdTrack::chi2() const {
     if (!mGenfitTrack)
         return false;
     // uses the default track rep
@@ -60,7 +60,7 @@ const double StFwdTrack::chi2() const {
     return fitStatus->getChi2();
 }
 
-const double StFwdTrack::ndf() const {
+double StFwdTrack::ndf() const {
     if (!mGenfitTrack)
         return false;
     // uses the default track rep
@@ -70,7 +70,7 @@ const double StFwdTrack::ndf() const {
     return fitStatus->getNdf();
 }
 
-const double StFwdTrack::pval() const {
+double StFwdTrack::pval() const {
     if (!mGenfitTrack)
         return false;
     // uses the default track rep
@@ -83,7 +83,7 @@ const double StFwdTrack::pval() const {
 /* momentum
  * get the track momentum at the first point (PV if included)
  */
-const StThreeVectorD StFwdTrack::momentum() const{
+StThreeVectorD StFwdTrack::momentum() const{
     if (!mGenfitTrack)
         return StThreeVectorD( 0, 0, 0 );
     auto cr = mGenfitTrack->getCardinalRep();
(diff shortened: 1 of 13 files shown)


genevb:
Commit [2052ab7](https://github.com/star-bnl/star-sw/pull/397/commits/2052ab789919c60fcba13b2fd2f02ea5674537b8) overlaps with #396 ?

jdbrice:
> Commit [2052ab7](https://github.com/star-bnl/star-sw/pull/397/commits/2052ab789919c60fcba13b2fd2f02ea5674537b8) overlaps with #396 ?

Yes, three PRs in a row that are sub branches of each other. 

396: Big full chain
397: StEvent additions
398: FCS Track match maker

If i understand git properly, after the first (396) is merged to main, the later ones will have those changes disappear but will inherit any changes we make during this review

StRoot/StEvent/StFwdTrack.h
```diff
@@ -0,0 +1,101 @@
...
+        return sqrt( cov[8] );
+    }
+};
+
+class StFwdTrack : public StObject {
+
+public:
+    StFwdTrack( genfit::Track * );
+    vector<StFwdTrackProjection> mProjections;
+    genfit::Track *mGenfitTrack;
```
plexoos, line 57: Let's wait for Jason to chime in but I think this dependency contradicts the original concept of StEvent

StRoot/StEvent/StFwdTrack.cxx
```diff
@@ -0,0 +1,147 @@
...
+    return fitStatus->getNFailedPoints();
+}
+
+const double StFwdTrack::chi2() const {
+    if (!mGenfitTrack)
+        return false;
+    // uses the default track rep
+    auto fitStatus = mGenfitTrack->getFitStatus();
+    if ( !fitStatus ) 
+        return false;
```
veprbl, line 59: Return `false` from double? Should those checks be moved to constructor/factory?

StRoot/StEvent/StFwdTrack.h
```diff
@@ -0,0 +1,101 @@
...
+    const unsigned int   numberOfFitPoints() const;
+    // const unsigned int   numberOfPossibleFitPoints() const;
+
+    // Number of points used in the track seed step
+    // const unsigned int   numberOfSeedPoints() const;
+
+    // momentum at the primary vertex
+    const StThreeVectorD momentum() const;
+    const StThreeVectorD momentumAt(int _id = 1) const;
+    const char charge() const;
```
veprbl: Too much const
```suggestion
    bool   didFitConverge() const;
    bool   didFitConvergeFully() const;
    int    numberOfFailedPoints() const;
    double chi2() const;
    double ndf() const;
    double pval() const;

    // error on pT upper and lower 1-sigma values
    // add access to cov matrix

    // Number of fit points used by GenFit
    unsigned int   numberOfFitPoints() const;
    // unsigned int   numberOfPossibleFitPoints() const;

    // Number of points used in the track seed step
    // unsigned int   numberOfSeedPoints() const;

    // momentum at the primary vertex
    StThreeVectorD momentum() const;
    StThreeVectorD momentumAt(int _id = 1) const;
    char charge() const;
```

StRoot/StEvent/StFwdTrack.cxx
```diff
@@ -0,0 +1,147 @@
...
+        id = id -1;
+        if ( id < 0 ){
+            LOG_WARN << " Cannot get momentum at point: " << _id << endm;
+            return StThreeVectorD( 0, 0, 0 );
+        }
+    } // while valid point
+
+    auto cr = mGenfitTrack->getCardinalRep();
+    TVector3 p = cr->getMom(mGenfitTrack->getFittedState(id, cr));
+    LOG_INFO << "Momentum at point: " << id << endm;
```
veprbl, line 116: Sounds like `LOG_DEBUG` at most.

StRoot/StEvent/StFwdTrack.h
jdbrice: OK, but just a reminder. We discussed the dependency issue in a few S&C meetings, and I implemented this after those discussions

StRoot/StEvent/StFwdTrack.h
jdbrice: why would you remove the const qualifier?

jdbrice: Do you mean it is redundant when the function is a const?

veprbl: You pass POD (bool,char,int,double,...), not by reference, so they are always copied by value, so const qualifier doesn't mean anything (there is nothing a caller can modify).

The StThreeVectorD is also not a container (i.e., we assume, it doesn't hold references inside), so, again, you have it copied all of its information over -- there is nothing to modify (unless you want to return `const StThreeVectorD&`, which is fine by me).

veprbl: So, yes, it's redundant, but for a different reason.

StRoot/StEvent/StFwdTrack.h
klendathu2k: Chiming in...  We don't add external package dependencies to StEvent. 

But looking at StFwdTrack.h... other than exposing genfit::Track as public... why do you need to store the genfit track?
It looks like you're just calculating / returning values which can be computed in StEventFiller and saved in member variables of StFwdTrack.

StRoot/StEvent/StFwdTrack.h
plexoos: Thank you @veprbl for the patch. I am committing it and resolving

starsdong:
Hi Daniel, just want to follow up on this? Based on our discussion two weeks ago
https://drupal.star.bnl.gov/STAR/event/2022/11/09/SC-management-meeting
the plan is that you will update this PR following the new strategy. So do we expect this to come soon before we take further action on this PR? Thanks

starsdong:
This PR is superseded by PR#492.

PR #400, opened 2022-09-14
Separate out FCS cluster seed thresholds

### LINKED ISSUE #394: StFcsClusterMaker : Make cluster seed threshold separate
Make cluster seed threshold separate from tower threshold in StFcsClusterMaker

akioogawa:
Separate out FCS cluster seed threshold from tower threshold to be a part of cluster.

Also trying another workflow
  Issue with a branch at STAR main repo -> PR to STAR main
instead of 
  a branch at my repo/fork -> PR to STAR main
although not sure if former is allowed...

### DIFF
--- a/StRoot/StFcsClusterMaker/StFcsClusterMaker.cxx
+++ b/StRoot/StFcsClusterMaker/StFcsClusterMaker.cxx
@@ -115,6 +115,7 @@ int StFcsClusterMaker::Make() {
       if(det==0){
 	mNeighborDistance = mNeighborDistance_Ecal;
 	mDistanceAdvantage = mDistanceAdvantage_Ecal;
+	mTowerEThreSeed = mTowerEThreSeed_Ecal;
 	mTowerEThreshold = mTowerEThreshold_Ecal;
 	mTowerEThreMoment = mTowerEThreMoment_Ecal;
 	mTowerERatio2Split = mTowerERatio2Split_Ecal;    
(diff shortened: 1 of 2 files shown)


veprbl (approved):
Makes sense

plexoos:
> a branch at STAR main repo

It is allowed and you can create a PR from it as you just did. Anyone with write permission can. Is there a good reason to prefer that workflow over the one with your fork?

plexoos (approved):
I assume this changes the algorithm creating FCS clusters. But I also believe it was never used in any processing chain according to this page https://www.star.bnl.gov/devcgi/dbProdOptionRetrv.pl (only "pp2022" depends on the "fcs" option). If this is not the case then how could one reproduce the same clusters without preserving the old logic?

akioogawa:
> It is allowed and you can create a PR from it as you just did. Anyone with write permission can. Is there a good reason to prefer that workflow over the one with your fork?

Not really for this PR. I was at EPIC software tutorials and they recommended to use "Issues" and associated branch on main GitHub (not user's fork). They said this is to make everyones aware of work in progress by others, so that we avoid double work and better working together if the issue/PR is worked on by multiple people. I guess disadvantage is it can create too many branches at main Github. I guess I just wanted to see how it would work out. Either way looks ok to me.

akioogawa:
> I assume this changes the algorithm creating FCS clusters. But I also believe it was never used in any processing chain according to this page https://www.star.bnl.gov/devcgi/dbProdOptionRetrv.pl (only "pp2022" depends on the "fcs" option). If this is not the case then how could one reproduce the same clusters without preserving the old logic?

It is the case. We are just about to start test production for run22. One can put new seed threshold equal or lower than old tower thresholds, logic goes back to old one.

akioogawa:
Daniel? We need your approval :-)

plexoos:
Akio, do you prefer to wait for the approval or proceed with merging? I think this PR's been pending long enough.

akioogawa:
Hi

go ahead!

OGAWA, akio		    Email : ***@***.***
BrookHaven National Lab. Physics Dep. 510A 2-235
Upton, NY, 11973-5000	       Tel. 631-344-5293

PR #401, opened 2022-09-16
Update to StFcsWaveformFitMaker to use new algorithm

dkapukchyan:
This is an update to the StFcsWaveformFitMaker algorithm to improve the peak finding method and energy determination from the raw signal data.


StRoot/StFcsWaveformFitMaker/StFcsWaveformFitMaker.cxx
```diff
@@ -151,51 +238,115 @@ int StFcsWaveformFitMaker::InitRun(int runNumber) {
...
       gStyle->SetOptStat(0);
+      char file[100];
+      if( mFitDrawOn>=0 ){
+	sprintf(file,"%s.pdf[",mFilename);//Opens fileA but does not save anything
+	mCanvas->Print(file);
+      }
     }
     if(mMeasureTime){
-      mTime=new TH1F("FitTime","FitTime; msec",200,0,6);
+      mTime=new TH1F("FitTime","FitTime;msec",1000,0,1000);
```
plexoos: I am using this line to comment on a common problem which this PR seems to introduce in quite a few other places.
The mTime histogram can be allocated multiple times during a job when the framework calls InitRun(), however, the corresponding object is deleted only once in the destructor, hence causing a memory leak. Depending on the logic you are trying to implement here, you may need to allocate histograms only once in the Init() method. It may be also easier to rely on the automatic object lifetime than trying to manage the memory manually like you do.

dkapukchyan: I thought InitRun was only called once during a job, which is why I put the histogram creations there. I agree that I only need these histograms created once per job. Where should I move the histogram allocations so that they only get created once? Hopefully I can move these before this pull request is made so that I don't need to make a new one.

akioogawa: Move to Init(). In principle, InitRun can be called multiple times per job if you are reading file(s) from multiple runs.

StRoot/StFcsWaveformFitMaker/PeakWindow.cxx
```diff
@@ -0,0 +1,428 @@
...
+  mPeakY = -1;
+
+  mStartLine=0;
+  mPeakMarker=0;
+  mEndLine=0;
+}
+
+PeakWindow::PeakWindow(const PeakWindow &oldpeak)
+{
+  ((PeakWindow&)oldpeak).Copy(*this);
```
plexoos:
```suggestion
  oldpeak.Copy(*this);
```

StRoot/StFcsWaveformFitMaker/PeakWindow.cxx
```diff
@@ -0,0 +1,428 @@
+#include "PeakWindow.h"
+
+ClassImp(PeakWindow)
+
+PeakWindow::PeakWindow()
```
plexoos: Please use a delegating constructor to reduce the boilerplate code

StRoot/StFcsWaveformFitMaker/PeakAnaVirtualPainter.h
```diff
@@ -0,0 +1,70 @@
...
+Author: David Kapukchyan
+@[March 9, 2022]
+> Virtual interface for PeakAnaPainter. This is needed since the non-virtual class runs into "cross-dependency" issues during compilation. The idea is that a virtual interface will get rid of this problem.
+> The design is inspired/copied from ROOT's TVirtualHistPainter.
+
+PeakAnaPainter abstract base class
+*/
+
+#ifndef STROOT_STFCSWAVEFORMFITMAKER_PEAKANAVIRTUALPAINTER_H_
+#define STROOT_STFCSWAVEFORMFITMAKER_PEAKANAVIRTUALPAINTER_H_
```
plexoos: STAR coding guidelines https://star-bnl.github.io/star-coding-guidelines/codingguide.xml#The__define_Guard

StRoot/StFcsWaveformFitMaker/PeakAnaPainter.h
```diff
@@ -0,0 +1,106 @@
...
+#include "TObject.h"
+#include "TKey.h"
+#include "TH1.h"
+
+//Custom Headers
+#include "PeakWindow.h"
+#include "PeakAnaVirtualPainter.h"
+#include "PeakAna.h"
+
+class PeakAnaPainter : public PeakAnaVirtualPainter
```
plexoos: Besides this class PeakAnaPainter, is there another class that inherits from PeakAnaVirtualPainter? If not, are you sure you need to create the hierarchy with virtual inheritance for this simple case?

StRoot/StFcsDbMaker/StFcsDbMaker.h
```diff
@@ -107,6 +108,7 @@ class StFcsDbMaker : public StMaker {
 private:
   StFcsDb *mFcsDb;
   int mDbAccess=1;
+  StFcsDbPulse* mFcsDbPulse;
```
plexoos, line 111: Technically, you need to increase the version in ClassDef when you change the class layout. But this is a maker, typically makers don't utilize streamers, so you can fix this class:

```diff
-  ClassDef(StFcsDbMaker,4)   //StAF chain virtual base class for Makers        
+  ClassDef(StFcsDbMaker,0)   //StAF chain virtual base class for Makers        
```

StRoot/StFcsDbMaker/StFcsDbPulse.cxx
```diff
@@ -0,0 +1,185 @@
...
+      mA1        = 2543.0/854.0/mGSigma;
+      mA2        = 0.0;
+      mXoff1     = 211.3-215.7;
+      mXoff2     = 0.0;
+      mTau1      = 4.375;
+      mTau2      = 0.0;
+      mP1        = 1.0;
+      mP2        = 0.0;
+    }
+  else{ std::cout << "WARNING:Invalid Tail value" << std::endl; }
```
plexoos: See https://star-bnl.github.io/star-coding-guidelines/formatguide.xml#Printing_Messages_in_STAR

StRoot/StFcsDbMaker/StFcsDbPulse.h
```diff
@@ -0,0 +1,87 @@
...
+  double mTau1;
+  double mTau2;
+  double mP1;
+  double mP2;
+
+ private:
+  int mTail;
+  static const int mAdcSaturation = 4000;
+
+  ClassDef(StFcsDbPulse, 1.0);
```
plexoos: Integer values should be used

```suggestion
  ClassDef(StFcsDbPulse, 1);
```

StRoot/StFcsWaveformFitMaker/PeakAna.cxx
```diff
@@ -0,0 +1,1336 @@
...
+    std::copy( mFilterWeights,mFilterWeights+(2*mFilterScale+1), ((PeakAna&)obj).mFilterWeights );
+  }
+  else{ ((PeakAna&)obj).mFilterWeights = 0; }
+  ((PeakAna&)obj).mTunnelScale = mTunnelScale;
+  ((PeakAna&)obj).mTunnelSigma = mTunnelSigma;
+  ((PeakAna&)obj).mTunnelThreshold = mTunnelThreshold;
+  
+  ((PeakAna&)obj).mPainter = 0;//Dont' copy painter
+  
+  return;
```
plexoos: Unnecessary

```suggestion
```

StRoot/StFcsWaveformFitMaker/PeakAna.cxx
```diff
@@ -0,0 +1,1336 @@
...
+    ana->mFoundPeak = this->mFoundPeak;
+  }
+  return ana;
+}
+
+PeakAna::~PeakAna()
+{
+  if(mG_Data!=0 && mInternalSignal ){delete mG_Data;}
+  if( mFilterWeights!=0 ){ delete [] mFilterWeights; mFilterWeights=0; }
+  if( mPainter!=0 ){ delete mPainter; mPainter=0; }
```
plexoos: The checks for nullptr are unnecessary

```suggestion
  if (mInternalSignal) { delete mG_Data; mG_Data= nullptr; }
  delete [] mFilterWeights; mFilterWeights = nullptr;
  delete mPainter; mPainter = nullptr;
```

StRoot/StFcsWaveformFitMaker/PeakAna.cxx
```diff
@@ -0,0 +1,1336 @@
...
+  if( mComputedIndex<0 ){this->AnalyzeForPeak();}
+  if( mFoundPeak.mP_Peak<0 ){ xval=0; yval=0; }
+  else{ mG_Data->GetPoint(mFoundPeak.mP_Peak,xval,yval); }
+}
+
+void PeakAna::Print(Option_t* opt) const
+{
+  std::cout << "|P::Graph:"<<mG_Data;
+  std::cout << "|RangeX:["<<mXRangeMin<<","<<mXRangeMax<<"]";
+  std::cout << "|RangeY:["<<mYRangeMin<<","<<mYRangeMax<<"]";
```
plexoos, line 263: See https://star-bnl.github.io/star-coding-guidelines/formatguide.xml#Printing_Messages_in_STAR

StRoot/StFcsWaveformFitMaker/PeakAna.h
```diff
@@ -0,0 +1,351 @@
...
+  virtual void Copy(TObject& obj) const;//Clone graph as opposed to just copying the pointer
+  virtual TObject* Clone(const char* newname) const;
+  
+  virtual void AddPeakStats(TPaveText* pave, const char* opt="");
+  //Peak finding methods are independent of the class
+  static TGraph* ConvertHistToGraph(TH1* graph, UInt_t numavgs=1);//The second argument can be used to do a running average over 'numavgs' number of bins (1 means no average)
+  static TGraph* ConvertPeaksToGraph(const std::vector<PeakWindow> &Peaks);//The purpose of this function is to help with really noisy data where the "peaks" are really just noise. Essentially the idea is that each "peak" should actually be a point on a kind of "running average" in this way you capture the overall shape of the data without noise
+  void ConvertPeaksToGraph();//same as above but creates new graph that replaces old one
+  static PeakAna* ConvertPeaksToAna(const PeakAna &Ana);//same as above but returns new PeakAna with same settings without modifying old one
+  PeakAna* ConvertPeaksToAna();//same as above but returns new PeakAna with same settings without modifying current one
```
plexoos: Doxygen style comments please

StRoot/StFcsWaveformFitMaker/PeakAna.cxx
```diff
@@ -0,0 +1,1336 @@
...
+}
+
+PeakAna* PeakAna::ConvertPeaksToAna(const PeakAna &Ana)
+{
+  TGraph* graph = new TGraph();
+  for( Int_t ipeak=0; ipeak<Ana.NPeaks(); ++ipeak )
+  {
+    graph->SetPoint(ipeak, Ana.GetPeak(ipeak).mPeakX, Ana.GetPeak(ipeak).MidPoint() );
+  }
+  PeakAna* NewAna = new PeakAna(Ana,graph);
```
plexoos: This can cause a memory leak. Why not return the object? EDIT: I mean "return by value"

StRoot/StFcsWaveformFitMaker/StFcsWaveformFitMaker.h
```diff
@@ -147,38 +155,103 @@ class StFcsWaveformFitMaker : public StMaker {
     float LedFit( TGraphAsymmErrors* g, float* res, TF1*& f);//! mEnergySelect=31 (TF1 is for LED pulse fit)
 
     //pulse shape functions
-    double pulseShape(double* x, double* p);
-    double multiPulseShape(double* x, double* p);
+    //double pulseShape(double* x, double* p);
+    //double multiPulseShape(double* x, double* p);
```
plexoos:
```suggestion
```

StRoot/StFcsWaveformFitMaker/StFcsWaveformFitMaker.h
```diff
@@ -52,35 +52,41 @@
...
     ~StFcsWaveformFitMaker();
-    int InitRun(int runNumber);
-    int Make();
-    int Finish(); 
-    void Clear(Option_t* option = "");
+    virtual int Init();
+    virtual int InitRun(int runNumber);
+    virtual int Make();
+    virtual int Finish(); 
+    virtual void Clear(Option_t* option = "");
```
plexoos: What you should really use here is the `override` keyword rather than `virtual`

StRoot/StFcsWaveformFitMaker/StFcsWaveformFitMaker.h
```diff
@@ -52,35 +52,41 @@
...
 class StFcsHit;
 class StFcsDb;
+class StFcsDbPulse;
 class TGraphAsymmErrors;
 class TGraph;
 class TCanvas;
-class TH1F;
-class TH2F;
+//class TH2F;
+//class TFile;
```
plexoos:
```suggestion
```

StRoot/StFcsWaveformFitMaker/StFcsWaveformFitMaker.h
```diff
@@ -52,35 +52,41 @@
 #define STROOT_STFCSWAVEFORMFITMAKER_STFCSWAVEFORMFITMAKER_H_
 
 #include "StMaker.h"
-#include "StFcsPulseFit.h"
+//#include "StFcsPulseFit.h"
```
plexoos:
```suggestion
```

plexoos (commented):
StFcsWaveformFitMaker is already included in the `fcs` chain option. According to this page https://www.star.bnl.gov/devcgi/dbProdOptionRetrv.pl it was used once in 
`pp2022 StiCA BEmcChkStat epdhit -hitfilt` for "Preliminary  chain for run 2022 fastoffline production". From this I conclude that changing the algorithm should not have significant consequences

StRoot/StFcsWaveformFitMaker/StFcsWaveformFitMaker.cxx
plexoos: If it needs to be done only once, is there a reason not to initialize the histograms in the constructor? Also, if there is no good reason to allocate them with the `new` operator, you can avoid the many `delete` statements in the destructor.

genevb: NB: During StMaker::Init(), any newly instantiated histograms are added to the maker's "hist" subdirectory (in TDataset terms) which allows them to be found using the TDataset methods and enables inclusion in the BFC's hist.root output files.

//              Add the Maker histograms in the Maker histograms list
//              and remove it from the ROOT system directory
      gROOT->cd();
      TIter nextHist(gDirectory->GetList());
      Int_t ready = !objLast;
      while((objHist=nextHist())) {// loop over gDirectory
        if (!ready && objHist!=objLast)         continue;
        ready = 1999;
        if (objHist==objLast)                   continue;
        if (!objHist->InheritsFrom("TH1"))      continue;

//              Move the histogram from the ROOT list into the "maker's" list
        ((TH1*)objHist)->SetDirectory(0);
        maker->AddHist((TH1*)objHist);
      }

dkapukchyan: Moving them to the constructor means that these histograms will always be created, whereas now they are only created when a certain test level is set using *setTest()*. I thought this would be best since in the production these histograms don't need to be created as they are mostly used to quantify how well the algorithm is working. If I move them to the constructor then I would have to add an argument for the test level and file name where those histograms will be saved. I think doing that will only make this class more complicated so I left them in Init(). That way whoever calls this class can turn testing on or off before it hits the chain.

StRoot/StFcsWaveformFitMaker/PeakWindow.cxx
dkapukchyan: Sorry I don't understand the terms "delegating constructor" and "boilerplate code". Could you please elaborate?

StRoot/StFcsWaveformFitMaker/PeakAnaVirtualPainter.h
dkapukchyan: Using this example, it should just be *PEAKANAVIRTUALPAINTER_H*? I ask because I was following the include guard style in StFcsWaveformFitMaker.h. Should that be changed to just *STFCSWAVEFORMFITMAKER_H*?

StRoot/StFcsWaveformFitMaker/PeakAnaPainter.h
dkapukchyan: This follows the ROOT TH1 code design for TVirtualHistPainter and THistPainter (among others), and it is necessary for the following reason. If you have just a PeakAnaPainter object then PeakAna must have a pointer to PeakAnaPainter which you can forward declare. However, forward declaration of a PeakAnaPainter is not sufficient since you need to call methods of PeakAnaPainter in PeakAna. This means that you must include PeakAnaPainter and you can't just declare it. This creates another problem because you also need to forward declare PeakAna in PeakAnaPainter because PeakAnaPainter has a pointer to PeakAna. Now you have the same problem as before because forward declaring PeakAna is not enough since you call on PeakAna methods in PeakAnaPainter. This means that both PeakAna must include PeakAnaPainter, and PeakAnaPainter must include PeakAna which is not possible. The virtual painter solves this problem because I can include the PeakAnaVirtualPainter which doesn't require the existence of PeakAna. Sorry for the long winded and maybe confusing explanation but I did spend a lot of time figuring this out and realized that without a virtual interface, a separate analysis and painter class could not exist.

StRoot/StFcsWaveformFitMaker/PeakAna.cxx
dkapukchyan: I think they are necessary because, even though I tried to always keep a graph object alive, it is a sanity check in case someone decides to call *SetData(0)* in which case mG_Data will equal 0 and then you are deleting a nullptr which is bad. The same goes for the others since mFilterWeights=0 unless *SetFilter(2,...)* is called, and mPainter=0 unless *Paint()* is called. I don't remember why I don't check for a nullptr in *SetData()* so I'll add that and then I agree that `mG_Data!=0` is unnecessary. Also, since this is in the destructor I probably don't need to set them to the nullptr afterwords.

StRoot/StFcsDbMaker/StFcsDbPulse.cxx
dkapukchyan: Which header do I need to be able to use LOG_XXX?

StRoot/StFcsWaveformFitMaker/PeakAnaVirtualPainter.h
plexoos: I would personally use `#pragma once` :) but since the current style guideline says to use FILENAME_H so be it.

StRoot/StFcsWaveformFitMaker/PeakAna.cxx
veprbl: It's fine to do `x=nullptr; delete x;`
See e.g. https://en.cppreference.com/w/cpp/language/delete

> If expression evaluates to a null pointer value, no destructors are called, and the deallocation function may or may not be called (it's unspecified), but the default deallocation functions are guaranteed to do nothing when passed a null pointer.

plexoos: > Also, since this is in the destructor I probably don't need to set them to the nullptr afterwords.

Good point. Normally they shouldn't be accessible in the derived classes anyway.

StRoot/StFcsWaveformFitMaker/PeakAna.cxx
dkapukchyan: I am not sure whether you are referring to *graph* or *NewAna*. I don't think *graph* will give a memory leak since the next line calls `NewAna->ForceInternal()`. This means that the newly created graph object is "owned" by NewAna and it will delete that newly created graph when it is destroyed or *SetData()* is called. There is no way to change mInternalSignal to false after that so NewAna will never think that it doesn't own *graph*. If you are referring to NewAna then I agree it will if someone forgets to delete it. I don't want to return the graph object itself because if someone does `PeakAna* newana = new PeakAna(OldAna,ConvertPeaksToAna(OldAna))` then the way this constructor works it will think the graph object is external to it and it won't delete it and that will be a memory leak. Which means that the caller will have to remember to always call `newana->ForceInternal()` in this situation. Here I called it automatically to prevent such a situation so that's why I don't return the graph object.

StRoot/StFcsDbMaker/StFcsDbPulse.cxx
plexoos: St_base/StMessMgr.h

StRoot/StFcsWaveformFitMaker/StFcsWaveformFitMaker.h
dkapukchyan: I tried using override and compiling with cons and it complains that symbol override is not defined. Can you provide a working example?

StRoot/StFcsWaveformFitMaker/PeakAna.cxx
plexoos: Actually, I meant you could return the NewAna object by value rather than by pointer, not the graph object. I am a bit lost on whether PeakAna owns the graph object or not and under what conditions... Hopefully, this complex (for me) logic is correct and there are no memory leaks.

StRoot/StFcsWaveformFitMaker/PeakAnaPainter.h
plexoos: Forward declaration of PeakAna should work in this header file since you refer to it via a pointer only. You can do #include "PeakAna.h" in the cxx file. Similarly, in PeakAna.h you can forward declare PeakAnaPainter and include in the implementation file.

Early versions of ROOT is not the best place to pick up design patterns 😄

StRoot/StFcsWaveformFitMaker/PeakWindow.cxx
plexoos: You can use one constructor in the member initializer list of another constructor. It lets you not to repeat the same initialization code

https://en.cppreference.com/w/cpp/language/constructor

StRoot/StFcsWaveformFitMaker/PeakWindow.cxx
dkapukchyan: done

StRoot/StFcsDbMaker/StFcsDbMaker.h
dkapukchyan: done

StRoot/StFcsDbMaker/StFcsDbPulse.h
dkapukchyan: done

StRoot/StFcsWaveformFitMaker/PeakAna.cxx
dkapukchyan: done

StRoot/StFcsWaveformFitMaker/StFcsWaveformFitMaker.h
dkapukchyan: done

StRoot/StFcsWaveformFitMaker/PeakAna.cxx
dkapukchyan: I thought about returning by value but my understanding is that when you return by value it returns a copy of the object not the object created in the function itself. This is not something I want for a large object like PeakAna. Also, if you try to get around this by returning a reference instead then the returned reference is to the object created in the function which has now been invalidated and hence does not exist. This is why I return the pointer to the new object. Is my understanding incorrect and is there a safe way to do this without copying the object?

StRoot/StFcsWaveformFitMaker/PeakAna.cxx
dkapukchyan: @veprbl I didn't know that. Thanks! this is great to know. I'll change it.

StRoot/StFcsWaveformFitMaker/PeakAna.cxx
plexoos: It is safe to assume that your compiler implements the return value optimization. Moreover, copy elision is guaranteed when an object is returned directly (as of C++17)

StRoot/StFcsWaveformFitMaker/PeakAna.h
dkapukchyan: I am still learning the syntax for doxygen style comments. Is there a way to generate the html for a given file or files so that I can see how it is working?

StRoot/StFcsWaveformFitMaker/StFcsWaveformFitMaker.h
dkapukchyan: done

StRoot/StFcsWaveformFitMaker/StFcsWaveformFitMaker.h
dkapukchyan: done

StRoot/StFcsWaveformFitMaker/PeakAnaPainter.h
dkapukchyan: Thanks, that worked! I thought I had tried something similar to that but it didn't work that's why I opted to try to use the virtual methods I saw in the ROOT code.

starsdong:
Akio, Daniel, we should try to conclude on this PR. As codeowners, could you please review and approve this PR?
Dmitri, in the meantime, the PR seems to be out-of-sync with main branch. Could you update the branch?

dkapukchyan:
Xilin looked at invariant mass plot and finds no difference in the pi0 peak position and shape using the StFcsWaveformFitMaker in this commit and current StFcsWaveformFitMaker.

akioogawa (approved):
This is now well tested and ready to go.

StRoot/StFcsWaveformFitMaker/StFcsPulseAna.cxx
```diff
@@ -0,0 +1,813 @@
...
+
+  return ana;
+}
+
+StFcsPulseAna::~StFcsPulseAna()
+{
+  if(mH1_Baseline!=0)     { delete mH1_Baseline; }
+  if(mF1_SignalFit!=0)    { delete mF1_SignalFit; }
+  if(mF1_BaselineFit!=0)  { delete mF1_BaselineFit; }
+
```
plexoos:
```suggestion
  delete mH1_Baseline;
  delete mF1_SignalFit;
  delete mF1_BaselineFit;
```

StRoot/StFcsWaveformFitMaker/StFcsPulseAna.cxx
```diff
@@ -0,0 +1,813 @@
...
+    case 1 :
+      //std::cout << "    - SumDep0Mod::Case1|peak:"<<peak << "|radc:"<<radc<<"|sum:"<<sum << std::endl;
+      if(radc>sum) peak = 1 ;
+      sum += radc ;
+      last = radc ;
+      break ;
+    case 2 :
+      //if( peak>0 && radc<=last) peak ^= 1<<rightshift ;
+      //std::cout << "    - SumDep0Mod::Case2|peak:"<<peak << "|radc:"<<radc<<"|last:"<<last << std::endl;
+      //if( peak!=0 || radc<last) peak = -1 ;
```
plexoos: Can you please clean up the code and remove the commented code in all modified files? These distract focus from the important lines

StRoot/StFcsWaveformFitMaker/PeakAna.cxx
plexoos: Is it going to be addressed?

StRoot/StFcsWaveformFitMaker/PeakAna.cxx
plexoos: Is it going to be addressed?

StRoot/StFcsWaveformFitMaker/PeakAna.cxx
```diff
@@ -0,0 +1,1336 @@
...
+  NewAna->ForceInternal();//new graph so needs to be deleted
+  return NewAna;
+}
+
+PeakAna* PeakAna::MeanFilter( Int_t sizeavgs, bool copy )
+{
+  if( this->GetData()==0 ){return this;}
```
plexoos: You use `this->` a lot for class members. In most of our code it is usually omitted. I suggest you do the same.

StRoot/StFcsWaveformFitMaker/StFcsPulseAna.h
```diff
@@ -0,0 +1,123 @@
...
+  explicit StFcsPulseAna( TGraph* Sig, std::string name = "StFcsPulseAna"); //!< Construct with a known TGraph
+  StFcsPulseAna(const StFcsPulseAna& old, const char* post_name="_copy", TGraph* graph=0); //!< Copy Constructor can be called with a new graph and post-fix to the name
+  StFcsPulseAna& operator=(const StFcsPulseAna& rhs);   //!< Assignment operator
+
+  virtual ~StFcsPulseAna();  //!< Destructor
+
+  virtual void Copy(TObject& obj) const;              //!< Internal method use Clone instead.
+  virtual TObject* Clone(const char* newname) const;  //!< Clones internal graph as opposed to just copying the pointer
+
+  virtual Int_t AnalyzeForPeak();  //!< Overwritten from #PeakAna to process peak tunneling after finding all peaks
```
plexoos, line 47: There is quite a few new warnings related to hidden overloaded functions. E.g.
```
2022-10-13T14:55:16.2870933Z #10 1078.3 In file included from .sl79_gcc485/OBJ/StRoot/StFcsWaveformFitMaker/StFcsPulseAna.h:31:0,
2022-10-13T14:55:16.2871717Z #10 1078.3                  from .sl79_gcc485/OBJ/StRoot/StFcsWaveformFitMaker/StFcsPulseAna.cxx:1:
2022-10-13T14:55:16.2874812Z #10 1078.3 .sl79_gcc485/OBJ/StRoot/StFcsWaveformFitMaker/PeakAna.h:164:17: warning: 'virtual Int_t PeakAna::AnalyzeForPeak(Double_t, Double_t)' was hidden [-Woverloaded-virtual]
2022-10-13T14:55:16.2877440Z #10 1078.3    virtual Int_t AnalyzeForPeak(Double_t peak, Double_t width);//<! same as #AnalyzeForPeak() but also sets search parameters for peak
2022-10-13T14:55:16.2878444Z #10 1078.3                  ^
```

Please take care of them

StRoot/StFcsWaveformFitMaker/StFcsWaveformFitMaker.cxx
```diff
@@ -216,6 +379,68 @@ int StFcsWaveformFitMaker::Finish(){
...
+    
+    if( mTime!=0 ){ mTime->Write(); }
+    
+    if( mH1_NPeaksAkio!=0 ){ mH1_NPeaksAkio->Write(); }
+    if( mH1_NPeaksFilteredAkio!=0 ){ mH1_NPeaksFilteredAkio->Write(); }
+    if( mH1_PeakTiming!=0 ){mH1_PeakTiming->Write();}
+    
+    if( mH2_NPeakvsPeakXdiff!=0 )      {mH2_NPeakvsPeakXdiff->Write();}
+    if( mH2_NPeakvsPeakYratio!=0)      {mH2_NPeakvsPeakYratio->Write();}
+    if( mH1_VOverlap!=0         )      {mH1_VOverlap->Write();}
```
plexoos, line 423: Have you tried to make a single call to `rootFile->Write();` instead of doing this for individual objects?

StRoot/StFcsWaveformFitMaker/StFcsPulseAna.cxx
dkapukchyan: done

StRoot/StFcsWaveformFitMaker/StFcsPulseAna.cxx
dkapukchyan: I removed what I could. Please let me know if there is more that I missed. Long comments describing what the algorithms do and how they work I left in since I don't know where else to put them.

StRoot/StFcsWaveformFitMaker/PeakAna.cxx
dkapukchyan: Removed what I found. I only left the ones relating to virtual methods.

StRoot/StFcsWaveformFitMaker/StFcsPulseAna.h
dkapukchyan: should be fixed now

StRoot/StFcsWaveformFitMaker/StFcsWaveformFitMaker.cxx
dkapukchyan: I have tried it before for other classes and it doesn't always work. It probably has to do with the order in which you create TFile and histograms so I explicitly write them in now. If it is really a big problem I can try to see if rootFile->Write() will work in this instance.

StRoot/StFcsWaveformFitMaker/PeakAna.cxx
dkapukchyan: I thought in a Print function such as this using `cout` should not be a problem since  calling a Print function means that you want it to print to the standard out. Are we not allowed to have `std::cout` anywhere and have to use the STAR handlers only?

StRoot/StFcsWaveformFitMaker/PeakAna.cxx
dkapukchyan: done

StRoot/StFcsWaveformFitMaker/StFcsWaveformFitMaker.cxx
plexoos: Yes, please try again. ROOT has this notion of a current directory so, when you open a new file it switches to that current directory and it becomes a parent for all TObject-s created after that. The common trick with ROOT is to save the current directory and restore it later when you are done creating your histograms. Since you do it in a Maker this can interfere with other opened files, although I recall you have a switch which disables histogram filling by default. Since no tests were provided for all of this functionality we can only hope that it all works as intended.

StRoot/StFcsWaveformFitMaker/PeakAna.cxx
plexoos: I am not aware of any exclusion to the rule requiring the use of LOG_ stream objects. The advantage is that we can turn off or limit logging to a predefined level. I hope in production only errors and maybe warnings are logged to avoid unnecessary writes to, for example, LOG_DEBUG and LOG_INFO streams

StRoot/StFcsWaveformFitMaker/StFcsWaveformFitMaker.cxx
```diff
@@ -86,116 +87,277 @@ ClassImp(StFcsWaveformFitMaker)
...
+  if( mH2_HeightvsSigma!=0  )     {delete mH2_HeightvsSigma;}
+  if( mH2_HeightvsSigmaTrig!=0 )  {delete mH2_HeightvsSigmaTrig;}
+  if( mH1_ChiNdf!=0         )     {delete mH1_ChiNdf;}
+  if( mH2_HeightvsChiNdf!=0 )     {delete mH2_HeightvsChiNdf;}
+  if( mH2_MeanvsChiNdf!=0   )     {delete mH2_MeanvsChiNdf;}
+  if( mH2_SigmavsChiNdf!=0  )     {delete mH2_SigmavsChiNdf;}
+
+  if( mH1_PeakTimingGaus!=0 )     {delete mH1_PeakTimingGaus;}
+  if( mH1_PeakTimingPuls!=0 )     {delete mH1_PeakTimingPuls;}
+  if( mH2_PeakTimingCompare!=0 )  {delete mH2_PeakTimingCompare;}
```
plexoos: I thought we already discussed this: there is no need to check for the pointer to be non null before deleting the object. Isn't nicer to have less code than more code?

```suggestion
  delete mH1_PeakTimingGaus;
  delete mH1_PeakTimingPuls;
  delete mH2_PeakTimingCompare;
```

StRoot/StFcsWaveformFitMaker/StFcsWaveformFitMaker.cxx
dkapukchyan: I tried with just rootFile->Write() and it didn't work. I understand the issue but I don't know what else to do to avoid this. You are correct that if no test level is called then no output needs to be saved. The test levels are for single file tests and should never be called in production. They are intended to help understand and improve the algorithm rather than for generating generic QA histograms for StFcsWaveformFitMaker.

StRoot/StFcsWaveformFitMaker/PeakAna.cxx
dkapukchyan: I changed some cout statements so that it uses the STAR messenger system where appropriate. Let me know if there are other places where it needs to be changed.

StRoot/StFcsWaveformFitMaker/StFcsWaveformFitMaker.cxx
dkapukchyan: Done. Sorry I had forgotten about this file :)

jdbrice (approved):

PR #402, opened 2022-09-22
cons: Use mysql_config to set up external dependency for mysql

plexoos:
When we configure the build with mysql as external dependency we can fully rely on `mysql_config` found in PATH

This is similar to #357 where we also use `config` utilities to set up `fastjet` and `gsl` packages.

### DIFF
--- a/mgr/ConsDefs.pm
+++ b/mgr/ConsDefs.pm
@@ -868,41 +868,19 @@
 
     # search for the config    
     my ($MYSQLCONFIG,$mysqlconf);
-    # if ( defined($ENV{USE_LOCAL_MYSQL}) ){
 	($MYSQLCONFIG,$mysqlconf) =
 	    script::find_lib($MYSTAR . "/bin " .  $MYSTAR . "/bin/mysql ".
 			     $MYSQL . " ".
 			     $MYSQL . "/bin ".
 			     "/usr/$LLIB/mysql /usr/bin/mysql /usr/bin " . dirname(`which mysql_config`),
 			     "mysql_config");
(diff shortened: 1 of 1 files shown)


klendathu2k (approved):
Seems reasonable.  Only issue is that if we need to build against a specific mysql install... it needs to be located in the path before /usr/bin/mysql_config.  (Presumably managed by spack or ...)

plexoos:
In case someone wants to use a specific mysql install, the same principle would apply, i.e. they just update/prepend their PATH with the location of their mysql_config. It is not very different from setting an additional variable such as USE_LOCAL_MYSQL

P.S. It is a bit hard to imagine a volunteer who would like to experiment with a new version of mysql client.  I think at this moment in STAR history the focus is more on the environment preservation rather than a new R&D... But the possibility is not closed with this update.

plexoos:
Gene, do you have any objections to the changes?

genevb:
> Gene, do you have any objections to the changes?

I don't know enough about this to object to the implementation. I don't have any objections to the idea of enabling flexible handling of mysql dependency as long as people compiling root4star outside of SDCC/RCF don't end up having  complications and confusion.

genevb (commented):
No bjections

plexoos:
Thank you for your feedback. I asked because you are a designated maintainer of this code in mgr/ Of course it can be changed at any time at your will.
If there are any specific concerns with installation of star-sw on a supported system we'll deal with them when details become available.

I'll interpret this response as an approval for these specific changes and in the similar PRs: #402, #404, and #405

PR #404, opened 2022-09-22
cons: Use xml2-config to set up external dependency

plexoos:
When we configure the build with xml2 as external dependency we can rely on xml2-config found in PATH

This is similar to https://github.com/star-bnl/star-sw/pull/357 where we also use config utilities to set up fastjet and gsl packages and #402 for mysql


klendathu2k (approved):
Perfectly reasonable.

genevb (commented):
No objections.

PR #405, opened 2022-09-23
cons: Use pkg-config to set up external dependency on log4cxx

plexoos:
When we configure the build with log4cxx as external dependency we can rely on pkg-config found in PATH

This is similar to https://github.com/star-bnl/star-sw/pull/357 where we also use config utilities to set up fastjet and gsl packages, https://github.com/star-bnl/star-sw/pull/402 for mysql, and #404 for xml2


klendathu2k (approved):
Perfectly reasonable.

genevb (commented):
No objections.

PR #389, opened 2022-08-30
Fix GCC warning: forbidden conversion to char*

plexoos:
GCC 11
warning: ISO C++ forbids converting a string constant to 'char*' [-Wwrite-strings]

GCC 485
warning: deprecated conversion from string constant to 'char*' [-Wwrite-strings]

See #392 for motivation

PR #406, opened 2022-09-23
Assorted updates to StRoot/RTS, including "char *" issue

jml985:
This updates the StRoot/RTS directories to solve issues with the conversion between char * and const char *

### DIFF
--- a/StRoot/RTS/include/DAQ1000/rb.hh
+++ b/StRoot/RTS/include/DAQ1000/rb.hh
@@ -67,7 +67,7 @@ public:
 	int emu_fifo_length[128] ;
 	volatile rbFifo_t *sfifo, *rfifo ;	// same for fifos
 
-protected:
+//protected:
 
 	int board ;
 	int ch ;	// also port!
--- a/StRoot/RTS/include/fcfClass.hh
+++ b/StRoot/RTS/include/fcfClass.hh
@@ -201,7 +201,7 @@ public:
 	int compare(u_int *p1[3], u_int *p2[3]) ;
 
 	void decode(u_int *data, fcfHit *h, u_int *sim = 0) ;	// utility: from FCF to local
-	void print_hit(char *, fcfHit *hit) ;	// utility
+	void print_hit(const char *, fcfHit *hit) ;	// utility
 
 	u_int do_swap ;		// 0=no, 1=yes => set by "burn"
 	u_int do_merge ;	// merge broken rows
--- a/StRoot/RTS/include/platform.h
+++ b/StRoot/RTS/include/platform.h
@@ -467,7 +467,7 @@ extern MSG_Q_ID *pMsgQArray;
 inline void pSetupMsgQArray(MSG_Q_ID *qarray) {
   pMsgQArray = qarray;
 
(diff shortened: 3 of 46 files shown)


plexoos:
Is this ready? @akioogawa

akioogawa (approved):

plexoos:
@tonko-lj ?

jml985:
Yes I think it is ready as far as we are concerned, though I don't have the metrics you were using.  If there are still issues let me know.

plexoos:
The metric is just the warnings you find when you build

jml985:
Right, but we don't build the way you do.   What we did was set the compiler to treat warnings as errors, then compile our code.   We had to exclude many of these, such as "unused function" warnings, which make no sense in code that is effectively libraries...   The CI seemed to go through correctly.   I thought you showed some plots of the number of various warnings and those are the metrics I meant.

plexoos:
> Right, but we don't build the way you do. What we did was set the compiler to treat warnings as errors, then compile our code.

Treating warnings as errors is exactly what we would like to do after we clean up the existing code. We haven't switch to that mode yet.

> We had to exclude many of these, such as "unused function" warnings, which make no sense in code that is effectively libraries... 

I saw such warnings in your code. They make sense because the functions are defined as static and that prevents them from being accessed (by name) in any compilation unit other than where they are defined. Here is an example with the static `t_delta` function that is not used in the same compilation unit: https://github.com/star-bnl/star-sw/blob/main/StRoot/RTS/src/DAQ_FGT/fstPed.cxx#L31

Unless there is a really good reason we should not ignore any from -Wall and perhaps -Wextra

> The CI seemed to go through correctly. I thought you showed some plots of the number of various warnings and those are the metrics I meant.

The plots just show the count of warnings to keep track of the progress. I did not split by the module but I can produce one specifically for StRoot/RTS/
To search the logs for warnings you can go to the "Checks" tab on the PR page, pick a job (e.g. "build (root5, gcc485)") and download the log by selecting from the menu:
  
Screen Shot 2022-09-28 at 3 01 41 PM

plexoos:
Here are the counts specifically for this PR #406 (as of 0951ee3f36c4d86185d4d9336c6e8387dfeb8af1) and StRoot/RTS module compared to the same in SL22c

fix_warns

```
 msg_id                                                                    msg
      0        deprecated conversion from string constant to [-Wwrite-strings]
      1                  variable set but not used [-Wunused-but-set-variable]
      2                                    unused variable [-Wunused-variable]
      3                               defined but not used [-Wunused-function]
      4     may be used uninitialized in this function [-Wmaybe-uninitialized]
      5                               statement has no effect [-Wunused-value]
      6                            value computed is not used [-Wunused-value]
      7 iteration invokes undefined behavior [-Waggressive-loop-optimizations]
      8     format expects argument of type, but argument has type [-Wformat=]
      9                               defined but not used [-Wunused-variable]
     10           extra tokens at end of #endif directive [enabled by default]
     11                            label defined but not used [-Wunused-label]
     12                                         redefined [enabled by default]
```

Looks like an improvement but still more to be fixed.
I'll just bring everyone's attention again to the fact that a good deal of these is already addressed in #389 and #395

PR #408, opened 2022-10-07
Stargenerator Vertexing Options

klendathu2k:
Added several options for sampling "flat" vertex distributions:
1) Line along z-axis
2) Cylinder along z-axis
3) Elliptical cylinder along z-axis
4) Rectangle centered on z-axis

Default is the standard gaussian distribution.  Otherwise user selects from
a menu of options.  The gaussian width parameters are used to define the
parameters of each of the four "flat" distributions.  Their meaning are specified
in the header file.


StRoot/StarGenerator/BASE/StarPrimaryMaker.cxx
```diff
@@ -69,6 +79,72 @@ StarPrimaryMaker::StarPrimaryMaker()  :
 
   SetAttr("FilterKeepHeader", int(1) );
 
+  // Default vertex function
+  mVertexFunction = [this]() -> TLorentzVector {
+    Double_t x=0,y=0,z=0,t=0;
+    TVector2 xy = StarRandom::Instance().gauss2d( this->mSx, this->mSy, this->mRho );
+    x = mVx + xy.X();
+    y = mVy + xy.Y();
+    z = mVz + StarRandom::Instance().gauss( this->mSz );
```
plexoos: I think the use of `this->` for `mVx`, `mVy`, and `mVz` is critical here

StRoot/StarGenerator/BASE/StarPrimaryMaker.cxx
```diff
@@ -69,6 +79,72 @@ StarPrimaryMaker::StarPrimaryMaker()  :
...
+    Double_t dist = TMath::Sqrt(x*x+y*y+z*z);
+    t = dist / TMath::Ccgs();
+    return TLorentzVector(x,y,z,t);
+  };
+  vertexFunctionMap["gaussXYZ"] = mVertexFunction;
+  vertexFunctionMap["flatZ"]  = [this]() -> TLorentzVector {
```
plexoos: Why do you use lambdas instead of regular functions?
- The functions are not very small, you declare new variables and even have loops inside lambdas
- It is a bit hard to see the signature of the function if it is not declared as `Out myfunc(Inp1, Inp2, ...);`
- You also capture by reference the entire object, which may lead to hard-to-find errors if there is a mistake
- If one day you decide to write a unit test for your random distributions, it is much easier to do for a regular function

StRoot/StarGenerator/BASE/StarPrimaryMaker.h
```diff
@@ -163,15 +167,24 @@ class StarPrimaryMaker : public StMaker
...
+  /// Set vertexing function from named options.  The vertex center parameters (Vx,Vy,Vz) set using SetVertex define the center of
+  /// the distribution.  The vertex widths Sx,Sy,Sz (plus Rho) set using SetSigma define the gaussain widths for the gaussian vertexing
+  /// option.  For the other options, they are used as follows.
+  /// @param name specifies the vertexing option.
+  /// name = "gausXYZ" is the default.
+  /// name = "flatZ"   throws points uniform on a line from Vz - Sz to Vz + Sz
+  /// name = "flatXYZ" thows within a rectangular box Vx+/-Sx, Vy+/-Sy, Vz+/-Sz
+  /// name = "flatRZ"  throws within a cylinder of radius Sx centered on Vx,Vy, from z=Vz-Sz to Vz+Sz
+  /// name = "flatABZ" throws within an eliptical cyilnder of major axis A minor axis B, rotated by Rho
+  void SetVertexing( const char* name );
```
plexoos: Are you sure you need this setter? Is there a point in switching to a different vertex distribution after the maker is initialized? This option can be added as an argument to the constructor with a default value.

StRoot/StarGenerator/BASE/StarPrimaryMaker.h
```diff
@@ -163,15 +167,24 @@ class StarPrimaryMaker : public StMaker
...
 
+  /// Set vertexing function from named options.  The vertex center parameters (Vx,Vy,Vz) set using SetVertex define the center of
+  /// the distribution.  The vertex widths Sx,Sy,Sz (plus Rho) set using SetSigma define the gaussain widths for the gaussian vertexing
+  /// option.  For the other options, they are used as follows.
+  /// @param name specifies the vertexing option.
+  /// name = "gausXYZ" is the default.
+  /// name = "flatZ"   throws points uniform on a line from Vz - Sz to Vz + Sz
+  /// name = "flatXYZ" thows within a rectangular box Vx+/-Sx, Vy+/-Sy, Vz+/-Sz
+  /// name = "flatRZ"  throws within a cylinder of radius Sx centered on Vx,Vy, from z=Vz-Sz to Vz+Sz
+  /// name = "flatABZ" throws within an eliptical cyilnder of major axis A minor axis B, rotated by Rho
```
plexoos: Please consider using an enum instead of a collection of strings.

StRoot/StarGenerator/BASE/StarPrimaryMaker.cxx
```diff
@@ -29,12 +29,22 @@ ClassImp(StarPrimaryMaker);
...
 #include "tables/St_vertexSeed_Table.h"
+#include <map>
+#include <string>
 
 using namespace std;
 
 // 1 mm / speed of light
 const double mmOverC = 1.0E-3 / TMath::C();
 
+std::map< std::string, std::function< TLorentzVector()> > vertexFunctionMap;
```
plexoos: This map is defined outside of the maker but is initialized in the maker. This is an undefined behavior when two instances are created (as in MT :) ) even though they do the same thing.

klendathu2k: To be fixed in the next commit / push.

StRoot/StarGenerator/BASE/StarPrimaryMaker.cxx
klendathu2k: Mainly because it was less boilerplate than binding to a member function.  However, this is changed in the next set of commits / push.

StRoot/StarGenerator/BASE/StarPrimaryMaker.h
klendathu2k: At some point I would like to provide a method which adds distributions to the menu.   Enum would require a recompile of the maker.

StRoot/StarGenerator/BASE/StarPrimaryMaker.h
klendathu2k: The StMaker framework design discourages setting maker parameters as constructor arguments.... but you are correct that a setter is not quite right here.  I should set an string attribute.  To be done...

StRoot/StarGenerator/BASE/StarPrimaryMaker.h
```diff
@@ -222,6 +225,17 @@ class StarPrimaryMaker : public StMaker
   StarFilterMaker *mFilter;
   TEventList      *mAccepted; //*< event list containing accepted events
 
+#ifndef __CINT__
+  std::function< TLorentzVector() > mVertexFunction;
+  std::map< std::string, std::function< TLorentzVector() > > vertexFunctionMap;
```
plexoos: Per coding guidelines: vertexFunctionMap -> mVertexFunctionMap

StRoot/StarGenerator/BASE/StarPrimaryMaker.cxx
```diff
@@ -710,3 +722,74 @@ void StarPrimaryMaker::RotateBeamline( Double_t &px, Double_t &py, Double_t &pz,
   pz = moment[2];
 
 }
+
+
+//_________________________________________________________________________________________
+TLorentzVector StarPrimaryMaker::vertexGaussXYZ() {
+    double x=0,y=0,z=0,t=0;
+    TVector2 xy = StarRandom::Instance().gauss2d( this->mSx, this->mSy, this->mRho );
```
plexoos: For consistency 'this' may be dropped as in the other cases

StRoot/StarGenerator/BASE/StarPrimaryMaker.cxx
```diff
@@ -69,6 +75,22 @@ StarPrimaryMaker::StarPrimaryMaker()  :
...
+  vertexFunctionMap["flatXYZ"]  = std::bind( &StarPrimaryMaker::vertexFlatXYZ,  this );
+
+  SetAttr( "vertexDistribution", "gaussXYZ" );
+  /// name = "gaussXYZ" is the default.
+  /// name = "flatZ"   throws points uniform on a line from Vz - Sz to Vz + Sz
+  /// name = "flatXYZ" thows within a rectangular box Vx+/-Sx, Vy+/-Sy, Vz+/-Sz
+  /// name = "flatRZ"  throws within a cylinder of radius Sx centered on Vx,Vy, from z=Vz-Sz to Vz+Sz
+  /// name = "flatABZ" throws within an eliptical cyilnder of major axis A minor axis B, rotated by Rho
+
+  mVertexFunction = vertexFunctionMap[ SAttr("vertexDistribution") ];
```
plexoos: You may want to verify the user input. If non-existent key is provided it'll probably yield (0,0,0) for vertex position (maybe OK) or crash? (not OK)

plexoos (approved):
LGTM. Only minor comments from my side for you consideration. The class seems to support the slope in XZ and YZ but it is not implemented in the random distributions. An easy fix to add or may be in the next iteration...

klendathu2k:
> 

FYI... the beamline constraint (or user specified slope) is applied after the vertex is sampled (via the RotateBeamline method).

StRoot/StarGenerator/BASE/StarPrimaryMaker.cxx
klendathu2k: Done.  Default will be gaussian.  ... with a warning issued.

plexoos:
Jason, don't worry about the failed job. It is probably something I introduced by updating the star-test-data container with new test files. Investigating...

I think there is no point in waiting for other approvals so, let's merge this soon.

klendathu2k:
Agreed it is ready to move (do_events.C shouldn't have any dependence on 
StarGenerator... that
macro is at least 10 years older than StarGen).  Thanks.

Jason

PR #412, opened 2022-10-17
SpinPool Fcs pi0 finder fixing a crash

akioogawa:
Fix crash on StSpinPool/StFcsPi0FinderForEcal when Mudst has no primary vertex.

### DIFF
--- a/StRoot/StSpinPool/StFcsPi0FinderForEcal/StFcsPi0FinderForEcal.cxx
+++ b/StRoot/StSpinPool/StFcsPi0FinderForEcal/StFcsPi0FinderForEcal.cxx
@@ -247,8 +247,10 @@ Int_t StFcsPi0FinderForEcal::Make() {
       if(tpcvtx) {
 	zTPC=tpcvtx->position().z();
       }else{
-	StMuPrimaryVertex* mutpcvtx=StMuDst::primaryVertex();
-	if(mutpcvtx) zTPC=mutpcvtx->position().z();
+	if (StMuDst::numberOfPrimaryVertices() > 0){
+	  StMuPrimaryVertex* mutpcvtx=StMuDst::primaryVertex();
+	  if(mutpcvtx) zTPC=mutpcvtx->position().z();
+	}
       }
 
       //BBC ZVERTEX


jdbrice (approved):

StRoot/StSpinPool/StFcsPi0FinderForEcal/StFcsPi0FinderForEcal.cxx
```diff
@@ -247,8 +247,10 @@ Int_t StFcsPi0FinderForEcal::Make() {
       if(tpcvtx) {
 	zTPC=tpcvtx->position().z();
       }else{
-	StMuPrimaryVertex* mutpcvtx=StMuDst::primaryVertex();
-	if(mutpcvtx) zTPC=mutpcvtx->position().z();
+	if (StMuDst::numberOfPrimaryVertices() > 0){
+	  StMuPrimaryVertex* mutpcvtx=StMuDst::primaryVertex();
+	  if(mutpcvtx) zTPC=mutpcvtx->position().z();
+	}
```
plexoos, line 253: Does this new check imply that `StMuDst::arrays[muPrimaryVertex]` is not a valid pointer to a TClonesArray? It might be better to modify `StMuDst::primaryVertex()` directly...

genevb: The finding which I reported to Akio that prompted this PR was that the TClonesArray is valid, but empty. Howeer, `StMuDst::primaryVertex()` invokes an unchecked array lookup:

  /// return pointer to current primary vertex
  static StMuPrimaryVertex* primaryVertex() { return (StMuPrimaryVertex*)arrays[muPrimaryVertex]->UncheckedAt(mCurrVertexId); }
  
There is no value of  `mCurrVertexId` that is valid to use on an empty array, so the current implementation of `StMuDst::primaryVertex()` is only safely used after a check that the array has some size.

plexoos: I find it a bit strange that we don't see this issue with the empty TClonesArray `StMuDst::arrays[muPrimaryVertex]` when we select the default vertex in StPicoDstMaker:

https://github.com/star-bnl/star-sw/blob/6dd0b8e4e0317b2b82c84a9bb7bd744e414ddf85/StRoot/StPicoDstMaker/StPicoDstMaker.cxx#L2528-L2533

There seems to be a similar check for at least one vertex but it is enabled only when `__TFG__VERSION__` is defined.

https://github.com/star-bnl/star-sw/blob/6dd0b8e4e0317b2b82c84a9bb7bd744e414ddf85/StRoot/StPicoDstMaker/StPicoDstMaker.cxx#L821-L826

Just curious how the code in StSpinPool/StFcsPi0FinderForEcal is different from what we have in StPicoDstMaker?

genevb: From what I can tell, the difference is that `StFcsPi0FinderForEcal` tries to use the returned pointer to access the object, while StPicoDstMaker does not use this pointer for anything other than a check on whether it is non-zero, doing nothing critical with that information other than a print statement:.
https://github.com/star-bnl/star-sw/blob/6dd0b8e4e0317b2b82c84a9bb7bd744e414ddf85/StRoot/StPicoDstMaker/StPicoDstMaker.cxx#L834

genevb: Ah, it appears to also decide whether to build the pico event based on the returned pointer evaluating to true or false. The picoDst code doesn't appear robust for events without a primary vertex either to my eyes.

plexoos: By looking at the default TObjArray constructor I see that it reserves zero'ed space for 16 elements:

https://github.com/root-project/root/blob/250a360349e60dc2dc187ecaa63fc9f5f81a9d74/core/cont/src/TObjArray.cxx#L62-L75

So, I believe even if there are no vertices a call to `arrays[muPrimaryVertex]->UncheckedAt(0);` would still return a nullptr

The difference between the PicoDst and Akio's code is likely in this call `mMuDst->setVertexIndex(0);` before `mMuDst->primaryVertex();` If one does not set the current vertex id to a reasonable value the default value is used which is:

https://github.com/star-bnl/star-sw/blob/6dd0b8e4e0317b2b82c84a9bb7bd744e414ddf85/StRoot/StMuDSTMaker/COMMON/StMuDst.cxx#L89

The crash in Akio's code can be probably fixed as:

```diff
-	  StMuPrimaryVertex* mutpcvtx=StMuDst::primaryVertex();
+	  StMuPrimaryVertex* mutpcvtx=StMuDst::primaryVertex(0);
```

It may be better to set the default vertex id in StMuDst to 0

plexoos: Digging a bit deeper it appears the default vertex id was set to 0 for 11 years until about six years ago it was changed to -1:

https://github.com/star-bnl/star-sw/commit/d13bebd38a80771e165d74abe90f74ee41ae279e

apparently, the value of -1 did not work as it was causing crashes and it was changed to -2:

https://github.com/star-bnl/star-sw/commit/1e01d945ea05993c7b313410239d22aea26ca0ac

genevb: > So, I believe even if there are no vertices a call to `arrays[muPrimaryVertex]->UncheckedAt(0);` would still return a nullptr

This may not be correct if the array has not been re-populated with zeros after having been populated with non-zeros. The crash I investigated was on an event with 0 primary vertices following an event with 22 primary vertices.

plexoos: Gene, do you have the command to reproduce?

genevb: bfc.C(50,"DbV20221012,pp2022,StiCA,BEmcChkStat,ftt,-picowrite,-hitfilt,-evout","/star/data03/daq/2022/040/23040001/st_fwd_23040001_raw_0000003.daq")

plexoos: > if the array has not been re-populated with zeros after having been populated with non-zeros

Am I backtracing this incorrectly? We call StMuDstMaker::Clear() before/after each event which calls [TClonesArray::Clear](https://github.com/root-project/root/blob/v5-34-38/core/cont/src/TClonesArray.cxx#L393-L428) which calls [TObjArray::Clear](https://github.com/root-project/root/blob/v5-34-38/core/cont/src/TObjArray.cxx#L298-L307) which calls [TObjArray::Init](https://github.com/root-project/root/blob/250a360349e60dc2dc187ecaa63fc9f5f81a9d74/core/cont/src/TObjArray.cxx#L567-L584) which calls `memset(fCont, 0, fSize*sizeof(TObject*));`. To me it looks as all pointers are zeroed.

genevb: Actually, the array may even get deleted and rebuilt each event?

https://github.com/star-bnl/star-sw/blob/main/StRoot/StMuDSTMaker/COMMON/StMuDstMaker.cxx#L256

Just thinking hypothetically (I didn't check this), I don't see anywhere that `mCurrVertexId` gets reset between events. What if a new array is instantiated with 16 zeroed elements, but `mCurrVertexId` is greater than 16 from the previous event? Maybe that's the issue in this particular case?

genevb: Nope....I put in a print statement and found that `mCurrVertexId` is -2 all the way through the crash. Since -2 is not a valid index and goes unchecked, its understandable that it doesn't work to use it as an index, but still not clear why it ever does work.

plexoos: > but still not clear why it ever does work.

Accessing an array element by index outside of the allocated space is a UB (undefined behavior) in C++. Anything can happen, even correct results can be produced (sometimes).

In the Akio's code a request for primary vertex is ill formed because the vertex index is not specified. The way muDst is coded now, it is the responsibility of the caller to pick a valid vertex. I've checked with the test command posted above that setting it to 0 passes the test. It makes more sense to set the default vertex id in muDst back to 0 rather than -2 and call `collectVertexTracks()` in the setter unconditionally.

jdbrice: So it sounds like I should implement this change instead of (or in addition to) the proposed change here.

plexoos: Hold on... I think a somewhat better fix for this particular problem is to modify StFcsPi0FinderForEcal  as:

```diff
-StMuPrimaryVertex* mutpcvtx=StMuDst::primaryVertex();
+StMuPrimaryVertex* mutpcvtx=StMuDst::primaryVertex(0);
```

or

```diff
+StMuDst::setVertexIndex(0);
 StMuPrimaryVertex* mutpcvtx=StMuDst::primaryVertex();
```

it is "better" because it is consistent with how it is done in PicoDst where we don't check for the number of vertices in the array.

I am about to add the test using the data file as provided by Gene. Using it I want to check if we see the issue and the proposed solutions fix it.

As for the default vertex id = -2 in MuDst... yeah, but to me it looks like a secondary issue now w.r.t. other priorities 😄

zlchang (approved):

PR #416, opened 2022-10-20
Fix copy of eTOF good channel flags to PicoEvent

PhilippWeidenkaff:
Fixes bug introduced in #160


StRoot/StPicoEvent/StPicoEvent.cxx
```diff
@@ -360,7 +360,7 @@ void StPicoEvent::setETofGoodEventFlag( std::vector<bool> flagVec ) {
 	if( flagVec.size() != 108 ){
 	    LOG_INFO << "StPicoEvent::setETofGoodEventFlag() - eTof flag vector wrong size " << flagVec.size() <<" / 108"<< endm;
 	}else{
-		for( auto iterCounter : flagVec ){
+		for( int iterCounter = 0; iterCounter < flagVec.size(); iterCounter++ ){
```
plexoos: This is clearly a bug. Is the goal to copy the content of the `vector<bool> flagVec` into a C-array `bool mETofGoodEventFlag[108]`? Just use the function from the standard library:

```c++
std::copy(flagVec.begin(), flagVec.end(), mETofGoodEventFlag);
```

or

```c++
std::copy_n(flagVec.begin(), 108, mETofGoodEventFlag);
```

PhilippWeidenkaff: Ok, i have no problem to change that, but could you explain me why this is clearly a bug? Looping over the vector and copying entry by entry into the array seemed to be the most fool-proof option to me and in my testing seemed to work correctly.

plexoos:
It was a bug before your fix. At most the first two elements were copied due to implicit conversion of bool values to int. Thank you for finding and fixing it

StRoot/StPicoEvent/StPicoEvent.cxx
plexoos: It was a bug before your fix. At most the first two elements were copied due to implicit conversion of bool values to int. Thank you for finding and fixing it

PR #421, opened 2022-10-25
ci: Add test for FST chain with st_fwd stream data

plexoos:
See #412 for motivation


plexoos:
This test confirms the problem reported in #412. It also confirms that the check for at least one vertex in MuDst event (already merged into the main branch) addresses the issue. My suggestion to rather specify the vertex index when accessing it was not accepted in #412, the PR branch was merged before I had a chance to run the test and implement it. Leaving it up to the maintainers from this point.

PR #422, opened 2022-10-26
Encode Geometry Tag w/in DB macros

klendathu2k:
The Geometry.{tag}.C "database" macros establish the connection between the geometry tag (defined by the BFC chain options and/or timestamp) and the code which instantiates the geometry model.  The macro is loaded / executed by St_db_Maker whenever GetDataBase("VmcGeometry") is called.

This PR encodes the geometry tag inside of each database macro, rather than using the ROOT interpreter's introspection capabilities to read the tag from the filename of the currently executing macro. This works well under ROOT 5 / CINT.  But I think the way that we load macros in St_db_Maker confuses ROOT 6 / CINT.

The simplest solution is to encode the geometry tag within each database macro.

### DIFF
--- a/StarDb/AgMLGeometry/Geometry.dev13.C
+++ b/StarDb/AgMLGeometry/Geometry.dev13.C
@@ -1,32 +1,3 @@
+class TDataSet;
 #include "CreateGeometry.h"
-TDataSet *CreateTable() 
-{
-  // Name of the macro including the full path
-  TString myname = gInterpreter->GetCurrentMacroName();
-  
-  // Get an array of TString separated by "/"
-  TObjArray *array = myname.Tokenize("/");
-
-  // Get the name of the macro itself
-  TString name = ((TObjString *)array->Last())->GetString();
-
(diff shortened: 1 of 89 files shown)


plexoos:
diff stat: +178 −2,759 ❤️ Perfect refactoring assuming the functionality is not affected.

I think the next step is just to converge to a single macro accepting one parameter "tag"

klendathu2k:
> diff stat: +178 −2,759 heart Perfect refactoring assuming the functionality is not affected.
> 
> I think the next step is just to converge to a single macro accepting one parameter "tag"

Each of these files is considered to be a "database table".  In this case it is a trivial table whose title just corresponds to the geometry tag.  But it is being exercised through the general DB macro API, which expects there to be one table specified per file.

plexoos (approved):

genevb (approved):
Reduced redundancy, fewer lines of code to "get right" for each new geometry.... LGTM.

PR #427, opened 2022-10-28
Avoid pointer cast to time_t from incompatible type

### LINKED ISSUE #426: Invalid cast between incompatible types for std::gmtime argument
**Describe the bug**

The cast between unrelated types (as the one shown below) is not safe when the sizes of `unsigned` and `time_t` are different:

```c++
unsigned int ts_cast = ts;
struct tm *tm_cast = gmtime((time_t*) &ts_cast)
```

On our target system the sizes of the two types happened to be the same when using the -m32 compiler option but differ for -m64

This case was found in `StRoot/StDaqLib/SSD/SSD_Reader.cxx` due to the failed test 24. [The log from the job](https://github.com/star-bnl/star-sw/actions/runs/3329826613/jobs/5509640124). More specifically,
```
bfc.C(20, "P2005,tofDat,logger,MakeEvent,ITTF,ssddat,spt,SsdIt,SvtIt,pmdRaw,OShortR,OSpaceZ2", "/star/rcf/test/daq/2005/048/st_physics_6048025_raw_1020002.daq")
```

**To Reproduce**

Here is a small program to demonstrate the problem:

```c++
#include <ctime>
#include <iostream>
#include <vector>

using namespace std;

int main ()
{
  // $ date -d  @1108647599
  // Thu Feb 17 08:39:59 EST 2005

  cout << "sizeof(unsigned): " << sizeof(unsigned) << '\n';
  cout << "sizeof(time_t):   " << sizeof(time_t) << '\n';

  for (int ts : vector<int>{0, 5*24*3600+33, 1108647599})
  {
    time_t ts_nocast = ts;
    struct tm *tm_nocast = gmtime(&ts_nocast);
  
    unsigned int ts_cast = ts;
    struct tm *tm_cast = gmtime((time_t*) &ts_cast);

    cout << "tm_cast:   " << asctime(tm_cast);
    cout << "tm_nocast: " << asctime(tm_nocast);
  }

  return 0;
```

The erroneous output can be seen when comparing the -m32 and -m64 options respectively:
https://godbolt.org/z/xs9oezxro
https://godbolt.org/z/h3M5Pvq7f

**Expected behavior**

gmtime should return a valid date for the given Unix time when compiled with either -m32 or -m64 options

PS Many thanks to @veprbl for private discussion

plexoos:
Resolves #426

### DIFF
--- a/StRoot/StDaqLib/SC/SC_Reader.cxx
+++ b/StRoot/StDaqLib/SC/SC_Reader.cxx
@@ -97,8 +97,8 @@ SC_Reader::SC_Reader(EventReader *er) {
   //Keep BBCBkg scalers flipped as theyStRoot/StDaqLib/SC/SC_Reader.cxx were historically before 2009
   //Note that new DAQ reader leads to UTime = 0, or tm_year=70 (1970)
   //but new DAQ reader only gets used for 2009+ anyhow
-  unsigned int UTime = er->getEventInfo().UnixTime;
-  struct tm *time=gmtime((time_t*) &UTime);
+  std::time_t utime = er->getEventInfo().UnixTime;
+  std::tm *time = gmtime(&utime);
   flipBBCBkg = (time->tm_year > 95 && time->tm_year < 109 ? 1 : 0) ;
 
   //  LDate = (((1900+time->tm_year)*100 + 1 + time->tm_mon)*100 + time->tm_mday)*100;
--- a/StRoot/StDaqLib/SC/SC_Reader.cxx
+++ b/StRoot/StDaqLib/SC/SC_Reader.cxx
@@ -9,6 +9,7 @@
 
 #include <assert.h>
 #include <math.h>
+#include <ctime>
 
 using namespace OLDEVP;
 
--- a/StRoot/StDaqLib/SSD/SSD_Reader.cxx
+++ b/StRoot/StDaqLib/SSD/SSD_Reader.cxx
@@ -118,8 +118,8 @@ SSD_Reader::SSD_Reader(EventReader *er) {
   datap=er->getDATAP(); 
   if (datap) {
 
-  unsigned int UTime = er->getEventInfo().UnixTime;
-  struct tm *time=gmtime((time_t*) &UTime);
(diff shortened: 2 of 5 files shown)


genevb (commented):
If the proposed solution works for builds going forward, I have no objection to it.

veprbl (approved):

veprbl:
Minor issue:
```
#10 3672.1 .sl79_gcc485/OBJ/StRoot/StarRoot/TUnixTime.cxx: In member function 'TString TUnixTime::GetGString()':
[27021](https://github.com/star-bnl/star-sw/actions/runs/3347375021/jobs/5545286240#step:3:27024)
#10 3672.1 .sl79_gcc485/OBJ/StRoot/StarRoot/TUnixTime.cxx:135:3: error: 'time_t' is not a member of 'std'
[27022](https://github.com/star-bnl/star-sw/actions/runs/3347375021/jobs/5545286240#step:3:27025)
#10 3672.1    std::time_t utime = fUTime;
[27023](https://github.com/star-bnl/star-sw/actions/runs/3347375021/jobs/5545286240#step:3:27026)
#10 3672.1    ^
```

zlchang (approved):

starsdong (approved):
Looks good to me.

plexoos:
> If the proposed solution works for builds going forward, I have no objection to it.

Gene, do you know why we don't see this kind of problems with the 64-bit tests on the farm?

klendathu2k (approved):

plexoos:
> > If the proposed solution works for builds going forward, I have no objection to it.
> 
> Gene, do you know why we don't see this kind of problems with the 64-bit tests on the farm?

The answer to my own question is that only a handful of the nightly tests on the farm use 64-bit libraries. By looking carefully at this summary page https://www.star.bnl.gov/devcgi/weekDEVjobStatus.pl one can notice a "_64bit" suffix added to the values in the **Path** column. Assuming these are the only tests for 64-bit builds, the failed test in CI due to the error reported in #426 is never executed on the farm.

genevb:
> the failed test in CI due to the error reported in #426 is never executed on the farm.

Thanks for doing the investigative work, @plexoos

PR #436, opened 2022-11-10
Adding clang format.

DanielWielanek:
Adding clang format to code.

### DIFF
--- a/StRoot/StGmtUtil/geometry/StGmtGeom.h
+++ b/StRoot/StGmtUtil/geometry/StGmtGeom.h
@@ -10,6 +10,7 @@
 #define _ST_GMT_GEOM_H_
 
 #include <TVector3.h>
+#include <TString.h>
 #include <string>
 #include <sstream>
 #include <cstdlib>


.clang-format
```diff
@@ -0,0 +1,11 @@
+---
+BasedOnStyle: Google
+ColumnLimit: '100'
+IndentPPDirectives: None
+IndentWidth: '4'
+MaxEmptyLinesToKeep: '1'
+SpacesInParentheses: 'false'
+TabWidth: '4'
```
veprbl:
```suggestion
IndentWidth: '3'
MaxEmptyLinesToKeep: '1'
SpacesInParentheses: 'false'
TabWidth: '3'
```
is the standard in the ROOT world: https://github.com/root-project/root/blob/master/.clang-format

plexoos (dismissed):

.clang-format
```diff
@@ -0,0 +1,11 @@
+---
+BasedOnStyle: Google
+ColumnLimit: '100'
```
plexoos:
```suggestion
ColumnLimit: '120'
```

We can increase the max line length to be consistent with ROOT format:

https://github.com/root-project/root/blob/4389feb6d700c1d515014e56a3bcdee99cb6b4a5/.clang-format#L39

plexoos (commented):
Daniel, how do you propose to enforce the style? Have you looked at any solutions?

.clang-format
DanielWielanek: What about add aditional step in test? As far as I rememeber in FairRoot they have some scripts that perform additional check - if this check fail then you get an information.
However I have no Idea how root4star is tested.

plexoos:
While you are testing your new workflow, please don't trigger the regular build. Maybe there is a better way to disable a workflow but you can just remove the yaml file temporarily

JohnTheBlindMilkman (approved):

DanielWielanek:
Hi, Im' still trying to test this code - but It is hard to do this on single PC because this use github stuff.
With first commint I managed to use sucesfully the format check - however this check all files - what as you can expect triggers thousands of errors. Now I'm trying to limit check to changed files only.

plexoos:
Will this action check only modified files?

DanielWielanek:
Hi,
in theory it checks only modified files with some extensions like cxx or
hpp (source  codo in c++) that's why it took do long (default version
checked everything). I have checked on my repository and looks fine.

czw., 1 gru 2022, 21:34 użytkownik Dmitri Smirnov ***@***.***>
napisał:

JohnTheBlindMilkman (approved):

plexoos:
Thank you Daniel for implementing this check. I tested it in #461 and I think it works as intended. However, I wonder if you saw this action https://github.com/marketplace/actions/get-all-changed-files Looks like a straight forward solution to get a list of added or modified files in a PR

plexoos:
Moved from https://github.com/star-bnl/star-sw/pull/445#issuecomment-1344647192

> Actually is a funny story.
> For over two weeks, I had a problem with getting only the list of modified files. The script that worked on my local computer didn't work on GitHub servers. Then I found this plugin. As far as I remember, I was not able to connect both plugins correctly (I do not remember why). However, when I removed the plugin for getting the list of modified files, surprisingly, I noticed that the older script is working now. I suppose that by accident, I enabled some flags during tests :)
> So in principle, this plugin actually helped me :)

plexoos:
Here is another one :) I asked GPT to come up with a GitHub action to format C++ code and here is what I got:

Screen Shot 2022-12-09 at 5 19 59 PM

For a single attempt it is not bad at all!

DanielWielanek:
@plexoos  
Ok, so my question is - should be check all files when user call push request or only call this for changed files?
In case of checking everything it might take a lot of time (we have a lot of files).

PR #437, opened 2022-11-10
EventQA dependency

klendathu2k:
There is a MuDST dependency in the St_QA_Maker library that is not explicitly declared in the EventQA option.  Propose to make the dependence explicit.

To demonstrate:

```
$ root.exe 
root [0] .L bfc.C
root[1] bfc(-1,"EventQA nodefault")
```


plexoos (approved):

genevb:
Thanks for pointing this out, @klendathu2k . I know from where this dependency comes. Please allow me the opportunity to resolve this dependency by removing it rather than supporting it by modifying BFC.

klendathu2k:
Of course.  Better to remove a dependency that isn't needed than to 
support it.

starsdong:
Hi Gene, do you have other findings? Shall we proceed with this PR or go with another approach? Thanks

genevb:
Hi, Xin

Is there a rush? I informed the people who are responsible for this dependency, but I haven't told them it is urgent, and I'd actually prefer to close this PR unmerged and wait for the proper solution to play out over the next several weeks (months).

Thanks,
-Gene

starsdong:
Gene, there is no rush. It is glad to hear that you are on top of this. We can follow on this later certainly.

genevb:
Just a heartbeat note here (don't close this) ... this is still on me to follow up.

genevb:
I believe this PR is resolved now. @klendathu2k , would you be OK with closing it?

PR #440, opened 2022-11-13
[root6] StEvent/StEnumerations.cxx fix

klendathu2k:
StEvent/StEnumerations.cxx uses a _ROOT macro_ (under StRoot/macros/detectorId.C) to build a reflection map between the StDetectorId enum and the name of each enum.

Under root 5, the current code works fine.  But it is ugly.  We should at refactor the code to use a helper function compiled in the library.  This PR makes that update.

Under root 6, the current code fails.  Because root 6 tries to compile the macro, and the arguments to strncpy should not be const.  Again, this PR makes detectorId a helper function w/in StEnumerations.

---

We move the functionality provided by StRoot/macros/detectorId.C into the compiled StEvent/StEnumerations.cxx file.

We retain StRoot/macros/detectorId.C in case one or more user macros depend on it.


StRoot/StEvent/StEnumerations.cxx
```diff
@@ -3,10 +3,64 @@
...
+ gSystem->ExpandPathName(myPath);
+
+ int notExi = gSystem->AccessPathName(myPath.Data(),kFileExists);
+ if (notExi)    { ids[0]=-1; return;}
+ FILE *fp = fopen(myPath.Data(),"r");
+ if (!fp)       { ids[0]=-1; return;}
+ char buf[400];
+
+ int kase = 0;
+ while (2015) {
```
genevb: Why `2015`?

StRoot/StEvent/StEnumerations.cxx
```diff
@@ -30,23 +84,5 @@ StDetectorId detectorIdByName(const char *name)
...
-  for (int ifa=0;paths[ifa];ifa++) {
-    TString path(paths[ifa]);
-    gSystem->ExpandPathName(path);
-    int notExi = gSystem->AccessPathName(path.Data());
-    if (notExi) continue;
-    ids[0]=0;
-    TString com;
-    com.Form("%s((int*)%p,(const char**)%p)",path.Data(),ids,cds);
-    gROOT->Macro(com.Data());
-    assert(ids[0]);
```
genevb, line 49: That different programming languages evaluate negative numbers as true or false differently always drives me to be clear if there's the possibility of a negative value, i.e. `asserts(ids[0]!=0);`, but it's just a suggestion because C/C++ behaves consistently.

StRoot/StEvent/StEnumerations.cxx
klendathu2k: Possibly b/c Victor added this code in 2015?  Regardless, the point was to have an infinite loop, which is broken out of when the termination of the StDetectorId enum block was found.

StRoot/StEvent/StEnumerations.cxx
klendathu2k: Another good reason to remove that particular bit of code.  The assert statement is restored b/c we should do have the error detection.  The code should crash if there are no detectors defined.  (Indicates failure to parse the header file...)

plexoos:
> We retain StRoot/macros/detectorId.C in case one or more user macros depend on it.

I'd suggest to remove this macro entirely. It is likely that the only users of this are the Sti and Stv makers.

In the current implementation of an enum stringification in `detectorId(int*, char**)`, it parses a header file assumed to be in a certain location and following some format then interprets the parsed lines with ROOT. Would it make sense to use a `switch case` instead?

genevb (approved):
I'm OK with a staged removal of `StRoot/macros/detectorId.C` if you want to call it "deprecated" for now (which should probably include that specific wording written into the macro as an easily noticed comment) and remove it at a later date. No further comments/objections from me.

plexoos:
I don't think [4789bc8](https://github.com/star-bnl/star-sw/pull/440/commits/4789bc831b8e848e1d20b74bba26fe7d24d5d156) belongs here

klendathu2k:
> I don't think [4789bc8](https://github.com/star-bnl/star-sw/pull/440/commits/4789bc831b8e848e1d20b74bba26fe7d24d5d156) belongs here

Dang it.  Pushed out on the wrong branch.  How do I undo this?

plexoos:
`git revert <hash>`

plexoos:
Ok, I fixed the problem with compilation by reverting the changes in [e82889c](https://github.com/star-bnl/star-sw/pull/440/commits/e82889ceebe74348a115d42c0bb5705c4a8c91ed) Nothing seems to be obviously wrong with the proposed change but it shows how fragile the original implementation is. A simple `switch case` for a enum to string conversion seems even more appropriate now.

plexoos (approved):

PR #439, opened 2022-11-12
[root6] Add include statements to StarDb macros

klendathu2k:
ROOT6/cling needs to have include files for the table classes used in these macros.

This PR consists of one single change... to many files:  Adding the lines
```
#ifdef __CLING__
#  include "tables/TABLECLASSHEADER.h"
#endif
```
at the top of each macro.  We only make this change for ROOT6/cling, as ROOT5/cint is
happy with the status-quo.  (And ROOT5 does not like the added includes.)

This change has been tested for a single nightly test job (job 10).

PR #441, opened 2022-11-13
Set search paths for included files for ROOT C++ interpreter

plexoos:
It is my understanding that `TSystem::AddIncludePath()` sets search paths to look for header files only when ACLiC is used. On the other hand, `TInterpreter::AddIncludePath()` sets search paths used by the both interpreter and ACLiC (see the documentation at https://root.cern/root/html534/TCint.html#TCint:AddIncludePath)

### DIFF
--- a/StRoot/macros/rootlogon.C
+++ b/StRoot/macros/rootlogon.C
@@ -30,5 +30,10 @@
 
   std::cout << " *** Start at Date : " << TDatime().AsString() << std::endl;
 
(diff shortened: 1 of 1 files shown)


plexoos:
Gene, looks like we need your approval. This change is instigated by #439 which seems to work for both ROOT5 and ROOT6

genevb (approved):
I don't know of any reason to object. Some people may have their own rootlogon.C, but if so, then it is up to them to maintain their private versions.

PR #442, opened 2022-11-16
[root6] Eliminate crash when creating the MuDst

klendathu2k:
When creating the MuDst, ROOT6 has a problem with creating the TClonesArray
of StStrangeAssoc objects.   Running under GDB shows that TClass::GetClass enters
an infinite recursion...   almost as if the system is confused, and believes the class is it's
own base class.  

Disabling the call to initializeForWriting() on loading the share library resolves the issue.  
Though I don't know what side effects this may cause.

### DIFF
--- a/StRoot/StMuDSTMaker/COMMON/StMuDstMaker.cxx
+++ b/StRoot/StMuDSTMaker/COMMON/StMuDstMaker.cxx
@@ -479,9 +479,7 @@ void  StMuDstMaker::streamerOff() {
   StMuL3EventSummary::Class()->IgnoreTObjectStreamer();
 #ifndef __NO_STRANGE_MUDST__
   StStrangeMuDst::Class()->IgnoreTObjectStreamer();
-#if ROOT_VERSION_CODE >= ROOT_VERSION(6,0,0) 
   StStrangeAssoc::Class()->IgnoreTObjectStreamer(); 
-#endif
   StV0MuDst::Class()->IgnoreTObjectStreamer();
   StXiMuDst::Class()->IgnoreTObjectStreamer();
   StKinkMuDst::Class()->IgnoreTObjectStreamer();
--- a/StRoot/StMuDSTMaker/COMMON/StMuDstMaker.cxx
+++ b/StRoot/StMuDSTMaker/COMMON/StMuDstMaker.cxx
@@ -479,6 +479,9 @@ void  StMuDstMaker::streamerOff() {
   StMuL3EventSummary::Class()->IgnoreTObjectStreamer();
 #ifndef __NO_STRANGE_MUDST__
   StStrangeMuDst::Class()->IgnoreTObjectStreamer();
+#if ROOT_VERSION_CODE >= ROOT_VERSION(6,0,0) 
+  StStrangeAssoc::Class()->IgnoreTObjectStreamer(); 
+#endif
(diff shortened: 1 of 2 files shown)


veprbl:
The original code was introduced over two decades ago: https://github.com/star-bnl/star-sw/commit/a20c9a81dc7073db17f92f110bd9c7102d3abc3c#diff-9b5c84904f9c5ac927afddab6e4379e15fe6fa757afa4130bde3883f5ec15a3a

It is a bit suspicious that the modification to the global StObject streamer configuration is made conditional on loading of StStrangeMuDstMaker (which, IIRC, is itself loaded unconditionally in most of our BFC configurations).

I think, it would be nice to aim at removing this option for ROOT 5 as well. There may be some file size implications for MuDsts, but those can be addressed on a class-by-class basis.

veprbl (approved):
This advances the ROOT6 tests in CI to the `fCurrentRow` woes. Very exciting.

plexoos:
This is a great finding indeed. The function calls which seem to disturb ROOT6 can be moved to `StStrangeMuDstMaker`. That is how it is done in MuDst and PicoDst for the persistent objects:

https://github.com/star-bnl/star-sw/blob/25274b61e9c1f810c4f8f753c54449d28d8b5acc/StRoot/StMuDSTMaker/COMMON/StMuDstMaker.cxx#L478-L509

https://github.com/star-bnl/star-sw/blob/25274b61e9c1f810c4f8f753c54449d28d8b5acc/StRoot/StPicoDstMaker/StPicoDstMaker.cxx#L227-L250

genevb:
I consider the 2001 commit a hack on my part, so if it can be removed, that's a step in the right direction. And since I stopped maintaining this code something like 15 years ago, I don't feel protective of this code: please do whatever you need to. Thanks.

klendathu2k:
It certainly looked like a duck and walked like a duck... and hack rhymes with quack, so good to have confirmation.

It may be an issue of timing, that ROOT6 needs to defer the call to suppress the TObject streamer until later (eg initialization time) rather than load time.  I'll push that change shortly.

plexoos:
In the context of this change I tried to understand the relation between MuDst and StrangeMuDst. It appears to me that the latter has become a part of MuDst long time ago. If this is the case, can we get rid of the `__NO_STRANGE_MUDST__` branching in the code?

StRoot/StMuDSTMaker/COMMON/StMuDstMaker.cxx
```diff
@@ -479,6 +479,9 @@ void  StMuDstMaker::streamerOff() {
   StMuL3EventSummary::Class()->IgnoreTObjectStreamer();
 #ifndef __NO_STRANGE_MUDST__
   StStrangeMuDst::Class()->IgnoreTObjectStreamer();
+#if ROOT_VERSION_CODE >= ROOT_VERSION(6,0,0) 
+  StStrangeAssoc::Class()->IgnoreTObjectStreamer(); 
+#endif
```
plexoos: I think it is the right place for this but don't we really need to hide this from ROOT5? I think it should work fine for ROOT5

klendathu2k:
@jdbrice @nigmatkulov Would appreciate reviews as code-owners.

plexoos (approved):

PR #444, opened 2022-11-18
[root6] Set attributes on StrangeMuDstMaker in StBFChain, rather than using interpreter to configure

klendathu2k:
Cling appears to have problems with using ProcessLine to get a pointer to a maker... at least in this context and/or at least in the default environment.  So... 

Refactor to use maker attributes, and set configuration options at initialization time.  (This is also  cleaner and more inline with the StMaker philosophy).


genevb (approved):
Yes!

plexoos (approved):

Issue #431, opened 2022-11-08
Variable redefined in inner scope in OnlTools/Jevp/StJevpBuilders/fttBuilder.cxx


plexoos:
**Describe the bug**

In OnlTools/Jevp/StJevpBuilders/fttBuilder.cxx variables `ix0` and `ix1` are redefined inside an `if` statement, the assigned values do not change the variables defined in the outer scope:

https://github.com/star-bnl/star-sw/blob/e48dc9d0e7bbb7f84776bba64573401ee4dd8277/OnlTools/Jevp/StJevpBuilders/fttBuilder.cxx#L438-L446

A similar pattern can be seen for `iy0` and `iy1` variables:

https://github.com/star-bnl/star-sw/blob/e48dc9d0e7bbb7f84776bba64573401ee4dd8277/OnlTools/Jevp/StJevpBuilders/fttBuilder.cxx#L458-L466

**To Reproduce**

Compile a trivial program:

```c++
#include <iostream>

int main ()
{
    int ix0 = 1;

    { int ix0 = 2; }
    std::cout << ix0 << "\n";

    { ix0 = 3; }
    std::cout << ix0 << "\n";
}
```

**Expected behavior**

It appears the authors intended to modify the values of the variables inside the `if` statement

plexoos:
@jml985 git blame shows that you were the last one who touched these lines. Any comment from you?

jml985:
Yeah its obviously bug, but not critical for anything.  Its on my list to fix

jml985:
And its "blamed" on me because I do all the commits for the production Jevp stuff, not necessarily because I changed the code.

fgeurts:
Hi @jdbrice and @jml985 ,
Could you let me know if this issue has been resolved?

jml985:
Probably not resolved, but I did comment out the code to get rid of the compile warnings and make a pull request.   I'm not sure who the maintainer of this code is now, if there is one.   It certainly appears as if the author intended to  modify the indexes for this portion of the detector and screwed up redefining the variables, however, the code has been in use for several years so presumably it doesn't matter.   The worst case scenario is that the online plots show the data shifted by a bin, but I'm not even sure that this code is relevant as it might have been for a protype version of the detector.

sunxuhit:
Hi Frank, this is certainly beyond my knowledge. Probably Daniel or David should have a look.

fgeurts:
[...]

PR #445, opened 2022-11-18
build: Treat all warnings as errors

plexoos:
These changes do not affect how the code is built by default using `cons` but allows one to interrupt compilation when the compiler (GCC 4.8.5) produces a warning.

```
cons EXTRA_CXXFLAGS="-Werror"
```

The source files matching the regex patterns in `mgr/warnoff_dirs.txt` are excluded from the above rule in order to let CI pass with known problems. We hope this will prevent introduction of questionable code in all other directories or new packages.

See issue #392


mgr/ConsDefs.pm
```diff
@@ -468,7 +468,7 @@
 	$CERNLIB_CPPFLAGS .= " -DCERNLIB_GCC" . $CXX_MAJOR;
         print "CXX_VERSION : $CXX_VERSION MAJOR = $CXX_MAJOR MINOR = $CXX_MINOR\n";
 
-        $CXXFLAGS    = "$XMACHOPT -fPIC -pipe -Wall -Woverloaded-virtual";
+        $CXXFLAGS    = "$XMACHOPT -fPIC -pipe -Wall -Woverloaded-virtual -Werror";
```
veprbl: Can we *please* limit this to CI. Unconditional `-Werror` can be a huge pain for some of the end users.

plexoos: yes, we can

veprbl: So, for implementation, we could define a cons variable like `WARNINGS_AS_ERRORS` and make this line conditional on it. What do you think?

plexoos: For now, I tried to use the known to cons variable `EXTRA_CXXFLAGS`. So, for pull request builds we can run it as:
```
cons EXTRA_CXXFLAGS="-Werror"
```

plexoos:
Ready for review

plexoos:
Any comments from the code owners?

genevb (commented):
To clarify, this allows pre-existing warnings in old codes to not trigger an alert in CI, while new codes will be monitored for such warnings and either fixed or added to this exclusion list?

Is there a clear (documented?) place to make change that would allow someone compiling the whole library to see the warnings for all codes again?

plexoos:
Excellent questions, Gene. The answer to the first question is yes, that is how it is intended to work. With only one comment that the goal should be to shrink the exclusion list rather than adding new codes with questionable constructs. Warnings detect real bugs #431 

For the second one, no special action needs to be taken, the warnings for the excluded paths will still appear in the build log, they just won't fail CI.

plexoos:
If no more comments can we merge?

Dockerfile
```diff
@@ -7,6 +7,8 @@ ARG compiler=gcc485
 
 FROM ghcr.io/star-bnl/star-spack:v0.2.3-${starenv}-${compiler}
 
+ARG compiler
```
veprbl, line 10: There is an `ARG compiler=gcc485` above. Do we need this line?

veprbl (approved):

Dockerfile
plexoos: Yes, it's a feature. That is how you reuse a variable set before the FROM command

veprbl: Good to know!

plexoos:
@klendathu2k your last word 🥺

klendathu2k (approved):
Should be fine, and if/when it breaks the build we will fix.

DanielWielanek:
Actually is a funny story.
For over two weeks, I had a problem with getting only the list of modified files. The script that worked on my local computer didn't work on GitHub servers. Then I found this plugin. As far as I remember, I was not able to connect both plugins correctly (I do not remember why). However, when I removed the plugin for getting the list of modified files, surprisingly, I noticed that the older script is working now. I suppose that by accident, I enabled some flags during tests  :)
So in principle, this plugin actually helped me :)

plexoos:
Cool story, wrong thread 🤣 

Moved to https://github.com/star-bnl/star-sw/pull/436#issuecomment-1344833413

PR #447, opened 2022-11-21
Minimum viable pentagonal sTGC geometry with misalignment

jdbrice:
This is a new sTGC geometry (based on CAD models) which implement the final detector's pentagonal structure.
Setup for misalignment.
This is a stripped down version (without brackets, supports etc.) as we work on resolving issues with full geom.

Updates to geometry definition, St_geant_maker, and g2t 

@klendathu2k I probably want to update the volume definition but I couldn't remember the rules / how to make sense of it, some help might be appreciated.


plexoos:
Just a reminder. Every commit to a PR branch triggers CI. If the change is only a clean-up or a change in comments it is better to skip the actions by adding "[skip ci]" or "[skip actions]" to the log message as explained here: https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs

Yes, we use the GitHub's resources for free but let's be nice to Microsoft and not waste the shared resources 🙇 . Currently our test builds are quite heavy and take a few hours. We are working on speeding up the builds with cache but until then let's use responsibly.

jdbrice:
On closer inspection the volume ID is already good as is, so no change is needed there. So AFAIK this is ready as is.

StRoot/St_geant_Maker/St_geant_Maker.cxx
```diff
@@ -1363,7 +1363,7 @@ int St_geant_Maker::Make() {
   AddHits<St_g2t_fts_hit>( "FTSH",{"FTSA","FSIA", "FSIB", "FSIC"}, "g2t_fts_hit", g2t_fts ); 
 
   AddHits<St_g2t_fts_hit>( "FSTH",{"FTOS","FTIS","FTUS"},          "g2t_fsi_hit", g2t_fts );
-  AddHits<St_g2t_fts_hit>( "STGH",{"TGCG"},                        "g2t_stg_hit", g2t_stg );
```
klendathu2k, line 1366: Just a comment here.  The design of the AddHits function allows us to support hit collections in a backwards compatible way.  So if we still have a geometry model with the TGCG volume, that can be included in the list.

StarVMC/Geometry/StgmGeo/StgmGeo1.xml
```diff
@@ -1,968 +1,548 @@
...
-   will be created for each disk based on the dimensions
-   provided in the DISK structure -->
-  <Volume name="FTSD" comment="FTS disk">
-      
-      <Attribute for="FTSD" seen="0" colo="3" />
-      <Shape type="TUBE"
-      rmin="FDSK_rmin"
-      rmax="FDSK_rmax"
-      dz="FDSK_thksupp" />
+<Document  file="StarVMC/Geometry/StgmGeo/StgmGeo1.xml">
```
klendathu2k, line 1: This is a full replacement of the StgmGeo1 module.

StarVMC/Geometry/StgmGeo/StgmGeo1.xml
```diff
@@ -1,968 +1,548 @@
...
+            <Component name="C" a="12" z="6" w="28"  />
+            <Component name="H" a="1" z="1" w="54"  />
+            <Component name="O" a="16" z="8" w="11"  />
+        </Mixture>
+
+      <!-- Create and place the STGC Mother volume in the CAVE -->
+
+        <Create    block="STGM" />
+        <Placement block="STGM" in="CAVE" konly="ONLY" x="0" y="5.9" z="307.65" >
+            <Misalign table="Geometry/stgc/stgcOnTpc" row="0"/>
```
klendathu2k, line 219: The corresponding "ideal" misalignment tables should be a part of this PR.... 
StarDb/stgc/stgcOnTpc.20221220.000000.C   (explicit time stamp initialization)

Should return an ideal (unit matrix) table (eg StarDb/Geometry/tpc/TpcPosition.C)

pams/sim/g2t/g2t_stg.F
```diff
@@ -14,42 +14,43 @@ FUNCTION  G2T_STG (g2t_track_h,   g2t_track ,
...
-*
-         g2t_fts_hit(i).next_tr_hit_p = g2t_track(trac).hit_fts_p
-         g2t_track(trac).hit_fts_p    = i
-         g2t_track(trac).n_stg_hit    = g2t_track(trac).n_stg_hit + 1
-
-       enddo    
+      "Setup readout for the GAS volumes STG*"
+
+      Do Isys=1,3
+           if (isys.eq.1) CALL g2r_get_sys('STGM','STGP',Iprin,Idigi)
```
klendathu2k, line 20: Manages the three subvolumes that make up the sensitive pentagonal shape.

pams/sim/g2t/g2t_stg.F
```diff
@@ -14,42 +14,43 @@ FUNCTION  G2T_STG (g2t_track_h,   g2t_track ,
...
-           g2t_fts_hit(i).x(3)        = xx(3)
-           g2t_fts_hit(i).p(1)        = p(1)
-           g2t_fts_hit(i).p(2)        = p(2)
-           g2t_fts_hit(i).p(3)        = p(3)
-
-         g2t_fts_hit(i).tof           = tof
-         g2t_fts_hit(i).de            = Elos
-         g2t_fts_hit(i).ds            = Step
-         g2t_fts_hit(i).track_p       = trac
-         g2t_fts_hit(i).volume_id     = volume
```
klendathu2k, line 44: I presume that the overall structure of the detector remains the same... i.e. the paths to the sensitive volumes are the same, modulo the change in the names of the volumes.  But there are now   three SDs where previously there was only one.  So, does the g2t_volume_id encoding need to change? (pams/sim/g2t_volume_id.g).

klendathu2k (changes requested):
I expect the geometry has been thoroughly vetted w/in the working group.   It looks like this is essentially ready to commit.... modulo the two places I noted -- 
1) initial (ideal) misalignment tables specified 
2) if a change in the g2t_volume_id is required, we should do so now

StarVMC/Geometry/StgmGeo/StgmGeo1.xml
```diff
@@ -1,968 +1,548 @@
...
+                <Misalign table="Geometry/stgc/pentOnStation"  row="3" />
+                <Misalign table="Geometry/stgc/stationOnStgc"  row="0" />
+            </Placement>
+
+            station = 1;
+            <Info format="Positioning sTGC {2I}">station</Info>
+
+
+            zplane = -3.95; 
+            <Placement in="STGM" y="0" x="0" z="zplane" block="STFM"  >
```
klendathu2k: Even though these are marked as "assembly" volumes... they should be placed as konly="many", otherwise starsim will have problems with the overlaps.

klendathu2k:
Based on my reading of the code I see the following structure which is relevant for the hits...

STGM:
  STFM Full Module (4 per station)
  STMG Active Module Gas (2 per full module)
    STGP Triangle (1 per gas module)
    STGL Long box (1 per gas module)
    STGS Short box (1 per gas module) 

HALL, CAVE, STGM are all singly placed.  So... the numbv
array in g2t_volume_id will only consider STFM, STMG, and 
the STG* volumes in the numbering scheme
...

numbv(1) ==> will indicate the full module.
  1-4 station 1
  5-8 station 2
  9-12 station 3
  13-16 sation 4

numbv(2) ==> will indicate the gas module
  1 front 
  2 back 

You can / should add print statements to confirm the expectations.
  
At present, g2t_volume_id will just return numbv(1).
This will indicate which of the 16 full modules was 
hit.  If you are scoring hits with global coordinates,
this may be sufficient.  However, if you need to encode
the specific volumes which are struck, I would suggest
the following:

```
Integer function g2t_stg_volume_id ( numbv )
  Integer, intent(in) :: numbv(15)    
  Integer :: station, plane, gasmod, shape     
  station = numbv(1) / 4
  plane   = (numbv(1) -1) % 4 + 1
  gasmod  = numbv(2)
  shape = 0
  if (cd=='STGP') shape = 1
  if (cd=='STGL') shape = 2
  if (cd=='STGS') shape = 3
  g2t_stg_volume_id = shape + 10*gasmod + 100*plane +1000*station
End
```

Then g2t_volume_id should return a single unique ID for
each volume...  You can decode this in full and use / ignore what you need in the slow simulator.

StRoot/St_geant_Maker/St_geant_Maker.cxx
jdbrice: There is no longer a TGCG volume, so I remove it and replaced it with the three new volume names

akioogawa (approved):

StarVMC/Geometry/StgmGeo/StgmGeo1.xml
jdbrice: Addressed in 860d2af982f1094717c4fbbc522aee76fbbcfbdf

jdbrice:
OK, so I mostly took your suggested changes for the volume_id with two exceptions:
1. I dont care about the "shape" - it is an artifact not having a pentagon GEANT shape.
2. The station number was off by one in your example. Needs to be (numbv(1) -1)/4

klendathu2k (approved):
Looks good to me / lets get this merged.

StarVMC/Geometry/StgmGeo/StgmGeo1.xml
jdbrice: Done with 0aa214c1c8d4d1eaab23d0c0537d5cf927ab418b

PR #449, opened 2022-11-23
[root6] Address issues in TTable when interpreted by Cling

plexoos:
Our ROOT6 tests indicated a problem with the ROOT's TTable class when it is interpreted by the bfc.C macro. A few fixes seem to take care of that but now we need to build and use the local code from StRoot/Table instead of the external `libTable` provided by ROOT. This is achieved by building StRoot/Table conditionally only when external dependency on ROOT >= 6 is detected.

### DIFF
--- a/StRoot/Table/TTable.cxx
+++ b/StRoot/Table/TTable.cxx
@@ -2465,3 +2465,17 @@ TTable::piterator::piterator(const TTable *t,EColumnType type): fCurrentRowIndex
    }
 } // piterator(TTable *)
 
+void TTable::iterator::operator++()    { ++fCurrentRow;   }
+void TTable::iterator::operator++(int) {   fCurrentRow++; }
+void TTable::iterator::operator--()    { --fCurrentRow;   }
+void TTable::iterator::operator--(int) {   fCurrentRow--; }
+TTable::iterator TTable::iterator::operator+(Int_t idx)   { std::vector<Long_t>::iterator addition   = fCurrentRow+idx; return TTable::iterator(*fThisTable,addition); }
+TTable::iterator TTable::iterator::operator-(Int_t idx)   { std::vector<Long_t>::iterator subtraction = fCurrentRow-idx; return TTable::iterator(*fThisTable,subtraction); }
+void TTable::iterator::operator+=(Int_t idx)  {  fCurrentRow+=idx; }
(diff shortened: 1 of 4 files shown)


klendathu2k (approved):
Looks good to me.

StRoot/Table/TTable.cxx
```diff
@@ -2465,3 +2465,17 @@ TTable::piterator::piterator(const TTable *t,EColumnType type): fCurrentRowIndex
    }
 } // piterator(TTable *)
 
+void TTable::iterator::operator++()    { ++fCurrentRow;   }
```
veprbl, line 2468: Ingenious!

mgr/Conscript-standard
```diff
@@ -11,12 +11,8 @@ my $STAR_SYS     = $env->{ENV}->{STAR_SYS};
 my $STAR_VERSION = $env->{ENV}->{STAR_VERSION};         # print "#####  $STAR_VER\n";
 my $AFS_RHIC     = $env->{ENV}->{AFS_RHIC};
 
-my $ROOT_FEATURES = `root-config --features`;
```
veprbl, line 14: What about the case when root is built with the tables? Are include/library paths preferring our version?

veprbl (approved):

mgr/Conscript-standard
plexoos: Yes, in general everything in our build/install directory (i.e. .slXX_gccNNN) precedes the external locations. With those conditions in Conscript we have never really built or tested the local TTable because both ROOT 5.34 and 6.16 still have libTable support.

There is, however, a different issue... Our local TTable in StRoot is from ROOT 6.16 and it is not compatible with ROOT 5.34. The CI fails due to that. Should we freeze our TTable at ROOT 5.34 if it still works with ROOT6 as dependency?

veprbl: The error states that our TTable provides 8 arguments to `TDefaultInitBehavior::CreateClass` instead of 9. I doubt we can resolve this by just downgrading TTable. Either would have to have conditionals, or give up on providing our TTable for ROOT5.

plexoos: Just for the record, I went with the second option, i.e. our local TTable is built and used only when ROOT version is 6 and higher. The case of ROOT5 should not be affected by this. The PR's description has been updated accordingly.

PR #452, opened 2022-11-28
ci: Enable ROOT6 tests

plexoos:


StRoot/macros/bfc.C
```diff
@@ -140,7 +140,7 @@ void bfc(Int_t First, Int_t Last,
   // "-" sign before requiest means that this option is disallowed
   // Chain = "gstar" run GEANT on flight with 10 muons in range |eta| < 1 amd pT = 1GeV/c (default)
   // Dynamically link some shared libs
-  if (gClassTable->GetID("StBFChain") < 0) Load(Chain);
+  Load(Chain);
```
genevb: What issue does this solve? The `if` statement seems a reasonable condition, doesn't it?

plexoos: The issue is that the `Load()` function is not called when using ROOT6. It is probably due to this statement at the head of the macro `#pragma cling load("StBFChain")` which loads the library and makes the class know to ROOT

genevb: Ah....I can believe that. But then to retain the same functionality, wouldn't you put something like this?

    #ifndef __CLING__
    if (gClassTable->GetID("StBFChain") < 0)
    #endif
    Load(Chain);

plexoos: Yes, but what is the underlying reason for not loading libSt_base, liblog4cxx, StStarLogger, libStChain, libStUtilities, and libStBFChain when ROOT knows about StBFChain class?

plexoos:
This one is ready to go. Looks like all of the ROOT6 tests we have set up in CI pass. Only the usual checks are performed, i.e. the bfc.C macro can run on small real and simulated samples from various years without checking the quality of the output. Let me know if more inputs and options should be checked this way.

iraklic (approved):
Do I understand it correctly that all these includes were not needed with ROOT5 and now they are needed in ROOT6?

plexoos:
@iraklic Yes, rootcint is more strict in that sense. It needs to see the definitions of all used types while processing the macros

veprbl (approved):
Happy to see this milestone reached

Issue #455, opened 2022-12-05
Error in CreateGeometry.h in DEV


genevb:
The following reconstruction chain in DEV (ROOT 5, not ROOT 6):

    root4star -b -q -l bfc.C\(10,\"P2021a,StiCA,EbyET0,BEmcChkStat,-hitfilt,SCScalerCal,-mtd,-btof,-etofA,-picoWrite\",\"/star/data03/daq/2021/180/22180043/st_physics_22180043_raw_2000004.daq\"\)

....results in the following error after approximately 1-2 minutes of execution, and after approximately 2120 lines of output.

LoadTable: .L /afs/rhic.bnl.gov/star/packages/DEV/StarDb/AgMLGeometry/Geometry.y2021a.C
Error: cannot open file "StBFChain/StBFChain.h"  /afs/rhic.bnl.gov/star/packages/DEV/StarDb/AgMLGeometry/CreateGeometry.h:2:
*** Interpreter error recovered ***
root4star: .sl73_gcc485/OBJ/StRoot/St_db_Maker/St_db_Maker.cxx:932: virtual TDataSet* St_db_Maker::LoadTable(TDataSet*): Assertion `!ee' failed.
Abort

genevb:
I should add that this error does not occur in SL22c, so it is likely to do with something merged in the last couple months.

klendathu2k:
That narrows it down ;)  Looking at it now.

klendathu2k:
Culprit is PR #448.  (Confirmed by wrapping the #include statement with #ifdef __CLING__). Suspect that ROOT5 / CINT does not have the include search path for StarDb.  Easy technical fix.. once I find the right place to make it.

klendathu2k:
Okay... I shouldn't debug without a cup of coffee handy.   StarDb should not be the issue.  Include path needs to see StRoot/StBFChain, and as far as I can tell... it should be setup correctly.  Investigating...

plexoos:
I can't see this test job in the usual location in `/star/u/starreco/devtest/`. What am I missing?

Run 22180043 does not show up anywhere on this page https://www.star.bnl.gov/devcgi/weekDEVjobStatus.pl either
But there is a bunch of jobs with `Assertion failed` status. Is it related? When did this happen?

genevb:
> I can't see this test job in the usual location in `/star/u/starreco/devtest/`. What am I missing?
> 
[...]

PR #456, opened 2022-12-05
Hide include directive from ROOT5 interpreter

klendathu2k:
Proposed fix for Issue #455

It is not clear to me from the tests I've done on issue #455 why StBFChain/StBFChain.h is not located in the include path by root5 / cint.  Simplest fix is to hide the include directive from CINT.

### DIFF
--- a/StarDb/AgMLGeometry/CreateGeometry.h
+++ b/StarDb/AgMLGeometry/CreateGeometry.h
@@ -1,4 +1,4 @@
-#ifdef __CLING__
+#ifndef __CINT__
 #include "StBFChain/StBFChain.h"
 #endif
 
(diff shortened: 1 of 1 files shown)


plexoos:
There is no reason to hide the header from ROOT5/CINT and the CI tests confirm just that. The reason why ROOT5/cint can't find the header is because ROOT5 is not given the paths to search for it in interpreter mode. It is done by using `gInterpreter->AddIncludePath()` as in #441 . I don't really understand the resistance to update the rootlogon.C macro to include those lines... I think the same effect can be achieved if one adds these calls directly to the bfc.C macro before calling `gSystem->ProcessLine()`...

My point is that we should minimize the branching of the code for ROOT5 and ROOT6 and use #ifdefs only if nothing else works.

klendathu2k:
Agreed.  I had not appreciated that the gInterpreter -> AddIncludePath(...) function only accepted one path at a time.  Will close this PR.

plexoos (approved):
Having no control over the system-wide root config files, this seems like a viable solution (at least for now)

genevb (approved):
Seems OK to me.

PR #458, opened 2022-12-06
bfc.C: Set search paths for ROOT interpreter

plexoos:


veprbl:
This only works for `root4star bfc.C(...)`, but not for macros having their own entry point.

plexoos:
That is correct. I am trying to avoid modifying rootlogon.C in $ROOTSYS/etc/ on the farm.

genevb:
So the idea is to move functionality that is in the STAR-specific rootlogon.C for ROOT5 into bfc.C for ROOT6? If the STAR-specific rootlogon.C is to be deprecated, I recommend doing that in coordination with Jerome, perhaps with @starsdong  involved if you're concerned about having such a discussion on your own.

plexoos:
> So the idea is to move functionality that is in the STAR-specific rootlogon.C for ROOT5 into bfc.C for ROOT6?

Um... no, I guess... The idea is to somehow let ROOT5 know where to search for included header files. We know that ROOT6 requires the header files and ROOT5, it seems, can work without them but does not mind having them if it can find the files. The common denominator is to include header files

All solutions I can think of have their cons and pros. This one, for example, does not require any change in rootlogon.C but is not universal and applies only to the specific macro. The solution Jason proposed in #456 is also independent of rootlogon.C but splits the code for ROOT5/6. The solution with .rootrc in either $HOME or $PWD is good because it can point to rootlogon in the install location (under version control in some sense) but is not readily available for all users. Finally, the original suggestion to update the existing rootlogon in $ROOTSYS should be very transparent but, in my opinion, is the least reproducible in the long term. Unless I am missing something, I don't see a preferred solution so let's just pick one. How about the latter? If there is an agreement, those six lines with the search paths can be added

plexoos:
We'll update $ROOTSYS/etc/rootlogon.C with search paths for header files

PR #463, opened 2022-12-12
update TpcRS parameters for Run19 AuAu19.6 GeV embedding

zhux97:


starsdong:
Hi Xianglei, do you have some slides showing MC/data comparison with this set of tuning parameters? It would be good to attach here for documentation purpose. Thanks

zhux97:
Hi Xin, Sure, please find all the data/mc comparisons for proton embedding here,
https://www.star.bnl.gov/protected/embedding/AutoIndex.php?dir=pr463datamc/
PID cuts for real data proton is 0.6<|mass2|<1.2. you can check the realmass2.pdf for possible contamination and realnsigma.pdf for the nsigma_proton of those proton after the mass2 cut.

starsdong:
Thank you, Xianglei.
Irakli, could you please take a review of the change at your earliest convenience? We should try to merge this in asap.

plexoos (approved):
I believe the authors do not consider reproducibility of simulation results as something to be concerned of. Fine.

iraklic:
I looked at the histos and the MC/RC matching looks ok, but I believe the results were presented and vetted, correct? with this assumption I will approve the new parameters.

@plexoos, do you mean that we are loosing the old parameterset by replacing the y2019 without a timestamp, or something else? @starsdong , can you remind me if there is a corresponding table for these parameters or it's just a 'cint' file?

iraklic (approved):

starsdong:
@iraklic my understanding is that for historical reasons, all these parameters are stored only here. There are no tables for this.

plexoos:
> do you mean that we are loosing the old parameterset by replacing the y2019 without a timestamp, or something else?

Yes. But it is not really lost, it is just not readily available.

PR #467, opened 2022-12-15
cons: Extend SKIP_DIRS to match arbitrary path

plexoos:
This fix allows to skip top level dirs and/or their subdirs such as
`OnlTools`, `asps`, etc.


klendathu2k (approved):
Okay.  Seems useful to (e.g.) exclude individual event generators, sub directories in  pool areas, etc...

PR #462, opened 2022-12-09
Remove StBFChain dependence in StarDb/AgMLGeometry/CreateGeometry.h

klendathu2k:
This PR Should resolve #455.

The StBFChain dependence supported two features in this macro. 

1) It allowed us to name the geometry cache file based on the output filename for the chain.
2) It allowed us to detect when Sti (CA or V) was present in the chain options.

Item 1 is nice to have but not required.

Given the "assert(0)" in the block of code which detects Sti... we don't actually use item 2.  (I suspect it was deprecated at some point).

Therefore we remove the dependency on StBFChain.

PR #477, opened 2023-01-04
Invoke CreateGeometry() from StBFChain

plexoos:
- Add function CreateGeometry() to StarGeometry library
  The code from StarDb/AgMLGeometry/CreateGeometry.h is copied to

      StarVMC/StarGeometry/CreateGeometry.cxx
      StarVMC/StarGeometry/CreateGeometry.h

- StBFChain: Call compiled CreateGeometry()

### DIFF
--- a/StRoot/StBFChain/StBFChain.cxx
+++ b/StRoot/StBFChain/StBFChain.cxx
@@ -18,6 +18,7 @@
 #include "StEnumerations.h"
 #include "TTree.h"
 #include "TEnv.h"
+#include "StarVMC/StarGeometry/CreateGeometry.h"
 #define STAR_LOGGER 1
 // PLease, preserve the comment after = { . It is used for documentation formatting
 //
@@ -1023,21 +1024,7 @@ Int_t StBFChain::Init() {
 	  if (GetOption("VmcGeo")) {
 	    path  = "./StarDb/VmcGeo:$STAR/StarDb/VmcGeo";
 	  }
-	  TString geom("Geometry.");
-	  geom +=  DbAlias[i].geometry;
-	  geom += ".C";
-	  Char_t *file = gSystem->Which(path.Data(),geom,kReadPermission);
-	  if (file) {
-	    LOG_INFO << "StBFChain::Init force load of " << file << endm;
-	    TString command = ".L "; command += file;
-	    gInterpreter->ProcessLine(command);
-	    gInterpreter->Calc("CreateTable()");
-	    command.ReplaceAll(".L ",".U ");
-	    gInterpreter->ProcessLine(command);
-	    delete [] file;
-	  } else {
-	    LOG_INFO << "StBFChain::Init file for geometry tag  " << geom << " has not been found in path" << path << endm;
-	  }
+	  CreateGeometry(DbAlias[i].geometry);
 	  break;
 	}
       }
--- a/StarVMC/StarGeometry/CreateGeometry.cxx
+++ b/StarVMC/StarGeometry/CreateGeometry.cxx
@@ -1,0 +1,75 @@
+#include <TGeoManager.h>
+
+#include "StBFChain/StBFChain.h"
+#include "StarVMC/StarAgmlLib/AgModule.h"
+#include "StarVMC/StarAgmlLib/StarTGeoStacker.h"
+#include "StarVMC/StarGeometry/StarGeo.h"
+
+extern StBFChain* chain;
+
+/**
+ * Adapted from StarDb/VmcGeometry/CreateGeometry.h
+ */
+TDataSet *CreateGeometry(const Char_t *name="y2011") {
+  TObjectSet *geom = 0;
+  if ( gGeoManager ) { 
+    cout << "AgML geometry:  Existing TGeoManager " << gGeoManager->GetName() 
+	 << " detected, ignoring request for " 
+	 << name << endl;
+    return geom;
+  }
+
+  // Cache geometry to a TFile.  Geometry will be restored from TFile on subsequent calls.
+  TString filename = "";  
+  if  (chain)  { filename = chain->GetFileOut(); if ( filename=="" ) filename = chain->GetFileIn();  }
+  else { filename = name;  }
+
+  // Strip out @ symbol
+  filename = filename.ReplaceAll("@",""); 
+  // Strip off the last extention in the filename
+  filename = filename( 0, filename.Last('.') );
+  // Append geom.root to the extentionless filename
+  filename+=".geom.root";
+
+  // Detect second call to the system
+  if ( AgModule::Find("HALL") ) {
+    if ( chain->GetOption("Sti")    ||
+	 chain->GetOption("StiCA")  ||
+	 chain->GetOption("StiVMC") ){
+      cout << "AgML geometry:  HALL exists.  Restore from cache file " 
+	   << filename.Data() << endl;
+      gGeoManager = 0;
+      assert(0);
+      TGeoManager::Import( filename );
+      assert(gGeoManager);
+    }
+    return geom;
+  }
+
+  cout << "AgML: Building geometry " << name << " " << endl;
+
+  // Create the geometry using TGeo
+  AgBlock::SetStacker( new StarTGeoStacker() ); 
+
+  Geometry *build = new Geometry();  
+
+  // Suppress copious ROOT warnings 
+  Long_t save = gErrorIgnoreLevel; gErrorIgnoreLevel = 9999;
+  build->ConstructGeometry(name);
+  gErrorIgnoreLevel = save;
+
+  if ( gGeoManager ) 
+    {
+      gGeoManager->CloseGeometry();
+      geom = new TObjectSet("Geometry",gGeoManager, false );
+      geom -> SetTitle( Form("AgML Geometry: %s",name) );
+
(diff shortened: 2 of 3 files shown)


genevb (commented):
No objection.

StarVMC/StarGeometry/CreateGeometry.cxx
```diff
@@ -0,0 +1,75 @@
...
+  // Detect second call to the system
+  if ( AgModule::Find("HALL") ) {
+    if ( chain->GetOption("Sti")    ||
+	 chain->GetOption("StiCA")  ||
+	 chain->GetOption("StiVMC") ){
+      cout << "AgML geometry:  HALL exists.  Restore from cache file " 
+	   << filename.Data() << endl;
+      gGeoManager = 0;
+      assert(0);
+      TGeoManager::Import( filename );
```
veprbl, line 43: This is a new behaviour, right? This may be surprising to users who expect that chains only write output ROOT files.

plexoos: No, nothing new --- since at least 2013 if I read the blame correctly. As noted in the description, the code is just copied from StarDb/AgMLGeometry/CreateGeometry.h as is. (There was an attempt to clean up this header in #462 but that broke something else, so can revisit if this compiled version works but in another PR) The CreateGeometry.h header is(was) "interpreted" (via other macros) by the snippet I removed from StRoot/StBFChain/StBFChain.cxx

plexoos:
> No objection.

No objection to merging this failing code? Hm... interesting 🤔 😄

StarVMC/StarGeometry/CreateGeometry.cxx
veprbl: Oh, I though this code was somehow equivalent to the one removed from StBFChain. But, actually, it just relies on the fact that for each XXX, the StarDb/AgMLGeometry/Geometry.XXX.C only performs `CreateGeometry("XXX")`.

Thanks for looking up the blame. If this is an existing behavior, that's good to know, that's not an issue of this PR.

StRoot/StBFChain/StBFChain.cxx
```diff
@@ -1028,21 +1029,7 @@ Int_t StBFChain::Init() {
...
-	    LOG_INFO << "StBFChain::Init force load of " << file << endm;
-	    TString command = ".L "; command += file;
-	    gInterpreter->ProcessLine(command);
-	    gInterpreter->Calc("CreateTable()");
-	    command.ReplaceAll(".L ",".U ");
-	    gInterpreter->ProcessLine(command);
-	    delete [] file;
-	  } else {
-	    LOG_INFO << "StBFChain::Init file for geometry tag  " << geom << " has not been found in path" << path << endm;
-	  }
```
veprbl, line 1040: I suppose, we might want to remove the macro files, if they are not used anymore?

StarVMC/StarGeometry/CreateGeometry.cxx
```diff
@@ -0,0 +1,75 @@
+#include <TGeoManager.h>
+
+#include "StBFChain/StBFChain.h"
+#include "StarVMC/StarAgmlLib/AgModule.h"
+#include "StarVMC/StarAgmlLib/StarTGeoStacker.h"
+#include "StarVMC/StarGeometry/StarGeo.h"
+
+extern StBFChain* chain;
```
veprbl, line 8:
```output
dlopen error: /star-sw/.sl79_gcc485/LIB/libStarGeometry.so: undefined symbol: chain
```
Looks like we've never landed the 205f44ac5beac1d92a81f61f6fb8934e4d720f25.

plexoos: Ah, yes! I think that is the reason why half of the tests crashed

StRoot/StBFChain/StBFChain.cxx
klendathu2k: The StarDb/AgMLGeometry macros are still required.  Any maker in the chain can request the geometry through GetDatabase("VmcGeometry") at any point in the lifetime of the application (during or after StChain::Init()).  When that call is made, the DB looks for macros under StarDb/AgMLGeometry (or AgiGeometry depending on alias settings...).

The initialization of the geometry in StBFChain is only performed when certain makers are in the chain (see line 1015).

StRoot/StBFChain/StBFChain.cxx
```diff
@@ -1028,21 +1029,7 @@ Int_t StBFChain::Init() {
...
+	  CreateGeometry(DbAlias[i].geometry);
```
klendathu2k, line 1027: CreateGeometry will not be found if the agml libraries are not loaded, e.g. if the AgML chain option is not specified.   We switched to using AgML for the geometry description ~ 2012.  Prior to that the geometry was supported from macros translated directly from the AgSTAR geometries (StarDb/AgiGeometry/).  Thpse macros also depend on "CreateGeometry"... albeit a much simpler macro.

StarVMC/StarGeometry/CreateGeometry.cxx
veprbl: Now ROOT5 seems to not handle the `extern *chain` in `bfc.C`.

genevb:
Maybe this is already known to @klendathu2k and @plexoos , but I figured I would document that I see that the current method of this in DEV does create some (not large) lost memory, according to valgrind in tests I was doing for something else:

==15423== 4 bytes in 1 blocks are definitely lost in loss record 15,244 of 312,718
==15423==    at 0x4029885: operator new(unsigned int) (vg_replace_malloc.c:328)
==15423==    by 0x150EDEC5: G__StarGeo_Cint_558_0_2(G__value*, char const*, G__param*, int) (StarGeo_Cint.cxx:544)
...
==15423==    by 0x4284B0F: TCint::Calc(char const*, TInterpreter::EErrorCode*) (TCint.cxx:655)
==15423==    by 0x1113149E: StBFChain::Init() (StBFChain.cxx:1039)

Where StBFChain.cxx:1039 is the call to CreateGeometry:

            LOG_INFO << "StBFChain::Init force load of " << file << endm;
            TString command = ".L "; command += file;
            gInterpreter->ProcessLine(command);
            gInterpreter->Calc("CreateTable()");             // <-- LINE 1039
            command.ReplaceAll(".L ",".U ");
            gInterpreter->ProcessLine(command);

Which is clear because the following is the only "force load of" in the output of the job:

BFC:INFO  - StBFChain::Init force load of /afs/rhic.bnl.gov/star/packages/DEV/StarDb/AgMLGeometry/Geometry.y2021a.C

genevb:
Actually it's a larger amount of "definitely lost memory" than I noted in my previous message as I see now several other such messages in the valgrind output, culminating in this biggest definite loss. I guess it may not be a big deal if the loss occurs after the geometry objects are done being used?


==15423== 107,204 (84,100 direct, 23,104 indirect) bytes in 725 blocks are definitely lost in loss record 312,478 of 312,718
==15423==    at 0x4029885: operator new(unsigned int) (vg_replace_malloc.c:328)
==15423==    by 0x4254E6E: TStorage::ObjectAlloc(unsigned int) (TStorage.cxx:325)
==15423==    by 0x808DF26: TObject::operator new(unsigned int) (TObject.h:156)
==15423==    by 0x13A0D2AC: AgPosition::rotation() (AgPosition.cxx:146)
==15423==    by 0x13A0D5EA: AgPosition::matrix() (AgPosition.cxx:190)
==15423==    by 0x13A356FC: StarTGeoStacker::Position(AgBlock*, AgPosition) (StarTGeoStacker.cxx:1208)
==15423==    by 0x1551353E: WCALGEO1::WMOD::Block(AgCreate) (WcalGeo1.cxx:206)
==15423==    by 0x139F4531: AgBlock::Create(char const*) (AgBlock.cxx:79)
==15423==    by 0x1551673E: WCALGEO1::WcalGeo1::ConstructGeometry(char const*) (WcalGeo1.cxx:828)
==15423==    by 0x149EB69C: WCAL::WCALv1::construct() (FcsmConfig.cxx:104)
==15423==    by 0x14FEE8E2: y2021a::construct(std::map<std::string, int, std::less<std::string>, std::allocator<std::pair<std::string const, int> > >) (StarGeo.cxx:1700)
==15423==    by 0x150E643D: StarGeometry::Construct(char const*) (StarGeo.cxx:19972)
==15423==    by 0x150F0784: Geometry::ConstructGeometry(char const*) (StarGeo.h:1005)
...
==15423==    by 0x4284B0F: TCint::Calc(char const*, TInterpreter::EErrorCode*) (TCint.cxx:655)
==15423==    by 0x1113149E: StBFChain::Init() (StBFChain.cxx:1039)

veprbl:
> 

https://github.com/star-bnl/star-sw/blob/fd34a521dbbe4d12632eec1a6d164b2231af7653/StarVMC/StarAgmlLib/AgPosition.cxx#L144-L146
https://github.com/star-bnl/star-sw/blob/fd34a521dbbe4d12632eec1a6d164b2231af7653/StarVMC/StarAgmlLib/AgPosition.cxx#L165-L168
https://github.com/star-bnl/star-sw/blob/fd34a521dbbe4d12632eec1a6d164b2231af7653/StarVMC/StarAgmlLib/AgPosition.cxx#L190-L198

PR #478, opened 2023-01-06
Synchronize dEdx calibration histograms with TFG, clean up tpcRDOMask…

fisyak:
…s.20071128.13225.C


plexoos:
Just a warning: This change enables the filling of a several 2D and 3D histograms which, in turn, may slow down the production jobs.

starsdong:
Yuri, the change in the dEdxY2Maker seems to be just moving a few macro definitions around, no logic or function change? Do you see any necessity for the change of dEdxY2Maker? For the removal of tpcRDOMask.XXX.C, I understand that this's now been handled in DB entries. So no problem. Thanks

fisyak:
Put in PR 529

PR #479, opened 2023-01-10
Bug fix for StFcsWaveformFitMaker and some minor additions

dkapukchyan:
This is a fix for a bug in StFcsWaveformFitMaker where the fit parameters where not being set correctly. The ability to recompute energy without fitting for the integral is added. Added some pedestal computing functions. Also, more doxygen style documentation.


StRoot/StFcsWaveformFitMaker/StFcsPulseAna.cxx
```diff
@@ -620,17 +620,22 @@ Double_t StFcsPulseAna::PulseFit(Int_t Start, Int_t End)
   else{ mFitSum = true; return mSumAdcFit; }//If fit failed return 0 don't do a histogram sum
 }
 
-void StFcsPulseAna::SetFitPars(TF1* func)
+void StFcsPulseAna::SetFitPars(TF1*& func, int start, int end)
```
plexoos, line 623: Why change the type of the first argument from TF1* to TF1*&?

akioogawa:
A lot more changes than a bug fix. But I believe you've already checked this produces good pi0 peaks and not loosing statistics. I approve.

akioogawa (approved):

dkapukchyan:
> A lot more changes than a bug fix. But I believe you've already checked this produces good pi0 peaks and not loosing statistics. I approve.

Yes, I forgot to mention that I got rid of some energy selects that are no longer relevant. I also added the pedestal computing functions and their corresponding energy selects for some methods; so now there are more energy selects so you can use my new methods with non-pedestal subtracted data.

StRoot/StFcsWaveformFitMaker/StFcsWaveformFitMaker.cxx
```diff
@@ -1558,27 +1708,32 @@ int StFcsWaveformFitMaker::PeakCompare(const PeakWindow& pwin1, const PeakWindow
...
+      if( !valididx.empty() && valididx.back()!=(i-1) ){ LOG_ERROR << "StFcsWaveformFitMaker::NPeaksPrePost: Found indices are not linear" << endm;  }
+      valididx.push_back(i);
       if( xmin > mPulseFit->GetPeak(i).mStartX ){ xmin = mPulseFit->GetPeak(i).mStartX; }
       if( xmax < mPulseFit->GetPeak(i).mEndX ){ xmax = mPulseFit->GetPeak(i).mEndX; }
     }
   }
-  if( !foundtrigidx ){ LOG_ERROR << "StFcsWaveformFitMaker::NPeakPre2Post1: Unable to find a matching triggered crossing possibly due to improper use of function" << endm; }
-  return npeaksxing;
+  //std::cout << "|fountrigidx:"<<foundtrigidx << "|trigidx:"<<trigidx << std::endl;
+  //for( unsigned int i=0; i<valididx.size(); ++i ){ std::cout << " - |valididx|i:" << i << "|at:"<< valididx.at(i) << std::endl; }
```
plexoos: Remove dead code

```suggestion
```

StRoot/StFcsWaveformFitMaker/StFcsPulseAna.cxx
```diff
@@ -640,25 +645,26 @@ void StFcsPulseAna::SetFitPars(TF1* func)
...
     func->SetParName(j,name);
-    para[j++] = GetPeak(i).mPeakY;
-    sprintf(name,"P%d_M",i);
+    para[j++] = GetPeak(start+i).mPeakY;
+    sprintf(name,"P%d_M",start+i);
     func->SetParName(j,name);
-    para[j++] = GetPeak(i).mPeakX;
-    sprintf(name,"P%d_S",i);
+    para[j++] = GetPeak(start+i).mPeakX;
+    sprintf(name,"P%d_S",start+i);
```
plexoos, line 657: This change introduced compiler warnings. Please address

```
2023-01-10T23:20:47.9029308Z #10 1302.4 .sl79_gcc11/OBJ/StRoot/StFcsWaveformFitMaker/StFcsPulseAna.cxx: In member function 'void StFcsPulseAna::SetFitPars(TF1*&, int, int)':
2023-01-10T23:20:47.9030257Z #10 1302.4 .sl79_gcc11/OBJ/StRoot/StFcsWaveformFitMaker/StFcsPulseAna.cxx:651:20: warning: '%d' directive writing between 1 and 10 bytes into a region of size 9 [-Wformat-overflow=]
2023-01-10T23:20:47.9030759Z #10 1302.4   651 |     sprintf(name,"P%d_A",start+i);
2023-01-10T23:20:47.9031024Z #10 1302.4       |                    ^~
2023-01-10T23:20:47.9031443Z #10 1302.4 .sl79_gcc11/OBJ/StRoot/StFcsWaveformFitMaker/StFcsPulseAna.cxx:651:18: note: directive argument in the range [0, 2147483647]
2023-01-10T23:20:47.9031866Z #10 1302.4   651 |     sprintf(name,"P%d_A",start+i);
2023-01-10T23:20:47.9032149Z #10 1302.4       |                  ^~~~~~~
2023-01-10T23:20:47.9032765Z #10 1302.4 .sl79_gcc11/OBJ/StRoot/StFcsWaveformFitMaker/StFcsPulseAna.cxx:651:12: note: 'sprintf' output between 5 and 14 bytes into a destination of size 10
2023-01-10T23:20:47.9033215Z #10 1302.4   651 |     sprintf(name,"P%d_A",start+i);
2023-01-10T23:20:47.9033503Z #10 1302.4       |     ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~
2023-01-10T23:20:47.9034178Z #10 1302.4 .sl79_gcc11/OBJ/StRoot/StFcsWaveformFitMaker/StFcsPulseAna.cxx:654:20: warning: '%d' directive writing between 1 and 10 bytes into a region of size 9 [-Wformat-overflow=]
2023-01-10T23:20:47.9034662Z #10 1302.4   654 |     sprintf(name,"P%d_M",start+i);
2023-01-10T23:20:47.9034935Z #10 1302.4       |                    ^~
2023-01-10T23:20:47.9035345Z #10 1302.4 .sl79_gcc11/OBJ/StRoot/StFcsWaveformFitMaker/StFcsPulseAna.cxx:654:18: note: directive argument in the range [0, 2147483647]
2023-01-10T23:20:47.9035751Z #10 1302.4   654 |     sprintf(name,"P%d_M",start+i);
2023-01-10T23:20:47.9036030Z #10 1302.4       |                  ^~~~~~~
2023-01-10T23:20:47.9036656Z #10 1302.4 .sl79_gcc11/OBJ/StRoot/StFcsWaveformFitMaker/StFcsPulseAna.cxx:654:12: note: 'sprintf' output between 5 and 14 bytes into a destination of size 10
2023-01-10T23:20:47.9037090Z #10 1302.4   654 |     sprintf(name,"P%d_M",start+i);
2023-01-10T23:20:47.9037373Z #10 1302.4       |     ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~
2023-01-10T23:20:47.9038379Z #10 1302.4 .sl79_gcc11/OBJ/StRoot/StFcsWaveformFitMaker/StFcsPulseAna.cxx:657:20: warning: '%d' directive writing between 1 and 10 bytes into a region of size 9 [-Wformat-overflow=]
2023-01-10T23:20:47.9038879Z #10 1302.4   657 |     sprintf(name,"P%d_S",start+i);
2023-01-10T23:20:47.9039140Z #10 1302.4       |                    ^~
2023-01-10T23:20:47.9039549Z #10 1302.4 .sl79_gcc11/OBJ/StRoot/StFcsWaveformFitMaker/StFcsPulseAna.cxx:657:18: note: directive argument in the range [0, 2147483647]
2023-01-10T23:20:47.9039968Z #10 1302.4   657 |     sprintf(name,"P%d_S",start+i);
2023-01-10T23:20:47.9040230Z #10 1302.4       |                  ^~~~~~~
2023-01-10T23:20:47.9040856Z #10 1302.4 .sl79_gcc11/OBJ/StRoot/StFcsWaveformFitMaker/StFcsPulseAna.cxx:657:12: note: 'sprintf' output between 5 and 14 bytes into a destination of size 10
2023-01-10T23:20:47.9041304Z #10 1302.4   657 |     sprintf(name,"P%d_S",start+i);
2023-01-10T23:20:47.9041591Z #10 1302.4       |     ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~
```

StRoot/StFcsWaveformFitMaker/StFcsWaveformFitMaker.cxx
```diff
@@ -459,40 +470,33 @@ int StFcsWaveformFitMaker::Make() {
...
-	    ped = float(p)/(mPedMax-mPedMin+1.0);
-	    mDb->setPedestal(hit->ehp(), hit->ns(), hit->dep(), hit->channel(), ped);
+	    AnaPed( hits[i], res[6], res[7] );
+	    mDb->setPedestal(hits[i]->ehp(), hits[i]->ns(), hits[i]->dep(), hits[i]->channel(), res[6] );
 	    if(GetDebug()>1){
 	      char name[100];
-	      mDb->getName(hit->ehp(), hit->ns(), hit->dep(), hit->channel(),name);
-	      printf("%s Pedestal=%d/(%d-%d+1)=%8.2f\n",name,p,mPedMax,mPedMin,ped);
+	      mDb->getName(hits[i]->ehp(), hits[i]->ns(), hits[i]->dep(), hits[i]->channel(),name);
+	      printf("%s Pedestal (%d-%d+1)=%8.2f\n",name,mPedMax,mPedMin,res[6] );
```
plexoos: In STAR makers you should use LOG_DEBUG instead of printf

StRoot/StFcsWaveformFitMaker/StFcsWaveformFitMaker.cxx
```diff
@@ -620,6 +615,45 @@ TGraphAsymmErrors* StFcsWaveformFitMaker::makeTGraphAsymmErrors(StFcsHit* hit){
...
+  for(int i=0; i<n; i++){
+    int tb1=t[i];
+    if(mPedMin<=tb1 && tb1<=mPedMax){    
+      p+=a[i];
+      sumsq += a[i]*a[i];
+    }
+  }
+  ped = float(p)/(mPedMax-mPedMin+1.0);
+  pedstd = sqrt( ( sumsq-((double(p)*double(p))/(mPedMax-mPedMin+1.0)) )/(mPedMax-mPedMin) );//Variance/StdDev using naive algorithm from https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance
+  return ped;
```
plexoos, line 635: Why return the same value if you modify the function argument?

StRoot/StFcsWaveformFitMaker/StFcsWaveformFitMaker.cxx
```diff
@@ -620,6 +615,45 @@ TGraphAsymmErrors* StFcsWaveformFitMaker::makeTGraphAsymmErrors(StFcsHit* hit){
...
+    int tb=hit->timebin(i);
+    if(mPedMin<=tb && tb<=mPedMax ){
+      int adc = hit->adc(i);
+      p+=adc;
+      sumsq += adc*adc;
+    }
+    if(tb>=mPedMax) break;
+  }
+  ped = float(p)/(mPedMax-mPedMin+1.0);
+  pedstd = sqrt( ( sumsq-((double(p)*double(p))/(mPedMax-mPedMin+1.0)) )/(mPedMax-mPedMin) );//Variance/StdDev using naive algorithm from https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance
```
plexoos: You don't need to cast `p` to double. It will be done automatically using the C++ rules for implicit type casting.

StRoot/StFcsWaveformFitMaker/StFcsWaveformFitMaker.cxx
```diff
@@ -620,6 +615,45 @@ TGraphAsymmErrors* StFcsWaveformFitMaker::makeTGraphAsymmErrors(StFcsHit* hit){
...
+  for(int i=0; i<n; i++){
+    int tb=hit->timebin(i);
+    if(mPedMin<=tb && tb<=mPedMax ){
+      int adc = hit->adc(i);
+      p+=adc;
+      sumsq += adc*adc;
+    }
+    if(tb>=mPedMax) break;
+  }
+  ped = float(p)/(mPedMax-mPedMin+1.0);
```
plexoos: You don't need to cast `p` to `float`. Remove, so it'd be easier to read the code. Why do you do the calculations in single precision rather than double?

plexoos (commented):
So which changes exactly fix the bug? Does it affect any of the past releases used for real data processing?

StRoot/StFcsWaveformFitMaker/StFcsWaveformFitMaker.cxx
dkapukchyan: The reason is that ped and pedstd get stored in the array "res" which is of type float (found in the function Make). So I chose float to match the array type. If I change it to double the whole array would have to be changed to double so would any method that uses the "res" array, so I thought it would be easier to preserve the type. Would you prefer we change everything to double?

StRoot/StFcsWaveformFitMaker/StFcsWaveformFitMaker.cxx
dkapukchyan: This has to do with the way the code was setup before. I didn't want to overhaul too many things and preserve the overall logic in Make. I would have to talk to @akioogawa before making such changes because it could break the way he has things set up and will require more time and testing to ensure that everything still works as intended.

StRoot/StFcsWaveformFitMaker/StFcsPulseAna.cxx
dkapukchyan: This argument was changed because in line 639 of that file there is a check if(func==0){create a new function}. Without the reference the pointer to this new function will be lost since SetFitPars() does not return the TF1* pointer that if may have created. This was certainly bad and now fixed but not the main issue.

dkapukchyan:
> So which changes exactly fix the bug? Does it affect any of the past releases used for real data processing?

The bug fix is in the StFcsPulseAna::SetFitPars() and the (newly named) StFcsWaveformFitMaker::NPeaksPrePost() function. Any releases or data processed using the StFcsWaveformFitMaker after pull request 401 will have the wrong calorimter energy if the default option for energy select was used. I am not sure which past releases this constitutes.

The problem was that the PulseFit2() algorithm is supposed to find and fit only to a select number of peaks (npeaks). The number of peaks it is supposed to fit is given by NPeaksPrePost(). Also, since the peaks are stored in a vector if the peak for the triggered crossing was now the second peak instead of the 4th this new index for the triggered crossing peak needed to be returned by NPeaksPrePost() so you pick up the correct parameter from the TF1.

However, The function SetFitPars assumes that all peaks will need to be fit up to npeaks. This means if you have 5 peaks and only want to fit the 3rd and 4th peak; then npeaks=2 so SetFitPars would actually set the parameters for the 1st and 2nd peaks NOT the 3rd and 4th. This is the bug.

The fix was to change SetFitPars to accept a starting index and ending index for which peaks to set the parameters for. Also, NPeaksPrePost() needed to be changed to return a vector of indexes for peaks that need to be fit. The trigfitidx now tells you the index in the vector of indexes where the triggered crossing was. This was done in this way because trigfitidx will match the index of the parameter space of the TF1 where to find the needed value.

Most of the other code was added to find this bug.

StRoot/StFcsWaveformFitMaker/StFcsWaveformFitMaker.cxx
dkapukchyan: done. Is is this because of the 1.0 in the denominator?

StRoot/StFcsWaveformFitMaker/StFcsWaveformFitMaker.cxx
dkapukchyan: done

StRoot/StFcsWaveformFitMaker/StFcsPulseAna.cxx
dkapukchyan: That's odd because I don't remember seeing this warning. I increased size of the char array and I don't see such warnings

StRoot/StFcsWaveformFitMaker/StFcsWaveformFitMaker.cxx
dkapukchyan: done

StRoot/StFcsWaveformFitMaker/StFcsPulseAna.cxx
plexoos: Fair enough. I haven't noticed that you actually create a new function and return it in the argument.

StRoot/StFcsWaveformFitMaker/StFcsPulseAna.cxx
plexoos: These are from GCC 11. You can see the build log if you go to the Checks tab
I see that these warnings are gone now. Thanks

StRoot/StFcsWaveformFitMaker/StFcsWaveformFitMaker.cxx
```diff
@@ -1066,27 +1105,94 @@ void StFcsWaveformFitMaker::drawCh(UInt_t detid, UInt_t ch) const
...
-      sumsq += a[i]*a[i];
-    }
-  }
-  float ped = float(p)/(mPedMax-mPedMin+1.0);
-  res[6] = ped;
-  res[7] = sqrt( ( sumsq-((double(p)*double(p))/(mPedMax-mPedMin+1.0)) )/(mPedMax-mPedMin) );//Variance/StdDev using naive algorithm from https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance
-  printf("Pedestal=%6.2f/(%6.2f-%6.2f+1)=%6.2f +- %6.2f\n",float(p),float(mPedMax),float(mPedMin),ped,res[7]);
-  return gausFit(g, res, func, ped);
+  AnaPed( g, res[6], res[7] );
+  printf("Pedestal (%6.2f-%6.2f+1)=%6.2f +- %6.2f\n",float(mPedMax),float(mPedMin),res[6],res[7]);
```
plexoos: Another printf should be changed to LOG_INFO. It is kind of important if this maker is to be used in production

StRoot/StFcsWaveformFitMaker/StFcsWaveformFitMaker.cxx
plexoos: yes, all values are converted to double in arithmetic expressions

plexoos (approved):

StRoot/StFcsWaveformFitMaker/StFcsWaveformFitMaker.cxx
dkapukchyan: Changed to LOG_DEBUG

PR #487, opened 2023-01-30
[root6] Fix call to memset in TpcAvgCurrent.C

plexoos:
```
StWarning: StDbSql::QueryDb(table,time) line=439 TpcAvgCurrent  has No data for query
LoadTable: .L /star-sw/StarDb/Calibrations/tpc/TpcAvgCurrent.C
In file included from input_line_983:1:
/star-sw/StarDb/Calibrations/tpc/TpcAvgCurrent.C:6:3: error: no matching function for call to 'memset'
  memset(row, 0, sizeof(TpcAvgCurrent_st));
  ^~~~~~
/usr/include/string.h:62:14: note: candidate function not viable: no known conversion from 'TpcAvgCurrent_st' to 'void *' for 1st argument; take the address of the argument with &
extern void *memset (void *__s, int __c, size_t __n) __THROW __nonnull ((1));
             ^
root4star: .sl79_gcc485/OBJ/StRoot/St_db_Maker/St_db_Maker.cxx:932: virtual TDataSet* St_db_Maker::LoadTable(TDataSet*): Assertion `!ee' failed.
Aborted
```

### DIFF
--- a/StarDb/Calibrations/tpc/TpcAvgCurrent.C
+++ b/StarDb/Calibrations/tpc/TpcAvgCurrent.C
@@ -1,7 +1,9 @@
+#include <cstring>
+
 TDataSet *CreateTable() {
   if (!gROOT->GetClass("St_TpcAvgCurrent")) return 0;
   TpcAvgCurrent_st row;
-  memset(row, 0, sizeof(TpcAvgCurrent_st));
+  std::memset(&row, 0, sizeof(TpcAvgCurrent_st));
   St_TpcAvgCurrent *tableSet = new St_TpcAvgCurrent("TpcAvgCurrent",1);
   tableSet->AddAt(&row);
   return (TDataSet *)tableSet;


StarDb/Calibrations/tpc/TpcAvgCurrent.C
```diff
@@ -1,7 +1,9 @@
+#include <cstring>
+
 TDataSet *CreateTable() {
   if (!gROOT->GetClass("St_TpcAvgCurrent")) return 0;
   TpcAvgCurrent_st row;
-  memset(row, 0, sizeof(TpcAvgCurrent_st));
+  std::memset(&row, 0, sizeof(TpcAvgCurrent_st));
```
klendathu2k, line 6: I would not be certain that root5 / CINT would work with std::memset vs memset... but seems to be passing tests.

klendathu2k (approved):

StarDb/Calibrations/tpc/TpcAvgCurrent.C
plexoos: No complaints from root5 AFAICT

genevb:
I can't even view this directory on github because of all the old files that came and went:
https://github.com/star-bnl/star-sw/tree/main/StarDb/Calibrations/tpc
message from github (not my browser): "This page is taking too long to load."
:-(

klendathu2k:
I think github must be having a problem at the moment.  I was able to 
review this earlier... but now I see the unfriendly unicorn.

iraklic:
I have a question about this, how did it ever work if it needed `&row` and had `row` for many years till now?

genevb:
> On Jan 30, 2023, at 11:54 AM, Irakli Chakaberia ***@***.***> wrote:
> I have a question about this, how did it ever work if it needed &raw and had raw for many years till now?

Yeah, I was wondering the same, which is why I was trying to look at the history. Indeed, it's been wrong since the initial version:
https://www.star.bnl.gov/cgi-bin/protected/viewvc.cgi/cvsroot/StarDb/Calibrations/tpc/TpcAvgCurrent.C?revision=1.1&view=markup
Hm.

BTW, the old CVS browser also takes a little while to view that directory too, but eventually comes through.
https://www.star.bnl.gov/cgi-bin/protected/viewvc.cgi/cvsroot/StarDb/Calibrations/tpc/

-Gene

klendathu2k:
ROOT/CINT is not a C++ compiler.  It is a C++ flavored interpreter,
laced with several usability hacks... one of which blurs the distinction
between objects and pointers.

This code runs happily under ROOT5/CINT...

struct A_t { int a; int b; int c; };

void testdum() {
   a.a = 1;  a.b = 2;  a.c = 3;
   std::cout << "before questionable memset ... " << std::endl << a.a << 
" " << a.b << " " << a.c << std::endl;
   memset( a, 0, sizeof(A_t) );
   std::cout << " after questionable memset ... " << std::endl << a.a << 
" " << a.b << " " << a.c << std::endl;
};

And produces the output
before questionable memset ...
1 2 3
  after questionable memset ...
0 0 0

So... the original code is not correct C++.  But under CINT it should 
work as the author intended.

Jason

plexoos:
The old code worked because in reality a pointer was passed to memset regardless of what the source code said. This is the result of CINT not following the C++ standard but with ROOT6 cling we can detect such leeways and fix them.

@iraklic you can still approve this PR if you feel like it. I was not sure it was worth waiting

iraklic:
Thank you. That is kind of what I suspected, but I was thinking since we are compiling the code it would pass through the compilator C++ rules and not CINT, but I guess there are details that are not as clear to me in this regard.

plexoos:
This particular file (actually all .C files in StarDB/Calibrations/) is "interpreted" by CINT/Cling

iraklic:
> This particular file (actually all .C files in StarDB/Calibrations/) is "interpreted" by CINT/Cling

Oh, right, that is the answer that puts things in place. Yes I forgot for a moment about our way with some cint files. Thanks.

PR #488, opened 2023-01-31
bfc.C: Call Load() function unconditionally

plexoos:
See discussion in #452 https://github.com/star-bnl/star-sw/pull/452#pullrequestreview-1203814837


plexoos:
@genevb Any comments from you? Thank you

genevb (approved):
> @genevb Any comments from you? Thank you

I doubt I've thought of anything that you haven't already, @plexoos . If there's no concern about twice-loading (either that it's certain that it won't happen, or that happening is harmless), then I have no objection to loading unconditionally.

PR #481, opened 2023-01-24
Code update for Run18 embedding

zhux97:
Patches needed for Run18 embedding are assembled in this PR. These patches are:
* RDO map is updated and iTPC RDO is added for sector 20 in Run18.
* TpcRS parameters for the second half of Run18 27 GeV dataset is added.
* Offline cluster finder (RTS and StTpcRTSHitMaker) is updated with the correct gain interface.
* A newer StTpcRSMaker is included to correct the 1st pad row.

PR #490, opened 2023-02-01
Backport CI machinery from SL23a_0

plexoos:
```
git diff SL19b_3..SL23a_0 -- docker/ .github/workflows/ Dockerfile tests/ asps/rexe/Cons* asps/rexe/*LinkDef*  mgr/Cons* mgr/config/ StRoot/macros/rootlogo*.C StRoot/macros/.rootrc > patch.diff
git am < patch.diff
```


plexoos:
This PR seemed to pass the tests as expected, although one test job 101 crashed on the first attempt and succeeded on the second try. I don't like to see such instabilities but there is no time to investigate right now so, I am going to merge and restart #481

PR #499, opened 2023-03-07
[starsim] Fix logic handling multiple hit banks in the stgc

klendathu2k:
g2t_stg.F copies hits from the three sensitive volumes into the single g2t_stg_hit table.
When no hits are registered in a single sensitive volume, the loop _should_ proceed to
query the next SV for hits.  Instead, the code branched to the return statement, resulting
in no hits being saved for the remaining SVs.  

This PR fixes the issue.


fvidebaek (approved):
Changes small and appropriate to solve issue

jdbrice (approved):
this fixes the missing hits in events with few tracks

plexoos (approved):
LGTM

fvidebaek (approved):

PR #500, opened 2023-03-15
RunXXI OO200GeV dEdx calibration

fisyak:


StarDb/RunLog/onl/triggerID.C
```diff
@@ -9,6 +9,14 @@ triggerID_st row;
 St_triggerID *tableSet = new St_triggerID("triggerID",1);
 //
 memset(&row,0,tableSet->GetRowSize());
+    row.runNumber	 =          0; // run number  ;
```
genevb, line 12: The `memset()` above this already zeroes all the members of the `row` struct. Is there a benefit from adding these? Maybe it simplifies trying non-zero values in tests to have these lines already written?

starsdong:
Dear Colleagues, We are urgently in need of this PR for the upcoming OO production (analysis aimed for QM). Could you please help take a review of the code change and comment/approve at your best? I also added Xianglei/Jason in. From today's discussion, I understand we can use timestamp control to avoid conflict with the embedding/simulation. There are a few changes on the tpcRDOMasks for previous years. Xianglei/Jason, if you can take a review and make sure this won't introduce issues with the embedding/simulation. I hope we can get this in by the end of this week so we can proceed with production soon.

iraklic (approved):

iraklic:
Can we have MDC change related improvement plot slides attached to this PR like we do for parameter tuning commits

starsdong:
Irakli and all, Yuri gave a report at the S&C meeting yesterday. Please find the presentation below
[https://drupal.star.bnl.gov/STAR/event/2023/03/15/SC-management-meeting/Run21-OO-calibration-update-dEdx-BTOF-etc]

dmarkh (approved):
Looks good to me

zhux97 (approved):
The RDOMask files seem to be used for Run20+ embedding only,  which we have not started yet. I will test them once these embedding production starts, for now I approve the changes. For current embedding productions for Run19-, if we don't propagate these files back to the old libraries, they should not be affected.

klendathu2k (approved):
The tpcRDOMasks.y*.C files were removed and replaced with symlinks to tpcRDOMasks.C.  No functional change there.

Addition of the tpcRDOMasks at various timestamps is, as Xianglei mentioned, something which should / will be tested.   I note that I didn't see any database entries in that time frame.

In short, looks good.

starsdong:
Thank you all for the quick comments/response. I think it is ready to merge.

genevb:
I thought we agreed at the S&C meeting that the numerous timestamped tpcRDOMasks CINT files would be entered into the database instead of committed to the code library?

genevb:
This is what Xin wrote in the S&C meeting minutes:

> All calibration par are in DB. Code change in PR#500 (urgently need for review)
> updated tpcRDOMasks tables
> Yuri/Xianglei, use timestamp control to avoid conflict.

However, there is **_NO_** timestamp control when CINT files are added to the code library. This is a point that has been emphasized over and over when timestamped CINT files are committed under `StarDb/`.

genevb:
An observation, but likely not a bug:
Nightly tests which are pure simulation showed a different number of reconstructed ("loaded") TPC hits after this PR was merged. Embedding and real data nightly tests show no such changes. I infer this to be due to the modified TPC RDO masks for simulation timestamps that were introduced with this PR, but I am not 100% certain of this interpretation.

PR #501, opened 2023-03-17
Update EPD hit IO classes to include the DEP information

plexoos:


klendathu2k (changes requested):
I don't see any issues with the code, save for it breaking StEpdHitMaker.  Please revise StEpdHitMaker (and any other dependencies) to reflect the changes made to the constructor, and submit as part of this PR.

plexoos:
> I don't see any issues with the code,

Do you know if the automatic ROOT schema evolution can handle a renamed data member?

klendathu2k:
> 

Now that you mention it... I'm not certain.  So probably better to leave the name of the pre-existing data members in tact.  Otherwise need to test.

plexoos:
I think so too, it is safer to keep the old name. I did a test once removing and adding members and it worked well but renaming may need some magic with pragma in LinkDef.h

genevb:
> Do you know if the automatic ROOT schema evolution can handle a renamed data member?

Good catch! I think if you write your own Streamer(), there is probably a way to do this, but I believe that incrementing the ClassDef otherwise makes it think a member with one name has been dropped and a member with a different name has been added and should be filled with default zeroes when reading a file with an old version.

akioogawa:
At few places in comments, it says "DEP data from TriggerData". DEP data is not in Trigger data but in FCS data.

Femtoscopist:
Thanks, Akio.  You did provide Joey the methods to access DEP data. I have added a class to StEpd directory to take muDSTs and recalibrate using the db. This is necessary since sometimes the db gains are incorrect at the time of production. Ding Chen is checking it and will then submit.

My plan then is to add the DEP stuff.  I should still do this?



- sent from a phone - please excuse typos!
—
Michael Lisa
Professor of Physics
The Ohio State University

PR #513, opened 2023-03-26
Added centrality procedure instructions

zsweger:
Added centrality procedure instructions


plexoos:
Can we avoid adding binary files in the repository?

zsweger:
Okay we can remove those

plexoos:
Ah... we let those root files slip in #340. Since it is still the HEAD of the main branch, I think it is ok if we force remove them.
However, my comment was about the pdf file. How do you think other users can update it when needed? Can you use the .md or .rst formats instead?

zsweger:
Okay I've added a markdown version of the instructions and deleted the pdf!

plexoos (approved):

PR #519, opened 2023-04-02
RHICf Sub-reconstruction tool (StRHICfPID)

ggfdsa10:
RHICf Particle Identification reconstruction tool
this used in StRHICfPointMaker


StRoot/StRHICfUtil/StRHICfPID.cxx
```diff
@@ -0,0 +1,106 @@
...
+void StRHICfPID::init()
+{
+  initChecker();
+  for(int it=0; it<kRHICfNtower; it++){
+    for(int ip=0; ip<kRHICfNplate; ip++){mPlateE[it][ip] = 0.;}
+    mPID[it] = -1;
+    mPlateSumE[it] = 0.;
+    mL20[it] = 0.;
+    mL90[it] = 0.;
+  }
```
plexoos: Unless there is a really good reason to initialize the c-arrays this way, I suggest you use `std::fill()` in C++ code.

StRoot/StRHICfUtil/StRHICfPID.h
```diff
@@ -0,0 +1,40 @@
...
+    float checkStep(int layer);
+    float layerSumEnergy(int tower, int layer);
+    float findRadiationLength(int tower, float ratio);
+    float calculateEquation(int tower, int layer, float sumE);
+
+    int mPID[kRHICfNtower];
+    float mPlateE[kRHICfNtower][kRHICfNplate];
+    float mPlateSumE[kRHICfNtower];
+    float mL20[kRHICfNtower];
+    float mL90[kRHICfNtower];
```
plexoos, line 31: This is not an IO class AFAICT. Why define these variables with single precision?

StRoot/StRHICfUtil/StRHICfPID.cxx
```diff
@@ -0,0 +1,106 @@
+#include "StRHICfPID.h"
+
+StRHICfPID::StRHICfPID()
+{
+}
+
+StRHICfPID::~StRHICfPID()
+{
+}
```
plexoos: Not needed.

StRoot/StRHICfUtil/StRHICfPID.cxx
```diff
@@ -0,0 +1,106 @@
...
+float StRHICfPID::getL20(int tower){return mL20[tower];}
+
+float StRHICfPID::getL90(int tower){return mL90[tower];}
+
+bool StRHICfPID::calculate()
+{
+  if(!mPlateEIs){return kRHICfFatal;}
+
+  for(int it=0; it<kRHICfNtower; it++){
+    for(int ip=0; ip<mPlateIdxNum; ip++){mPlateSumE[it] += layerSumEnergy(it, ip);}
```
plexoos, line 29: See if you can use `std::accumulate()` here and in the next lines. It is likely to be more efficient than adding elements in a loop with if-statements...

StRoot/StRHICfUtil/StRHICfPID.cxx
```diff
@@ -0,0 +1,106 @@
+#include "StRHICfPID.h"
+
+StRHICfPID::StRHICfPID()
+{
+}
```
plexoos: If you choose not to initialize the member arrays this constructor is automatically generated for you anyway.  Maybe you want to call `init()` from this constructor, otherwise the C-arrays will contain random values.

StRoot/StRHICfUtil/StRHICfPID.h
ggfdsa10: Hi Dmitri,

Thank you for your comments. 
This class is a reconstructor for PID and not an IO class. 
RHICf saving classes, such as StRHICfHit and StRHICfPoint, include mostly single-precision (float) variables. 
StRHICfPID also receives input data as mPlateE from the StRHICfHit class using the setPlateEnergy() function. 
Therefore, I didn’t see the need to change it to double precision.

However, I think calculation functions such as calculateEquation() could be changed to double precision. 
Because that function needs to be more precise.
If you think it would be better to change these variables to double precision, I will make the changes along with others.

Thank you

StRoot/StRHICfUtil/StRHICfPID.cxx
ggfdsa10: This line of code can be replaced with the following:


```
mPlateSumE[it] += (2*std::accumulate(mPlateE[it][0], mPlateE[it][9], 0));
mPlateSumE[it] += mPlateE[it][10];
mPlateSumE[it] += (2*mPlateE[it][11]);
mPlateSumE[it] += (4*std::accumulate(mPlateE[it][12], mPlateE[it][14], 0));
mPlateSumE[it] += (2*mPlateE[it][15]);
```
However, I’m not sure which version is better. 
In my opinion, the original code lines are better than the replacement. 
I think the original version seems more intuitive and simpler.

ggfdsa10:
All of the input data-checking functions for RHICf come from the StRHICfFunction class. 
However, when someone uses a merged RHICf tool, those functions or codes will always be frozen. 

Furthermore, all input data checkers simply ensure that the total number of input data is correct. For example, they check that the setPlateEnergy() function has been called as many times as there are plate energy arrays.

For this reason, I don’t think these checkers need to be included in all of the RHICf tools, so I plan to remove them from all RHICf tools soon.
Commit[d92b32d](https://github.com/star-bnl/star-sw/pull/519/commits/d92b32d6112640a1665908e4c15cf5b036f4c908) was made for the same reason.

plexoos (approved):

starsdong:
I think this one is ready to merge. Dmitri, please go head. Thanks

PR #492, opened 2023-02-02
Additon of StFwdTrack to StEvent, updates to StFwdTrackMaker to accomodate

jdbrice:
This PR replaces #397 (it was easier to start fresh)
Major updates
- New StEvent StFwdTrack model that does not depend on GenFit
- Updates to FwdTracker:
  - Add mapped BField to speed up tracking in forward region
  - Add Generalized broken lines refit for alignment studies
  - Add option for tracking in zero field running
  - Updates for new sTGC geometry 
  - Write new StFwdTracks to StEvent
  - Macros and READMEs for testing and running on simulation / DAQ / StEvent files

PR #522, opened 2023-04-18
fix: remove ClassDef from StEpdGeom

plexoos:
This type is neither an StMaker nor an IO class so, using ClassDef is not critical. On the contrary, it causes an error due to missing virtual destructor when included in compiled code. E.g. see https://github.com/star-bnl/star-sw/pull/492


plexoos:
This fix seemed to fix the compile error in #492

PR #527, opened 2023-04-22
New RHICf sub-reconstruction tool (StRHICfRecoEnergy)

ggfdsa10:
RHICf hit energy reconstruction tool
this used in StRHICfPointMaker

### DIFF
--- a/StRoot/StRHICfUtil/StRHICfRecoEnergy.cxx
+++ b/StRoot/StRHICfUtil/StRHICfRecoEnergy.cxx
@@ -1,0 +1,323 @@
+#include "StRHICfRecoEnergy.h"
+
+StRHICfRecoEnergy::StRHICfRecoEnergy()
+{
+  init();
+}
+
+StRHICfRecoEnergy::~StRHICfRecoEnergy()
+{
+  for(int it=0; it<kRHICfNtower; it++){
+    delete mLeakOutTableNeutron[it];
+    mLeakOutTableNeutron[it] = nullptr;
+    for(int ip=0; ip<kRHICfNplate; ip++){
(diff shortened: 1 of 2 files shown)


starsdong:
Seunghwan, is it possible that we may have a test and maybe include it in the CI to see the code works? Thanks

plexoos:
> is it possible that we may have a test and maybe include it in the CI to see the code works?

An excellent suggestion! 😍 Ideally, should be applied to all incoming changes

ggfdsa10:
> Seunghwan, is it possible that we may have a test and maybe include it in the CI to see the code works? Thanks

Yes, I agree too!
A few sub-Reco. tool for RHICf, StRHICfRecoEnergy, and StRHICfRecoPos are dependent on the StRHICfPointMaker.
But StRHICfPointMaker is not included in this GitHub yet. 

I think it should be better to request the PRs for this maker first, before uploading the CI test for these codes.
so I would like to suggest several steps:
1. PR for StRHICfRecoPos (already done)
2. PR for StRHICfRecoEnergy (in this PR)
3. PR for StRHICfHitMaker 
4. PR for StRHICfPointMaker 
5. Request CI test code for RHICf Reco. using BFC options (this test similar to RHICf raw data CI test code)
6. PR for RHICf PicoDst, etc... 

or, is there any way to CI test for this without some BFC level? 
if you can Yes, please let me know 

Thank you! 
Seunghwan

starsdong:
Seunghwan, thanks for the clarification. I think this is fine. The codes compile so I don't have a concern on this PR.

starsdong:
Dmitri, I don't see much of issue. If you are OK, let us also merge this PR then. Thanks

PR #531, opened 2023-04-26
Edited mistake in params file

zsweger:
There were two edits made to the params file. One was to change the refMultCorr value below which the trigger efficiency gets applied from 100 to 150. The second was to update one of the trigger efficiency parameters to 0 to indicate there is no Vz-dependent trigger efficiency applied.


plexoos:
Hi Zachary, do we expect this change to affect any past production in case we need to reproduce results?

starsdong:
Zach may confirm, but AFAIK, StRefMultCorr is only used at the analysis level, not in production. It is important that we should update PWGs about this change so the analysers should update their analysis.

marrbnl (approved):

nigmatkulov (approved):
Looks good!

PR #534, opened 2023-05-01
Star simulations 2023 ideal timestamp

klendathu2k:
STAR simulation (ideal) timeline moved to 04/10/2023.  

Production tag y2023a defined.  Will be frozen closer to offline data productions.


plexoos (approved):

genevb (approved):
I'm fine with the updated timestamp.

PR #535, opened 2023-05-02
STARSIM / user defined cylindrical decay region

klendathu2k:
Adds a package which enables users to define a cylindrical decay region for a list of up to 10 particles.

The package makes the assumption that particles are sufficiently energetic as to not be significantly deflected by the STAR magnetic field.


plexoos (approved):

plexoos:
Jason, please confirm it is ready and we'll merge it

PR #537, opened 2023-05-08
hotfix for broken event.root files in DEV with StFwdTracks

jdbrice:
I have been unable to read back event.root files created with FWD track info in DEV. The BFC ends with a set fault. 

I found that adding missing StObject and ClassDef to StFwdTrack causes the chain to run without segfault and the data to be read correctly. 

However, I am still getting messages like:
StIOMaker:ERROR - TBufferFile::CheckByteCount : object of class StEventClusteringHints read too few bytes: 1401 instead of 3177

So far I have not been able to pin down the source of this error. I tried using "//!" to remove various members from the StFwdTrack but nothing resolved the above ERROR message. I wanted to put this PR in so that the event.root files are at least readable without causing a segmentation fault.

### DIFF
--- a/StRoot/StEvent/StFwdTrack.h
+++ b/StRoot/StEvent/StFwdTrack.h
@@ -26,7 +26,7 @@
 class StFcsCluster;
 
 
-struct StFwdTrackProjection {
+struct StFwdTrackProjection : public StObject {
     StFwdTrackProjection() {}
     StFwdTrackProjection ( const StFwdTrackProjection & other) {
         mXYZ = other.mXYZ;
@@ -70,9 +70,11 @@ struct StFwdTrackProjection {
     float dz(){
         return sqrt( mCov[8] );
     }
+
+    ClassDef(StFwdTrackProjection, 1)
 };
 
-struct StFwdTrackSeedPoint {
+struct StFwdTrackSeedPoint : public StObject {
     StFwdTrackSeedPoint() {}
     StFwdTrackSeedPoint(    StThreeVectorD xyz, 
                             short sec, 
@@ -88,6 +90,8 @@ struct StFwdTrackSeedPoint {
     unsigned short mTrackId;
     short mSector;
     float mCov[9];
+    
+    ClassDef(StFwdTrackSeedPoint, 1)
 };
 
 class StFwdTrack : public StObject {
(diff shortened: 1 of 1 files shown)


jdbrice:
to reproduce:
```sh
root4star -b -l
.L bfc.C
bfc( 10, "in, y2023, useXgeom, AgML, db, StEvent, MakeEvent", "/gpfs01/star/pwg_tasks/FwdCalib//PROD/zfa/out3/st_physics_23072003_raw_1000002.event.root" )
```

this will segfault. But after recompiling with the included changes, it will run fine but emit the error mentioned previously.

plexoos:
Has the forward tracking library been made available at SDCC?

jdbrice:
Not that I know of, @genevb can you confirm or not?

jdbrice:
@plexoos, do we have a test that writes event.root to disk and then attempts to read back from disk? I ask because with the FWD detectors and tracking we have now had a few cases where the in-memory StEvent data structure works fine, but the I/O fails and as far as I can tell this was caught "manually" both times. May be worth trying to test for this failure case.

plexoos (approved):

plexoos:
> do we have a test that writes event.root to disk and then attempts to read back from disk?

Not that I am aware of

fisyak:
root.exe 'bfc.C(1,"in,analysis,nodefault","129/24129041/hlt_24129041_11_02_000.event.root")'


                                                              Yuri Fisyak

STAR                                           Phone: +1 631 344 3913
Brookhaven National Laboratory  Fax:    +1 631 344 4206
510A/1-161
http://www.star.bnl.gov/~fisyak       E-mail: ***@***.******@***.***>



From: Dmitri Smirnov ***@***.***>
Reply-To: star-bnl/star-sw ***@***.***>
Date: Thursday, May 18, 2023 at 6:24 PM
Cc: Subscribed ***@***.***>
Subject: Re: [star-bnl/star-sw] hotfix for broken event.root files in DEV with StFwdTracks (PR #537)


do we have a test that writes event.root to disk and then attempts to read back from disk?

Not that I am aware of

klendathu2k (approved):
Looks reasonable.  Will investigate the clustering hints error once this is merged.

PR #543, opened 2023-05-18
fix: load libiconv module into envs

plexoos:
While linking StDbLib the linker cannot find libiconv, an indirect dependency via libxml2, unless it searches paths in LD_LIBRARY_PATH


genevb (approved):
I confirm that this allows StDb.so to be loaded in root4star batch mode (root4star -b), but it seems odd this wasn't needed for non-batch mode.

PR #2, opened 2021-05-27
Initial revision of .github/CODEOWNERS

plexoos:

PR #544, opened 2023-05-19
RHICf Run and Event number added in StRHICfCollection

ggfdsa10:
PR 1. StRHICfCollection for RHICf Run and Event Number

I will request 4 PRs for adding the RHICf run and event number, before the new STAR release 
1) this PR
2) StRHICfDbMaker
3) StMuDstMaker/StRHICfRawHit and StRHICfUtil
4) StRHICfRawHitMaker

I believe reviewing these PRs is very simple and clear.
Please review and approve this 

Thank you, 
Seunghwan

### DIFF
--- a/StRoot/StEvent/StRHICfCollection.cxx
+++ b/StRoot/StEvent/StRHICfCollection.cxx
@@ -30,6 +30,8 @@ void StRHICfCollection::clear()
   mRHICfRawHitColl = new StRHICfRawHit();
   mRHICfHitColl = 0;
 
+  mRHICfRunNumber = 0;
+  mRHICfEventNumber = 0;
   mBunchNumber = 0;
   mRHICfTrigger = 0;
   mRunTime[0] = 0;
@@ -50,6 +52,8 @@ const std::vector<StRHICfPoint*>& StRHICfCollection::pointCollection() const {re
 void StRHICfCollection::isAllSave(){mRHICfHitColl = new StRHICfHit();}
 
 //=========== Set ===========//
+void StRHICfCollection::setRHICfRunNumber(UInt_t run){mRHICfRunNumber = run;}
+void StRHICfCollection::setRHICfEventNumber(UInt_t event){mRHICfEventNumber = event;}
 void StRHICfCollection::setBunchNumber(UInt_t bunch){mBunchNumber = bunch;}
 void StRHICfCollection::setRunType(UInt_t type){mRunType = type;}
 void StRHICfCollection::setTriggerNumber(UInt_t trigger){mRHICfTrigger = trigger;}
@@ -58,6 +62,8 @@ void StRHICfCollection::setRunTRGM(UInt_t trgm){mRunTRGM = trgm;}
 
 //=========== Get ===========//
 UInt_t StRHICfCollection::numberOfPoints() const {return mRHICfPointColl.size();}
+UInt_t StRHICfCollection::getRHICfRunNumber(){return mRHICfRunNumber;}
+UInt_t StRHICfCollection::getRHICfEventNumber(){return mRHICfEventNumber;}
 UInt_t StRHICfCollection::getBunchNumber(){return mBunchNumber;}
 UInt_t StRHICfCollection::getRunType(){return mRunType;}
 UInt_t StRHICfCollection::getTriggerNumber(){return mRHICfTrigger;}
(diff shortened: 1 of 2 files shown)


plexoos (approved):
LGTM

StRoot/StEvent/StRHICfCollection.h
```diff
@@ -31,13 +31,17 @@ class StRHICfCollection : public StObject
...
+    void setRHICfEventNumber(UInt_t event);
     void setBunchNumber(UInt_t bunch);
     void setRunType(UInt_t type);
     void setTriggerNumber(UInt_t trigger);
     void setRunTime(Int_t idx, UInt_t time);
     void setRunTRGM(UInt_t trgm);
 
     UInt_t numberOfPoints() const;
+    UInt_t getRHICfRunNumber();
+    UInt_t getRHICfEventNumber();
```
plexoos:
```suggestion
    UInt_t getRHICfRunNumber() const;
    UInt_t getRHICfEventNumber() const;
```

StRoot/StEvent/StRHICfCollection.cxx
```diff
@@ -58,6 +62,8 @@ void StRHICfCollection::setRunTRGM(UInt_t trgm){mRunTRGM = trgm;}
 
 //=========== Get ===========//
 UInt_t StRHICfCollection::numberOfPoints() const {return mRHICfPointColl.size();}
+UInt_t StRHICfCollection::getRHICfRunNumber(){return mRHICfRunNumber;}
+UInt_t StRHICfCollection::getRHICfEventNumber(){return mRHICfEventNumber;}
```
plexoos:
```suggestion
UInt_t StRHICfCollection::getRHICfRunNumber() const { return mRHICfRunNumber; }
UInt_t StRHICfCollection::getRHICfEventNumber() const { return mRHICfEventNumber; }
```

klendathu2k (commented):
The code is clear and unobjectionable (save for constness which Dmitri has noted).  What is not clear is the rationale.  Why do you need to propagate a run and event number, when these are already provided by StEvent?  Two possible reasons come to mind...

1) Downstream analysis software expects them to be part of the StEvent / StMuDst and/or pico dst objects.  If this is the case, then we _should not_ approve this pull request.

2) The RHICf detector has encoded a run and event numbering scheme that differs from the STAR scheme.   If this is the case, why do we need to have two different run/event numbering schemes in our event data model?  This is guaranteed to cause confusion in the future.

genevb:
> Why do you need to propagate a run and event number, when these are already provided by StEvent? ...
> 2. The RHICf detector has encoded a run and event numbering scheme that differs from the STAR scheme.   If this is the case, why do we need to have two different run/event numbering schemes in our event data model?  This is guaranteed to cause confusion in the future.

To expedite, I will try to summarize my understanding on behalf of the RHICf folks:

#2 is correct, and this is because RHICf did take their own raw data outside of the STAR DAQ system. That raw data has its own runs and events. By including the _RHICf_ run and event numbers in addition to the _STAR_ run and event numbers that are already accessible in the DSTs, the RHICf group can correlate between findings they made through their own raw data investigations with what they get through the STAR framework.

klendathu2k:
> https://github.com/star-bnl/star-sw/pull/2 is correct, and this is because RHICf did take their own raw data outside of the STAR DAQ system.

Thanks Gene.  I thought that would be the case.  Inclusion in the STAR event data model is reasonable,  But it absolutely must be documented in the header file.

StRoot/StEvent/StRHICfCollection.cxx
```diff
@@ -50,6 +52,8 @@ const std::vector<StRHICfPoint*>& StRHICfCollection::pointCollection() const {re
 void StRHICfCollection::isAllSave(){mRHICfHitColl = new StRHICfHit();}
 
 //=========== Set ===========//
+void StRHICfCollection::setRHICfRunNumber(UInt_t run){mRHICfRunNumber = run;}
```
klendathu2k, line 55: It is important that we not confuse people looking at the data in the future.  Please provide a comment block for the run and event number setters, getters and member variables.  These comment blocks should clearly state what is being stored/retrieved, and why it differs from the STAR run and event numbering scheme.

StRoot/StEvent/StRHICfCollection.h
```diff
@@ -49,6 +53,8 @@ class StRHICfCollection : public StObject
     StRHICfHit* mRHICfHitColl; 
     std::vector<StRHICfPoint*> mRHICfPointColl; 
 
+    UInt_t mRHICfRunNumber;
```
genevb, line 58:
```suggestion
    // Important note: the _RHICf_ Run and Event numbers are distinct from the _STAR_ Run and Event numbers,
    // originating from RHICf's own raw data acquisition and studies outside of the STAR framework
    UInt_t mRHICfRunNumber;
```

ggfdsa10:
Dear Jason,
Thank you for your comments 

and Gene, 
Thank you for your explanation on behalf of RHICf and me

plexoos (approved):

klendathu2k (approved):
LGTM

PR #565, opened 2023-07-13
Added vertex information to track projections in StFcsDb

dkapukchyan:
Changed functions in StFcsDb that are related to projecting a g2t track onto the FCS detector plane and now can also take vertex information from g2t vertex.


akioogawa (approved):

plexoos:
@dkapukchyan Can you remind me if StFcsDb was already used in any official data processing? If not, it's not worth going through the changes in logic

dkapukchyan:
@plexoos Yes it has been used for a while now.

plexoos:
In that case we need to be careful about the changes you introduce. Unless we are fixing a bug, our general guidance is to be able to reproduce the past productions with newer releases. Although, from the description in the first post and the code it looks like the changes mostly deal with simulated data.

dkapukchyan:
I think only the getDetectorOffset function was changed in such a way; it can now return the offset at any zdepth whereas before it just returned a single value. However, the default is to give the value it did before this addition and in my tests it worked fine both ways. The other changes arise because these are the way the projection functions were supposed to work but forgot to push those changes before my other pull request was approved.

dmixsmi (approved):

plexoos (approved):

PR #568, opened 2023-07-20
add configuration files to set up environments with ROOT6

klendathu2k:
## What this PR provides / fixes

- Adds configuration for root6 build.
- Modifies OnlTools, adding explicit header to access gVirtualX.
- Modifies forward tracker to utilize TDirectory::cd(), rather than the (deprecated) assigment operator.

## What doesn't compile under this PR

TPCCATracker does not compile against vc-1.4.1, which is used by star-env-0.2.3-root-6.24.06.  The star-env needs to stay with the older 0.7.4 version of vc (or TPPCA needs to be updated to use the newer vc-1.4.1).

## To build:
```
$ setup 64b
$ starver dev config/v0.2.3-rhel7-root6.34.38-64b
$ setenv SKIP_DIRS "StEbyePool StEbyePool StHighptPool StAngleCorrMaker StSpinMaker StEbyeScaTagsMaker StEbye2ptMaker StDaqClfMaker StFtpcV0Maker StStrangePool GeoTestMaker StSpinPool StFgtPool PWGTools StShadowMaker TPCCATracker StiCA"
$ cons
```
note:  remove the TPCCATracker from the `SKIP_DIRS` list, and you will reach the following error:
```
 #error "You are using the old VC_IMPL macro. Since Vc 1.0 all Vc macros start with Vc_, i.e. a lower-case 'c'"
```


mgr/config/v0.2.3-rhel7-root6.34.38-64b.config
```diff
@@ -0,0 +1,11 @@
...
+
+# this is in place of buggy `module clear` (module VERSION=3.2.10 DATE=2012-12-21)
+unsetenv _LMFILES_*
+module purge
+module unuse /cvmfs/star.sdcc.bnl.gov/star-spack/spack/share/spack/modules/linux-rhel7-x86/
+module use   /cvmfs/star.sdcc.bnl.gov/star-spack/spack/share/spack/modules/linux-rhel7-x86_64/
+module load star-env-0.2.3-root-6.24.06
+module load star-table-0.1.0-root-6.24.06
+module load libiconv-1.16
+echo Configure for roo6.24.06
```
genevb: Why is the file titled for root6.34.38 while configuring for root6.24.06?

StRoot/StFwdTrackMaker/StFwdTrackMaker.cxx
```diff
@@ -258,7 +258,7 @@ int StFwdTrackMaker::Finish() {
 
     mForwardTracker->finish();
 
-    gDirectory = prevDir;
+    prevDir->cd();
```
genevb, line 261: Agreed: we shouldn't be setting global ROOT variables ourselves.

plexoos: Do you see it as a problem with ROOT6 only?

veprbl:
It should be possible to install vc 0.7 externally and disable it in ROOT.

mgr/config/v0.2.3-rhel7-root6.34.38-64b.config
klendathu2k: Doh!  Should be fixed now.

StRoot/StFwdTrackMaker/StFwdTrackMaker.cxx
klendathu2k: gDirectory = prevDir compiles under root 5, but fails to compile under root 6.24.  I looked into the root 6.24 code.  They deleted the assignment operator, and made other changes to make TDirectory thread safe.

klendathu2k:
> It should be possible to install vc 0.7 externally and disable it in ROOT.

I believe it already is.   
```
$ module avail vc
---------------------------------------------------------- /cvmfs/star.sdcc.bnl.gov/star-spack/spack/share/spack/modules/linux-rhel7-x86_64/ 
vc-0.7.4 vc-1.4.1
```

plexoos:
Have you tried to use ROOT 6.16 first? That's the version we have used in CI since ROOT6 was included in CI for almost two years. 6.16 is also what we agreed on to start switching away from ROOT5.

StRoot/StFwdTrackMaker/StFwdTrackMaker.cxx
plexoos: it compiles fine with ROOT6.16 in our CI

klendathu2k:
> Have you tried to use ROOT 6.16 first? That's the version we have used in CI since ROOT6 was included in CI for almost two years. 6.16 is also what we agreed on to start switching away from ROOT5.

I will add a config for root 6.16.  

The reason that I went with 6.24 is because the 6.16 version provided in spack includes the root version of TTable, which we decided to move away from.

StRoot/StFwdTrackMaker/StFwdTrackMaker.cxx
klendathu2k: In root 6.16, gDirectory is declared in the TDirectory.h header as...
```
#define gDirectory (TDirectory::CurrentDirectory())
```
In root 6.24, ...
```
#define gDirectory (ROOT::Internal::TDirectoryAtomicAdapter(TDirectory::CurrentDirectory()))
```

With 6.24, it looks like they are trying to make TDirectory thread safe... so no direct writes to global variables.

genevb (commented):
So no 32-bit versions for ROOT6?
Otherwise, no further comments from me.

mgr/config/v0.2.3-rhel7-root6.24.06-64b.config
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env csh
+
+# this is in place of buggy `module clear` (module VERSION=3.2.10 DATE=2012-12-21)
+unsetenv _LMFILES_*
+module purge
+module unuse /cvmfs/star.sdcc.bnl.gov/star-spack/spack/share/spack/modules/linux-rhel7-x86/
+module use   /cvmfs/star.sdcc.bnl.gov/star-spack/spack/share/spack/modules/linux-rhel7-x86_64/
+module load star-env-0.2.3-root-6.24.06
+module load star-table-0.1.0-root-6.24.06
```
plexoos: Do we really need to load star-table here? Does the environment work without it?
star-table module loads the libraries built from this code https://github.com/star-bnl/star-table

klendathu2k: No, we don't need to load the table module... it is already provided by the star environment module.  Did a quick check... codes which depend on libTable compile w/out problems.

iraklic (approved):

plexoos (approved):

plexoos:
Should be good to go now...

PS Thank you Irakli! 🙇

PR #572, opened 2023-07-28
fix: consolidate 32-bit and 64-bit configs

plexoos:
Introduce a single config file capable of setting up either 32-bit or 64-bit environment.

Something like this should work:

```
setup 64b
starver dev
setup 32b
starver dev
```


genevb (approved):
LGTM.
-Gene

PR #577, opened 2023-08-10
Fix errant switch-case enumeration

genevb:
The cases in a switch statement went from 3 to 5, errantly skipping over 4. This patch affects no former productions as no detector > 3 has been used before now. However, for FXT data, we will want detector > 3, so we should fix this now for possible use in such productions.

It will also probably be worth patching this in SL23d (because this does not impact any past productions), allowing us to use it before the next major release step (e.g. SL23e).

### DIFF
--- a/StRoot/StEventUtilities/StEbyET0.cxx
+++ b/StRoot/StEventUtilities/StEbyET0.cxx
@@ -69,17 +69,17 @@ double StEbyET0::getTime(StEvent* event, int mode) {
                 if (info[0] > 0 && info[1] > 0)
                   coordinate = 0.5*(info[0]+info[1]);
                 break;
-      case  5 : // next 4 use east trigger detectors
+      case  4 : // next 4 use east trigger detectors
+      case  5 :
       case  6 :
-      case  7 :
-      case  8 : getTriggerInfo(event,static_cast<trigDetType> (detector-4),info);
+      case  7 : getTriggerInfo(event,static_cast<trigDetType> (detector-4),info);
                 if (info[0] > 0)
                   coordinate = info[0];
                 break;
-      case  9 : // next 4 use west trigger detectors
+      case  8 : // next 4 use west trigger detectors
+      case  9 :
       case 10 :
-      case 11 :
-      case 12 : getTriggerInfo(event,static_cast<trigDetType> (detector-8),info);
+      case 11 : getTriggerInfo(event,static_cast<trigDetType> (detector-8),info);
                 if (info[1] > 0)
                   coordinate = info[1];
                 break;


genevb:
BTW, the view of changes in github make this look a little more complicated than it is. Really, it's just replacing 5,6,7,8,9,10,11,12 with 4,5,6,7,8,9,10,11.

StRoot/StEventUtilities/StEbyET0.cxx
```diff
@@ -69,17 +69,17 @@ double StEbyET0::getTime(StEvent* event, int mode) {
                 if (info[0] > 0 && info[1] > 0)
                   coordinate = 0.5*(info[0]+info[1]);
                 break;
-      case  5 : // next 4 use east trigger detectors
+      case  4 : // next 4 use east trigger detectors
+      case  5 :
       case  6 :
-      case  7 :
-      case  8 : getTriggerInfo(event,static_cast<trigDetType> (detector-4),info);
```
veprbl, line 75: Interesting, so the original code here was able to provide unsupported kVPDeast = 8 - 4 = 12 - 8:
https://github.com/star-bnl/star-sw/blob/94fbc35658cd62fbeefee130575a7d385ac6ea9d/StRoot/StEventUtilities/StEbyET0.h#L17-L29
https://github.com/star-bnl/star-sw/blob/94fbc35658cd62fbeefee130575a7d385ac6ea9d/StRoot/StEventUtilities/StEbyET0.cxx#L143-L162

veprbl (approved):
LGTM

StRoot/StEventUtilities/StEbyET0.cxx
genevb: When detector = 4 (same as kVPDeast), then we want to pass kVPD to getTriggerInfo().

Are you concerned about doing arithmetic (i.e. 8 - 4) on the enumeration?

veprbl: I wasn't concerned. But I have a suggestion how to implement that, if you are interested. You could add
```c++
case kVPD:
case kVPDeast:
case kVPDwest:
info[0] = //...
```
to getTriggerInfo instead of that subtraction.

genevb: Oh, I see - yeah, that would work - thanks for the suggestion, @veprbl . I'll consider doing that in a separate PR that goes into a later library release. For now, I just want to make the minimal change so that we can patch SL23d.

plexoos (approved):

genevb:
@plexoos , could you please tag SL23d_1 to include this commit? I expect SL23e will be the first one that we tag with the spack configurations.

plexoos:
I believe we want to patch SL23d with this fix, here we go: https://github.com/star-bnl/star-sw/pull/578

PR #585, opened 2023-09-16
Picodst fixes

klendathu2k:
Leszek reported (on the simulations list) an issue with converting simulation MuDsts to picoDsts.  The code issues a couple of errors, and then crashes.  I believe this is due to a faulty assumption in initialization.  The code relies on the filename begining with "st_" in order to determine the basename of the file, when the full path is given.  This fails for Leszek's case (and for any pure simualtion).   When the assumption fails, the initialization routine returns before creating the TTree.  This results in the crash during Make, when the null pointer to the TTree is deferenced.

This PR removes the faulty assumption, and makes early returns from Init() a "fatal" return rather than just an error.

### DIFF
--- a/StRoot/StPicoDstMaker/StPicoDstMaker.cxx
+++ b/StRoot/StPicoDstMaker/StPicoDstMaker.cxx
@@ -278,20 +278,20 @@ Int_t StPicoDstMaker::Init() {
         
         if (setVtxModeAttr() != kStOK) {
 	  LOG_ERROR << "Pico Vertex Mode is not set ... " << endm;
-	  return kStErr;
+	  return kStFatal;
 	}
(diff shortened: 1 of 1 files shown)


plexoos (approved):

nigmatkulov (commented):
Hi Jason, as I said before, it is not a mistake (requiring the "st_") really, but rather a historical reason (do not remember the details but we may rediscover them again :-)  As for replacing StErr with StFatal, for some cases we may run a simplistic simulations without some detector subsystems.

Anyhow, I am okay with the changes but just wanted to say that. Please think about what I said and let me know. Then I'll merge the branch if you want this PR to go forward.

nigmatkulov (approved):
I worry about kStFatal, but let's see if this will work out.

Issue #590, opened 2023-09-28
Memory leak in StFstClusterMaker


genevb:
Processing 500 Run 22 pp500 events in valgrind without TPC, just the FWD detector makers, I get 2 definite memory leaks and 2 possible memory leaks, shown below.

Here is the chain I ran. This took ~1 hour outside valgrind and ~2 hours inside valgrind. Fewer events would probably be fine to show the problems if you want to run it faster.

    setup 64b
    stardev
    nohup valgrind --leak-check=full root4star -b -q -l 'bfc.C(500,"DbV20230818 pp2022a -btof -mtd -eventqa -emcDY2 -TpcHitMover -ittf -tpx -tpc Notpc_daq fst ftt fstRawHit -picoEvt -picoDst -picoWrite -picoVtxDefault -picoCovMtxWrite -trgSimu BEmcChkStat evout -hitfilt","/star/data03/daq/2022/036/23036001/FTEST/st_physics_23036001_raw_5000040.daq")' >& log &

Thanks
-Gene

==26363== 2,621,528 bytes in 46,813 blocks are definitely lost in loss record 278,294 of 278,315
==26363==    at 0x76AD203: operator new(unsigned long) (vg_replace_malloc.c:334)
==26363==    by 0x7E01968: TStorage::ObjectAlloc(unsigned long) (TStorage.cxx:325)
==26363==    by 0x297D875C: operator new (TObject.h:156)
==26363==    by 0x297D875C: StFstScanRadiusClusterAlgo::doClustering(StFstCollection const&, StFstRawHitCollection&, StFstClusterCollection&) (StFstScanRadiusClusterAlgo.cxx:55)
==26363==    by 0x297D840F: StFstIClusterAlgo::doClustering(StFstCollection&) (StFstIClusterAlgo.cxx:25)
==26363==    by 0x297D7AF3: StFstClusterMaker::Make() (StFstClusterMaker.cxx:59)
==26363==    by 0x14F1309E: StMaker::Make() (StMaker.cxx:937)
==26363==    by 0x14F1309E: StMaker::Make() (StMaker.cxx:937)
==26363==    by 0x14F0E4E7: StChain::EventLoop(int, int, StMaker*) (StChain.cxx:206)
[...]

PR #593, opened 2023-10-01
Fix memory leak in StFstClusterMaker

techuan-huang:
Memory leaks were found by Gene in issue #590.
There are 3 places causing the leak in `StFstClusterMaker/StFstScanRadiusClusterAlgo.cxx`:
1. Pointers assigned to std::vector in `line 55` were not deleted before calling std::vector::pop_back() in `line 69` and `line 80`.
2. Object created in `line 140` but only conditionally assigned to a std::vector in `line 148`.
3. In `line 203`, std::vector::erase() was called without deleting the pointers.

After fixing these places, there is no memory leak from StFstClusterMaker reported by valgrind.


genevb (commented):
This patch has resolved the significant increase in reconstruction time per event over the course of processing files that I [observed and documented](https://drupal.star.bnl.gov/STAR/blog/genevb/Run-22-pp500-reconstruction-slowdowns) when including FWD chain options. I'm very happy to see that!

jdbrice (approved):

plexoos (approved):

PR #596, opened 2023-10-06
Adding fcsDataPresent flag

akioogawa:
Adding a flag to tell if FCS data was present or not.


StRoot/StEvent/StFcsCollection.h
```diff
@@ -64,9 +67,9 @@ class StFcsCollection : public StObject {
     StSPtrVecFcsPoint   mPoints[kFcsNDet];  
 
     Int_t mFcsReconstructionFlag=0;     // undefined for now
+    int mDataExist=0;
```
klendathu2k: As this is involved in IO, the preference is to use Int_t  instead of int.

klendathu2k (dismissed):
Single change in StFcsCollection requested, otherwise good.

I am not reviewing anything in the pool directory... in fact... I suggest making the changes in the pool area a separate commit so that it is obvious that they have nothing to do with changing our event data model.

StRoot/StEvent/StFcsCollection.h
plexoos: From the name of this variable, it appears it can be boolean

akioogawa: It's single int per event.... I don't think we need this at the moment, but there are zero-suppressed data and raw-none-zero-suppressed data from DAQ. We only read zero-suppressed data at offline so far, unlike some online reader which reads raw. If we ever needed to read raw, we can turned this int meaning 0=no data, 1=ZS data, 2=raw, 3=embediing...etc. Keeping it Int_t.

akioogawa:
Many test failing with "St_db_Maker::LoadTable(TDataSet*): Assertion `!ee' failed."...  I have no idea. Seems unrelated to my change of 1 line in StEvent...

plexoos:
Those failed jobs are tests with ROOT6. Unrelated to your changes. I restarted them and pretty sure they will pass without issues.

plexoos (approved):

StRoot/StEvent/StFcsCollection.h
```diff
@@ -64,9 +67,9 @@ class StFcsCollection : public StObject {
     StSPtrVecFcsPoint   mPoints[kFcsNDet];  
 
     Int_t mFcsReconstructionFlag=0;     // undefined for now
```
zlchang, line 69: Just a suggestion, maybe put the initialization in the constructor would be better.

StRoot/StSpinPool/StFcsRawDaqReader/StFcsRawDaqReader.cxx
```diff
@@ -154,7 +154,7 @@ Int_t StFcsRawDaqReader::Make() {
       if(trgcmd != 4 && trgcmd !=10){  // 4=phys/ped 10=LED
 	nskip++;
 	nskiptot++;
-	if(mDebug) printf("trgcmd=%d skipping nskip=%d nskiptot=%d\n",trgcmd,nskip,nskiptot);
+	if(mDebug) LOG_INFO << Form("trgcmd=%d skipping nskip=%d nskiptot=%d",trgcmd,nskip,nskiptot)<<endm;
```
zlchang, line 157: Is there a need to introduce mDebug, maybe just use LOG_DEBUG<<Form(...)<<endm;?

StRoot/StSpinPool/StFcsRawDaqReader/StFcsRawDaqReader.cxx
```diff
@@ -330,7 +333,7 @@ Int_t StFcsRawDaqReader::Make() {
       if(detid<6) nvaliddata++;      
       
       if(mDebug){
-	printf("FCS %3s : S%d:%d [det %d, ns %d, dep %d ch %d] det=%d id=%3d : size=%d : adc=",
+	printf("FCS %3s : S%2d:%2d [det%1d ns%1d dep%2d ch%2d] det=%d id=%3d size=%3d adc=",
```
zlchang, line 336: Same as here, LOG_DEBUG?

StRoot/StSpinPool/StFcsRawDaqReader/StFcsRawDaqReader.cxx
```diff
@@ -353,7 +356,7 @@ Int_t StFcsRawDaqReader::Make() {
     dd = mRdr->det("stgc")->get("altro",r) ;
     while(dd && dd->iterate()) {    //per xing and per RDO    
       if(mDebug) printf("STGC ALTRO: stgc%02d(sec) RDO=%1d ALTRO=%03d(row) Ch=%02d(pad)\n",dd->sec,r,dd->row,dd->pad);
-      for(uint32_t i=0; i<dd->ncontent; i++) {
+      for(u_int i=0; i<dd->ncontent; i++) {
 	if(mDebug) printf(" TB=%3d ADC=%4d",dd->adc[i].tb,dd->adc[i].adc) ;
```
zlchang, line 360: LOG_DEBUG and the same applies to a few lines below if(mDebug) printf("\n")?

zlchang (approved):

klendathu2k (approved):
StEvent changes are fine.  Generally I agree w/ Zilong's comments about LOG_DEBUG the mDebug parameter.  But not a show stopper ... especially for a "pool" code.

PR #599, opened 2023-10-12
fix: adjust ROOT6 test commands for star-env@0.3.0

plexoos:


veprbl (approved):

plexoos:
@veprbl thanks for approving! This PR was kind of a check for the new env.
I think what we really want is #598 . What do you think?

veprbl:
Oh, yes. I see, this is a subset of that.

plexoos:
superseded by #598

PR #605, opened 2023-10-13
Change from int to StarMagField::EBField type

genevb:
Resolves a run-time error seen in tests of compiling software for ROOT 6.16.00. I also tested this fix in ROOT 5.34.38 to be sure it doesn't break it.

-Gene
____________________

    root4star -b -q -l 'bfc.C(-1)'

...

QA :INFO  - ProcessLine if (!StarMagField::Instance()) new StarMagField( 2, 1.000000, kTRUE);
input_line_668:2:37: error: no matching constructor for initialization of 'StarMagField'
 if (!StarMagField::Instance()) new StarMagField( 2, 1.000000, kTRUE);
                                    ^             ~~~~~~~~~~~~~~~~~~
/afs/rhic.bnl.gov/star/packages/SL23x/.sl73_x8664_gcc485/include/StarMagField.h:155:3: note: candidate constructor not viable: no known conversion from 'int' to 'StarMagField::EBField' for 1st argument
  StarMagField ( EBField map     = kMapped, Float_t Factor  =      1, 
  ^
/afs/rhic.bnl.gov/star/packages/SL23x/.sl73_x8664_gcc485/include/StarMagField.h:91:7: note: candidate constructor (the implicit copy constructor) not viable: requires 1 argument, but 3 were provided
class StarMagField : public TVirtualMagField 
      ^
BFC:ERROR - StBFChain::ProcessLine command:if (!StarMagField::Instance()) new StarMagField( 2, 1.000000, kTRUE); has failed. Quit job.

### DIFF
--- a/StRoot/StBFChain/StBFChain.cxx
+++ b/StRoot/StBFChain/StBFChain.cxx
@@ -1918,7 +1918,7 @@ void StBFChain::SetDbOptions(StMaker *mk){
 			 << "\",\"MagFactor\")" << endm;
       if ( gClassTable->GetID("StarMagField") >= 0) {
 	TString cmd =
-	  Form("if (!StarMagField::Instance()) new StarMagField( 2, %f, kTRUE);",
+	  Form("if (!StarMagField::Instance()) new StarMagField( StarMagField::EBField::kMapped, %f, kTRUE);",
 	       FieldOptions[k].scale);
 	ProcessLine(cmd);
       }


klendathu2k (approved):

plexoos (approved):
Hm, it is strange that we don't see this runtime error for StarMagField in our CI jobs with ROOT 6.16. We don't run specifically 'bfc.C(-1)' but I would think the initialization is the same.

The change itself is good. We should favor named constants

genevb:
> Hm, it is strange that we don't see this runtime error for StarMagField in our CI jobs with ROOT 6.16. We don't run specifically 'bfc.C(-1)' but I would think the initialization is the same.

Running `bfc.C(-1)` results in turning on the chain option `Test.default.ITTF` (see below) which leads eventually to turning on the option `FieldOn`, which causes this line of code to be called in StBFChain.cxx. Normal chains do not have these options turned on. It isn't the most important test, but it's simple to execute and does test a lot of details.

From bfc.C:
```
TString defChain("y2010,gstar,Test.default.ITTF,NosvtIT,NossdIT,-sfs,-ssdFast,sdt20100107.110000");
```

PR #607, opened 2023-10-17
More STV Cleanup

klendathu2k:
Creating a draft PR.  Additional Stv cleanup:

Removes IDL files.
Removes chain options, and the configuration logic in StBFChain...
(note:  broken out over several commits as I refactored the slightly complicated control flow...)

Removed the StXTrakMaker.  This depends on StvUtil.  StvUtil was removed, but... our code still compiles.  And I don't see any other xtrak dependencies... so draft PR should reveal whether it is a problem or not.


plexoos (approved):

iraklic (approved):
As I understand this is just a cleanup of Stv, and I assume this is full cleanup, I have not checked if there is Svt in any other option before the approval. But I also assume even if there is leftover it won't brake anything anyhow.

klendathu2k:
Hi Irakli,

Yes and no.  Yes, this is part of a general cleanup of Stv.  Yes, if 
there
is anything left it will not break things.  No, there is not a zero risk 
of
this breaking code.  In particular the changes to StBFChain.

There were several places where the presence (or lack-thereof) of an Stv 
chain
option altered the control flow.  Many of them were logical OR's, which 
were
easy to refactor...

if ( A || B || GetOption("Stv") ) { ... }

A few were slightly more complex.   I likely got those right, but it 
would be good
to have someone carefully review the changes to StBFChain.

Cheers,
Jason


There were several instances in StBFChain where the
presence of the Stv option, the StvMaker, etc... was checked in logic.  
I believe
that I have

iraklic:
Thanks for confirming what I thought. I can try to go over the options and land you a second pair of eyes. I will let you know if I find something but let's keep the "approval" :)

klendathu2k:
I think this is ready to squash/merge... but would like to give @genevb the opportunity to review the BigFullChain.h and StBFChain.cxx changes first...

genevb (approved):
I went carefully through the changes line by line under `StRoot/StBFChain` and I didn't spot any flaws. Possible I missed something too, but at least multiple eyes have looked now to improve our confidence.

PR #616, opened 2023-10-30
Remove the St_geant_Maker dependencies in the event generator to starsim bridge

klendathu2k:
Use direct calls into the starsim / GEANT3 libraries to push particles onto the G3 stack, eliminating the dependence on St_geant_Maker and the version of GEANT3 required by starsim.  This will allow the event generator framework to function
with either our legacy starsim / St_geant_Maker framework, or the new VMC / StGeant4Maker framework.

### DIFF
--- a/StRoot/StarGenerator/BASE/AgStarReader.cxx
+++ b/StRoot/StarGenerator/BASE/AgStarReader.cxx
@@ -5,13 +5,10 @@ ClassImp(AgStarReader);
 #include "TDatabasePDG.h"
 #include "TParticlePDG.h"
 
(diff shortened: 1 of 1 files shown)


plexoos:
Does it require any further testing?

klendathu2k:
> Does it require any further testing?

It shouldn't.  The refactoring was straightforward (just lifting the calls to add particles out of St_geant_Maker and into the bridge.)  Test with single particles showed that hits were being registered in detectors.  So this should be good to go.

plexoos (approved):

PR #619, opened 2023-11-01
ci: expand test job matrix with ROOT 6 environment

plexoos:


klendathu2k:
I took a look at the Sti geometry test failures.  I cannot reproduce the problem in the production environment.  Is there a way for me to run the code in the same environment and against the same build as the failed tests?

plexoos:
of course https://github.com/star-bnl/star-sw/issues/620#issuecomment-1832710446

plexoos:
duplicates #606

PR #621, opened 2023-11-02
Initialize variable geomtry

fisyak:


StRoot/StGenericVertexMaker/StiPPVertex/BtofHitList.cxx
```diff
@@ -46,6 +46,7 @@ BtofHitList::initRun(St_db_Maker* db_maker) {
 
   // Initialize BTOF geometry
   TObjectSet *geom = (TObjectSet *) db_maker->GetDataSet("btofGeometry");
+  geometry = 0;
```
klendathu2k: geometry is a member of the BtofHitList class.  It should be initialized in the class constructor rather than at this point.

klendathu2k (changes requested):

StRoot/StGenericVertexMaker/StiPPVertex/BtofHitList.cxx
genevb: ~~But...database calls are generally not done until the timestamp of the data has been established. Generally for makers, that doesn't happen before StMaker::InitRun(). I haven't looked, but moving it depends on when this class (not an StMaker) is constructed.~~

genevb: Sorry....Ignore my comment.... you aren't talking about the db_maker call.

iraklic: @klendathu2k @fisyak : so will this change be moved or we go ahead with this commit?

plexoos: It'd be interesting to see what kind of edge case induced this change. From the if statements it looks like the geometry object may not update for a new run.

klendathu2k (approved):
Thanks!  Looks good to me.

plexoos (approved):

Issue #620, opened 2023-11-02
Runtime errors in StarVMC/StarAgmlLib when using ROOT 6.24


plexoos:
After a couple of fixes, our code seems to compile fine against ROOT6.24 (See https://github.com/star-bnl/star-sw/pull/619)

However, some test jobs fail at what appears to be the geometry loading stage. The errors point to the code in `StarVMC/StarAgmlLib`. Here is an example error message:

https://github.com/star-bnl/star-sw/actions/runs/6727541469/job/18287828236?pr=619

```
In file included from /tmp/root/spack-stage/spack-stage-root-6.24.06-gbmdgvxscayjf6nqc5tatcs424pan4gs/spack-build-gbmdgvx/input_line_9:15:
In file included from /tmp/root/spack-stage/spack-stage-root-6.24.06-gbmdgvxscayjf6nqc5tatcs424pan4gs/spack-build-gbmdgvx/etc/dictpch/allHeaders.h:120:
In file included from /usr/lib/gcc/x86_64-redhat-linux/4.8.5/../../../../include/c++/4.8.5/set:60:
/usr/lib/gcc/x86_64-redhat-linux/4.8.5/../../../../include/c++/4.8.5/bits/stl_tree.h:1423:13: error: no viable conversion from returned value of type 'pair<...>' to function return type 'pair<...>'
            return _Res(0, _M_rightmost());
                   ^~~~~~~~~~~~~~~~~~~~~~~
/usr/lib/gcc/x86_64-redhat-linux/4.8.5/../../../../include/c++/4.8.5/bits/stl_tree.h:1673:19: note: in instantiation of member function 'std::_Rb_tree<int, std::pair<const int, std::vector<TString, std::allocator<TString> > >, std::_Select1st<std::pair<const int, std::vector<TString, std::allocator<TString> > > >, std::less<int>, std::allocator<std::pair<const int, std::vector<TString, std::allocator<TString> > > > >::_M_get_insert_hint_unique_pos' requested here
            auto __res = _M_get_insert_hint_unique_pos(__pos, _S_key(__z));
                         ^
[...]

PR #623, opened 2023-11-02
Hide AgML library codes from the macros which load the geometry.  

klendathu2k:
Possible fix for issue #620.


klendathu2k:
All tests succeeded.  Syncing with main and changing from draft to normal PR.

StarDb/AgMLGeometry/CreateGeometry.h
```diff
@@ -2,8 +2,6 @@
 
 #if !defined(__CINT__) || defined(__CLING__)
 #include "StBFChain/StBFChain.h"
-#include "StarVMC/StarAgmlLib/AgModule.h"
```
klendathu2k, line 5: Probable culprit for the test failures.  The AgML library codes are overly complex... especially the shape and structure handling codes.  Would not be surprise if cling doesn't handle them well...

iraklic (approved):

plexoos:
I've applied this patch to #619 but there are still errors. However, this PR passed the existing test so, I am OK merging it

plexoos (approved):

PR #627, opened 2023-11-06
Fix daq5k gains, add TPC23 to chain

fisyak:
Fix daq5k TPC gains, Add TPC23 to the BFC chain.
See https://drupal.star.bnl.gov/STAR/blog/fisyak/TPC-daq5k-offline-cluster-finder-updates


genevb (commented):
I would like some clarification on this....

At first glance, it looks like the new "TPC23" chain option is identical to the "TpxClu2D" chain option. However, the latter causes the "TpxClu2D" attribute to be set, while the former causes no attributes to be set, so they are slightly different. It does appear that the TPC hit maker codes do test for a "TPC23" attribute, but it's unclear where that gets set.

Is this the desired behavior, or should there be a line in StBFChain.cxx setting the "TPC23" attribute when this chain option is used?

StRoot/StTpcHitMaker/StTpcHitMaker.h
```diff
@@ -119,16 +120,20 @@ class StTpcHitMaker : public StRTSBaseMaker {
...
   void    PadMonitor(Int_t sector);
   Int_t   UpdateHitCollection(Int_t sector);
-  void    DumpPixels2Ntuple(Int_t sector);
+  void    DumpPixels2Ntuple(Int_t sector, Int_t row, Int_t pad);
   void    PrintSpecial(Int_t sector);
   Int_t   RawTpcData(Int_t sector);
   Int_t   RawTpxData(Int_t sector);
   void    InitializeHistograms(Int_t token);
+  void    CheckThrSeq();
+  TH2C   *Thr() {return fThr;}
```
klendathu2k, line 129: [minor].  Can diagnostic histograms be provided through the maker's histogram list mechanism, rather than direct access on the maker?

StRoot/StTpcHitMaker/StTpcRTSHitMaker.h
```diff
@@ -71,39 +71,35 @@ class daq_itpc;
...
   tpc23_base   *fTpx23; //!
   tpc23_base  *fiTpc23; //!
-#endif /*  __TFG__VERSION__ */
+  Int_t         FixGains(tpc23_base *tpc23, Int_t sector, Int_t row, Int_t npads); // fix dead pads and edges
   Bool_t    fNoiTPCLu; //!
   Double_t  fminCharge; // ! minimum cluster charge in ADC
   Int_t     maxHits[24];
   Int_t     maxBin0Hits;
   Int_t     bin0Hits;
+  static  StTpcRTSHitMaker *fgStTpcRTSHitMaker;
```
klendathu2k, line 102: Again minor.  Would prefer to use the standard StMaker::GetMaker rather than static access method.

iraklic (approved):

klendathu2k (changes requested):
Generally looks fine.  I flagged two places where I would recommend improvements -- using standard StMaker mechanisms for saving histograms and making them available, plus standard access to makers -- but don't consider them show stoppers.  Once Gene's comments are addressed, I'll approve.

genevb:
> should there be a line in StBFChain.cxx setting the "TPC23" attribute when this chain option is used?

Just a comment that I think Yuri addressed this question.

plexoos:
Should we include this in SL23e?

plexoos:
If no further comments, let's merge

PR #630, opened 2023-11-21
Moving some initiization from Init to InitRun for StFcsDb

akioogawa:
Those initialization had to wait till tables were filled from DB in StFcsDbMaker::InitRun.


dmarkh (approved):
LGTM

plexoos (approved):

genevb:
Do we need to update this in SL23f so it will be used in the Run 22 physics preview production we're about to start? We already tagged and built the library, but haven't started the production yet, so we could update. Akio?

akioogawa:
I don't think so. This only affect "ET gain match" calc, which is only used online trigger and trigger simulator, which is not part of production.

PR #633, opened 2023-12-04
A minor logic fix in StFcsClusterMaker::isNeighbor()

akioogawa:
Kumar found a logic flaw in isNeighbor() thus fixing it. Effect is minor


akioogawa:
Kumar showed the effect on this is very minor : https://drupal.star.bnl.gov/STAR/event/2023/12/01/FWD-Software/StFcsClusterMaker-Neighboring-Tower-Issues

jdbrice (approved):

plexoos (approved):

klendathu2k (approved):
Seems like this is what was originally intended.  Only comment is that line 284/285 could probably be optimized as ` if ( ehp==0 ){ ... } else {...}`.

Issue #631, opened 2023-11-28
SL23f crash in StFstRawHitMaker


genevb:
This doesn't crash for all DAQ files, but it does for some. Here is a crashing chain to try:

    starver SL23f
    root4star -b -q -l 'bfc.C(10,"DbV20231127 pp2022a StiCA fst ftt fstRawHit fstMuRawHit BEmcChkStat -hitfilt","/star/data03/daq/2021/352/22352002/st_fwd_22352002_raw_1500011.daq")'

I ran in the debugger and got this....

    StFstRawHitMaker:WARN  - StFstRawHitMaker::Make() - No raw ADC dataset found from simu data! 
    StFstRawHitMaker:WARN  - StFstRawHitMaker::Make() - No fstCollection found in simu dataset! 
     StFstRawHitMaker:INFO  -  Trying to read ALLdata
    *** Error in `/afs/rhic.bnl.gov/star/packages/SL23f/.sl73_gcc485/bin/root4star': malloc(): memory corruption: 0x123aff00 ***

....followed by a long backtrace. I'm not sure why it mentions `/tmp/smirnovd` in here:

(gdb) where
#0  0xf7fdb425 in __kernel_vsyscall ()
#1  0xf4f1e1f7 in raise () from /lib/libc.so.6
#2  0xf4f1fa33 in abort () from /lib/libc.so.6
#3  0xf4f5d5e5 in __libc_message () from /lib/libc.so.6
#4  0xf4f66a03 in _int_malloc () from /lib/libc.so.6
#5  0xf4f6818a in malloc () from /lib/libc.so.6
#6  0xf5113b27 in operator new(unsigned int) () from /lib/libstdc++.so.6
#7  0xf7dd48dd in TStorage::ObjectAlloc (sz=52)
    at /tmp/smirnovd/spack-stage/spack-stage-root-5.34.38-fta7antlmbz65avo4vw6tf7xsbtghfc4/spack-src/core/base/src/TStorage.cxx:325
#8  0x0808df27 in TObject::operator new (sz=52)
    at /cvmfs/star.sdcc.bnl.gov/star-spack/spack/opt/spack/linux-rhel7-x86/gcc-4.8.5/root-5.34.38-fta7antlmbz65avo4vw6tf7xsbtghfc4/include/TObject.h:156
[...]

PR #634, opened 2023-12-07
Fix job crashing for 9 time-bin FST data

techuan-huang:
Issue #631 was due to an inconsistent number of time bins used in the data and the codes.
In most FST data and in the codes, we used 3 time bins, while some data used 9 time bins.
Only changing a constant in `StRoot/StEvent/StFstConsts.h` is needed to fix this issue.


genevb:
Thanks, @techuan-huang . Two notes from me:
- I tried my test job with this patch and it ran to completion for me without crashing. I was careful to rebuild all the libraries that depend on this one file.
- If this passes review, we should tag `SL23f_1` and I will rebuild the library. `SL23f` has not been used for any other productions besides the preview production yet, so we should be fine with this patch.

-Gene

plexoos (approved):
If the tests confirm that the issues is fixed then it is fixed

jdbrice (approved):

genevb:
Are we waiting for @akioogawa 's approval, or can this be merged now?

PR #636, opened 2023-12-17
QA for StEvent and MuDst FwdTracks

jdbrice:
StFwdAnalysisMaker provides basic QA of the forward tracking from either the MuDst or StEvent collections.
It also serves as an example for how to analyze the FwdData.

Note - as it is currently envisioned this would never run during a production.
However, it may be beneficial to make it work with the production QA histogram system upon request, we can update to that end if needed.


akioogawa (approved):

StRoot/StFwdUtils/StFwdAnalysisMaker.h
```diff
@@ -27,6 +18,51 @@ class StFwdAnalysisMaker : public StMaker {
...
+    /**
+     * @brief Add a histogram to the map
+     * 
+     * Convenience method to avoid duplicate typing name / possible mismatch
+     * @param h1 TH1* histogram object
+     * @return TH1* return the same object for ease of use
+     */
+    TH1 * addHist( TH1 * h1){
+      mHists[h1->GetName()] = h1;
+      return h1;
```
genevb: Return the same pointer that was passed to it? Seems a bit odd, but generally harmless I suppose.

StRoot/StFwdUtils/StFwdAnalysisMaker.cxx
```diff
@@ -49,30 +55,101 @@
...
     LOG_DEBUG << "StFwdAnalysisMaker::Make" << endm;
     StEvent *event = (StEvent *)GetDataSet("StEvent");
-    if (!event){
-        LOG_INFO << "Cannot find StEvent" << endm;
-        return kStOK;
+    if (event){
+        StFttCollection *fttCol = event->fttCollection();
+        if (fttCol){
+            LOG_INFO << "The Ftt Collection has " << fttCol->numberOfPoints() << " points" << endm;
+        }
```
genevb: Maybe better to move this into the `ProcessFwdTracks()` routine and have a similar report from the MuDst in `ProcessFwdMuTracks()`? As it is, you obtain the StEvent pointer twice when using StEvent, and cast it differently the two times you obtain it. Not critical, but odd.

StRoot/StFwdUtils/StFwdAnalysisMaker.h
```diff
@@ -27,6 +18,51 @@ class StFwdAnalysisMaker : public StMaker {
...
+     * @brief Get the Hist object from the map
+     *  - Additional check and safety for missing histograms
+     * @param n Histogram name
+     * @return TH1* histogram if found, otherwise a 'nil' histogram with one bin
+     */
+    TH1* getHist( TString n ){
+      if (mHists.count(n))
+        return mHists[n];
+      LOG_ERROR << "Attempting to access non-existing histogram" << endm;
+      return new TH1F( "NULL", "NULL", 1, 0, 1 ); // returning nullptr can lead to seg fault, this fails softly
```
genevb, line 45: Why not return 0?

StRoot/StFwdUtils/StFwdAnalysisMaker.h
genevb: Ah....I see that you're using `new TH1/2F()` in the argument, so this let's you get that pointer if you need it, though I see you aren't currently using the returned pointer. Another approach would have been to define `AddHist()` to simply take either the arguments of a TH1F or TH2F constructor and call `new` inside `AddHist()`, since instantiating a new hist is what you always do, but your approach should work fine too.

StRoot/StFwdUtils/StFwdAnalysisMaker.h
genevb: I guess you want the code to run without crashing even if a histogram isn't present. But I also see that the expected histogram is sometimes a TH2F, not a TH1F, on which you call `Fill()` with two doubles as arguments....and luckily this is defined for a TH1F, but if you ever add a TH3F you will have an error.

StRoot/StFwdUtils/StFwdAnalysisMaker.cxx
```diff
@@ -49,30 +55,101 @@
...
+    fOutput->cd();
+    for (auto nh : mHists) {
+        nh.second->SetDirectory(gDirectory);
+        nh.second->Write();
+    }
+
+    // restore previous directory
+    gDirectory = prevDir;
+
+    LOG_INFO << "Writing StFwdAnalysisMaker output" << endm;
```
genevb: My suggestion depends on what you want from the log statement....
- If you want to say your code has made it to the writing stage, you generally put this before the `Write()` call.
- If you want to say your code has completed the writing stage, you generally output something like "Done writing StFwdAnalysisMaker output", or some variation on that.

genevb (commented):
All my comments are minor.

StRoot/StFwdUtils/StFwdAnalysisMaker.cxx
jdbrice: Updated in [030a2f5](https://github.com/star-bnl/star-sw/pull/636/commits/030a2f51a0cb1aa2368fac8ec6070243646aeda0)

StRoot/StFwdUtils/StFwdAnalysisMaker.h
jdbrice: I left this as is, indeed it will be an issue with TH3, but I do not plan to use any. But will keep in mind if that changes.

StRoot/StFwdUtils/StFwdAnalysisMaker.cxx
jdbrice: moved in [030a2f5](https://github.com/star-bnl/star-sw/pull/636/commits/030a2f51a0cb1aa2368fac8ec6070243646aeda0)

StRoot/StFwdUtils/StFwdAnalysisMaker.h
jdbrice: Removed this and instead use the `AddHist(...)` from StMaker
see: [030a2f5](https://github.com/star-bnl/star-sw/pull/636/commits/030a2f51a0cb1aa2368fac8ec6070243646aeda0)

jdbrice:
With [030a2f5](https://github.com/star-bnl/star-sw/pull/636/commits/030a2f51a0cb1aa2368fac8ec6070243646aeda0) I believe all of Gene's comments are addressed

StRoot/StFwdUtils/StFwdAnalysisMaker.h
genevb: Even better!

StRoot/StFwdUtils/StFwdAnalysisMaker.cxx
```diff
@@ -1,78 +1,110 @@
...
+        }
 
-#include "St_base/StMessMgr.h"
-#include "StarClassLibrary/StPhysicalHelix.hh"
-#include "StarClassLibrary/SystemOfUnits.h"
+        // restore previous directory
+        gDirectory = prevDir;
 
+        LOG_INFO << "Done writing StFwdAnalysisMaker output to local file : " << mLocalOutputFile << endm;
+    }
```
plexoos, line 45: Someone please remind me whether it is okay for analysis makers to write histograms in their own root files or they should add them to the common *_hists.root file.

genevb: My interpretation of this code is that as long as `mLocalOutputFile` isn't set (via `setLocalOutputFile()`), the histograms do go to the common hist.root file.

plexoos: Sounds good.

PR #639, opened 2024-01-17
Fwd picodst integration

jdbrice:
Adds initial PicoDst integration for Forward systems:
- StPicoFwdTrack
- StPicoFwdVertex
- StPicoFcsHit
- StPicoFcsCluster

Details:
- SPicoFwdTrack is a mirror of the StEvent/StMuDst version WITHOUT storing the hits associated with each track, and without keeping extra info about Genfit object
- StPicoFwdVertex: These are secondary (or potentially primary when running on data without TPC) vertices found by the FWD tracker. They are stored in the StVertex collection at the StEvent/StMuDst level and only stored separately for the PicoDst if they pass certain quality criteria on their tracks constituents. These are mostly for e.g. Lambda analysis
- StPicoFcsHit/Cluster - slimmed down versions of StEvent/MuDst equivalents. The hits are required for jet studies.

Size comparisons:

Vanilla: 12.8 Kb / Event
+FCS: 17.1 Kb / Event
+Fwd Tracks: 17.8 Kb / Event


StRoot/StPicoEvent/StPicoArrays.cxx
```diff
@@ -76,6 +80,8 @@ int StPicoArrays::picoArraySizes [NAllPicoArrays] = { 1,    // StPicoEvent
                                                       100,  // StPicoBEmcSmdPHit
                                                       100,  // StPicoETofHit
                                                       100,  // StPicoETofPidTraits
+                                                      10,  // StPicoFwdTrack
```
nigmatkulov: I suggest to move these two lines after the MC information. It will be easier to maintain backward compatibility.

StRoot/StPicoEvent/StPicoFcsCluster.h
```diff
@@ -0,0 +1,84 @@
...
+    void setSigmaMin(float sigmaMin) { mSigmaMin = sigmaMin; }
+    void setSigmaMax(float sigmaMax) { mSigmaMax = sigmaMax; }
+    void setTheta(float theta) { mTheta = theta; }
+    void setChi2Ndf1Photon(float chi2ndfph1) { mChi2Ndf1Photon = chi2ndfph1; }
+    void setChi2Ndf2Photon(float chi2ndfph2) { mChi2Ndf2Photon = chi2ndfph2;}
+    void setFourMomentum(TLorentzVector p4) { mFourMomentumX = p4.X(); mFourMomentumY = p4.Y(); mFourMomentumZ = p4.Z(); mFourMomentumT = p4.T(); }
+    
+
+protected:
+    UShort_t mId=0;             // Eventwise cluster ID
```
nigmatkulov: Not sure if this type of initialization is allowed in ROOT5. I would move it to the constructor.

genevb:
From the discussion at this week's analysis meeting, it sounds like we should freeze an `SL24a` library shortly after this code is merged, perhaps after a day of testing it in DEV, as there is interest in using it in the ongoing preview production of Run 22 data (which is already resuming without this).

StRoot/StPicoEvent/StPicoArrays.cxx
jdbrice: Moved in [61525da](https://github.com/star-bnl/star-sw/pull/639/commits/61525da6d3206ad43d1a79b14d9a0813758cd469)

StRoot/StFwdTrackMaker/StFwdTrackMaker.h
```diff
@@ -182,6 +182,184 @@ class StFwdTrackMaker : public StMaker {
     void FillTTree(); // if debugging ttree is turned on (mGenTree)
     void FitVertex();
 
+<<<<<<< Updated upstream
+=======
```
plexoos: Please remove the lines

```suggestion
```

StRoot/StFwdTrackMaker/StFwdTrackMaker.h
```diff
@@ -182,6 +182,184 @@ class StFwdTrackMaker : public StMaker {
...
+     * @param pvs : if true, smear vertex 
+    */
+    void setSmearMcPrimaryVertex( bool pvs = true ) { mFwdConfig.set<bool>( "TrackFitter.Vertex:smearMcVertex", pvs ); }
+  
+    /**
+     * @brief Sets geometry cache filename
+     * 
+     */
+    void setGeoCache( TString gc ) { mGeoCache = gc; }
+>>>>>>> Stashed changes
```
plexoos:
```suggestion
```

fgeurts:
Hi everybody, what is needed to finalize this PR? I see there are some conflicts(?) that need to be resolved. If this can be done then S&C can wrap this PR into the scheduled SL25a release.

plexoos:
Yes, exactly. Conflicts should be resolved, and I believe @jdbrice is in the best position to handle that.

fgeurts:
Hi @everybody, the recent changes pass all the checks (this was a previous issue). If there are no other concerns with @jdbrice's updates then we should approve.

Grigory @nigmatkulov, Jason @klendathu2k,  and Dmitri @plexoos ?

StRoot/StPicoEvent/StPicoFwdTrack.h
```diff
@@ -0,0 +1,170 @@
...
+ * Description: StPicoFwdTrack stores the Forward tracks built from Fst and Ftt
+ *
+ **************************************************************************/
+#ifndef StPicoFwdTrack_hh
+#define StPicoFwdTrack_hh
+
+#include <TObject.h>
+#include <vector>
+#include "TVector3.h"
+#include "TRefArray.h"
```
plexoos:
```suggestion
```

StRoot/StPicoEvent/StPicoFwdVertex.h
```diff
@@ -0,0 +1,80 @@
...
+ * Description: StPicoFwdVertex stores the Forward tracks built from Fst and Ftt
+ *
+ **************************************************************************/
+#ifndef StPicoFwdVertex_hh
+#define StPicoFwdVertex_hh
+
+#include <TObject.h>
+#include <vector>
+#include "TVector3.h"
+#include "TRefArray.h"
```
plexoos:
```suggestion
```

StRoot/StPicoEvent/StPicoFcsCluster.cxx
```diff
@@ -0,0 +1,39 @@
+#include "StPicoFcsCluster.h"
+#include <vector>
+#include "StPicoMessMgr.h"
+
+// ROOT headers
+#include "TMath.h"
+
+ClassImp(StPicoFcsCluster)
```
plexoos:
```suggestion
```

StRoot/StPicoEvent/StPicoFcsCluster.cxx
```diff
@@ -0,0 +1,39 @@
+#include "StPicoFcsCluster.h"
+#include <vector>
+#include "StPicoMessMgr.h"
```
plexoos:
```suggestion
```

StRoot/StPicoEvent/StPicoFcsCluster.h
```diff
@@ -0,0 +1,87 @@
...
+    Float_t mChi2Ndf1Photon=0.0;  // &chi;<sup>2</sup> / ndf for 1-photon fit
+    Float_t mChi2Ndf2Photon=0.0;  // &chi;<sup>2</sup> / ndf for 2-photon fit
+    Float_t mFourMomentumX=0.0;
+    Float_t mFourMomentumY=0.0;
+    Float_t mFourMomentumZ=0.0;
+    Float_t mFourMomentumT=0.0;
+
+    std::vector<UShort_t> mFwdTrackMatchIndex;//! do not save to disk (in memory only)
+
+    ClassDef(StPicoFcsCluster, 2)
```
plexoos:
```suggestion
    ClassDef(StPicoFcsCluster, 1)
```

StRoot/StPicoEvent/StPicoFcsHit.cxx
```diff
@@ -0,0 +1,26 @@
+#include "StPicoFcsHit.h"
+#include <vector>
+#include "StPicoMessMgr.h"
+
+// ROOT headers
+#include "TMath.h"
+
+ClassImp(StPicoFcsHit)
```
plexoos:
```suggestion
```

jdbrice:
Hi @fgeurts all, what are the next steps on this?

fgeurts:
> Hi @fgeurts all, what are the next steps on this?

Please reach out to the remaining expert reviewers so that the merger can be completed.

StRoot/StPicoDstMaker/StPicoDstMaker.cxx
```diff
@@ -1203,7 +1227,7 @@ void StPicoDstMaker::fillTracks() {
       picoTrk->setGlobalMomentum( momAtDca[0], momAtDca[1], momAtDca[2] );
       picoTrk->setOrigin( pos[0], pos[1], pos[2] );
     }
-    else {
+    else if ( mMuDst->primaryVertex() ){
```
plexoos, line 1235: By changing the selection condition and removing the default else block, do you change the set of tracks saved to the pico DSTs? Would this affect the already produced picoDst files?

StRoot/StPicoEvent/StPicoFwdTrack.h
```diff
@@ -0,0 +1,170 @@
...
+    /// ECal and HCal track projection positions
+    /// These are packed 
+    Short_t mECalX; 
+    Short_t mECalY;
+    Short_t mECalZ;
+    Short_t mHCalX;
+    Short_t mHCalY;
+    Short_t mHCalZ;
+
+    ClassDef(StPicoFwdTrack,5)
```
plexoos:
```suggestion
    ClassDef(StPicoFwdTrack, 1)
```

StRoot/StPicoEvent/StPicoFwdVertex.cxx
plexoos, line 1: Is this file necessary?

StRoot/StPicoDstMaker/StPicoDstMaker.cxx
jdbrice: Hi @plexoos 
so the original code has a bare else but uses `mMuDst->primaryVertex()->position()` inside the else block a few lines lower. So me adding this check on `mMuDst->primaryVertex()` should not change behavior, but protect against a nullptr dereference. In practice it seems it was never a problem in the past though

StRoot/StPicoDstMaker/StPicoDstMaker.cxx
```diff
@@ -1927,7 +1959,7 @@ void StPicoDstMaker::fillEvent() {
   } //for(UInt_t iPMT=0; iPMT<bbc.numberOfPMTs(); ++iPMT)
 
     // Set bunch crossing ID
-  picoEvent->setBunchId( ev->l0Trigger().bunchCrossingId() );
+  picoEvent->setBunchId( ev->l0Trigger().bunchCrossingId7bit(ev->runNumber()) );
```
nigmatkulov, line 1964: What is the difference between the returned values? The first one was requested by the spin PWG a while ago.

StRoot/StBFChain/BigFullChain.h
```diff
@@ -1949,6 +1949,7 @@ Bfc_st BFC[] = { // standard chains
   {"PicoVtxVpdOrDefault","","","-PicoVtxDefault"    ,"","","pico Vtx cut on Tof and VPD or default",kFALSE},
   {"PicoVtxFXT"     ,"","","-PicoVtxDefault"    ,"" ,"","pico Vtx constraint on FXT [198,202] mode",kFALSE},
   {"PicoVtxMtd"     ,"","","-PicoVtxDefault"             ,"" ,"","pico Vtx using MTD matching mode",kFALSE},
+  {"PicoVtxless"    ,"","","-PicoVtxDefault"             ,"" ,"","pico Vtx NOT required for FWD"   ,kFALSE},
```
genevb, line 1952: There needs to be a corresponding line in `StBFChain/StBFChain.cxx` to pass this detail on to `StPicoDstMaker`.

genevb (changes requested):
Review of BFC change: Just one further small modification needed.

plexoos:
Thanks for addressing my suggestions. I don’t have any further comments at this time. It would be great to finally hear from the actual picoDst owner, @nigmatkulov, so we can finalize and merge.

StRoot/StPicoDstMaker/StPicoDstMaker.cxx
nigmatkulov: Hi @jdbrice, I would not change this. It may affect old productions. I've not seen that nullptr would be passed. The idea of that part is that it will store covariance matrix parameters at dca to beamline.

StRoot/StPicoDstMaker/StPicoDstMaker.cxx
```diff
@@ -1828,7 +1856,11 @@ void StPicoDstMaker::fillEvent() {
 
   picoEvent->setGRefMult( ev->grefmult() );
   picoEvent->setNumberOfGlobalTracks( mMuDst->numberOfGlobalTracks() );
-  picoEvent->setNumberOfPrimaryTracks( mMuDst->numberOfPrimaryTracks() );
+  if (mMuDst->numberOfPrimaryVertices() > 0 && mMuDst->primaryTracks() ){ //calling mMudst->numberOfPrimaryTracks() will crash if there are no primary vertices
```
nigmatkulov: I'm not sure that these checks are needed. The numbers returned will be either 0 or the actual number of primary tracks. Why do you need these extra protections? Assuming you have the former part of the if statement, why do you need this mMuDst->primaryTracks()? The primary vertex cannot exist without primary tracks? I would think that most of these checks are extra and will not do much under the current paradigm except causing extra computation time.

StRoot/StPicoDstMaker/StPicoDstMaker.cxx
```diff
@@ -1013,7 +1037,10 @@ void StPicoDstMaker::fillTracks() {
   // don't use StMuTrack::primary(), it returns primary tracks associated with
   // all vertices
   std::unordered_map<unsigned int, unsigned int> index2Primary;
-
+  
+  if (mMuDst->primaryTracks() == nullptr) {
```
nigmatkulov: Do you want skip events without primary tracks? Why?

StRoot/StPicoDstMaker/StPicoDstMaker.cxx
```diff
@@ -2502,6 +2534,173 @@ void StPicoDstMaker::fillMtdHits() {
...
+
+    // fill a map of picodst FCS clusters between index and the cluster pointer
+    map<UShort_t, StPicoFcsCluster*> fcsClusterMap;
+    for ( auto iClu = 0; iClu < mPicoArrays[StPicoArrays::FcsCluster]->GetEntries(); iClu++ ){
+      StPicoFcsCluster * picoFcsCluster = (StPicoFcsCluster*)mPicoArrays[StPicoArrays::FcsCluster]->At(iClu);
+      fcsClusterMap[picoFcsCluster->index()] = picoFcsCluster; 
+    }
+
+    const StSPtrVecFwdTrack& evTracks = evc->tracks();
+    LOG_INFO << "Adding " << evc->numberOfTracks() << " StFwdTracks from StEvent to PicoDst" << endm;
```
nigmatkulov: Here and in the other locations. Do you need all this info to be populating log files? I suggest to move it to LOG_DEBUG.

StRoot/StPicoDstMaker/StPicoDstMaker.cxx
```diff
@@ -2502,6 +2534,173 @@ void StPicoDstMaker::fillMtdHits() {
...
+
+      picoFwdTrack.setMcTruth( evTrack->idTruth(), evTrack->qaTruth() );
+      picoFwdTrack.setVtxIndexRaw( evTrack->vertexIndexRaw() );
+      picoFwdTrack.setGlobalTrackIndex( evTrack->globalTrackIndex() );
+
+      // fill matched ecal and hcal clusters for the track
+      // ecal
+      for ( auto & cluster : evTrack->ecalClusters() ){
+        if ( mMapFcsIdPairIndex.count( make_pair( cluster->detectorId(), cluster->id() ) ) == 0 ){
+          LOG_WARN << "No FCS cluster found for ecal cluster with id " << cluster->id() << " and detectorId " << cluster->detectorId() << endm;
```
nigmatkulov: How often does this call happen?

StRoot/StPicoDstMaker/StPicoDstMaker.cxx
```diff
@@ -2502,6 +2534,173 @@ void StPicoDstMaker::fillMtdHits() {
...
+          continue;
+        }
+        int index = mMapFcsIdPairIndex[ make_pair( cluster->detectorId(), cluster->id() ) ];
+        picoFwdTrack.addEcalCluster( index );
+      }
+      // hcal
+      for ( auto & cluster : evTrack->hcalClusters() ){
+        if ( mMapFcsIdPairIndex.count( make_pair( cluster->detectorId(), cluster->id() ) ) == 0 ){
+          LOG_WARN << "No FCS cluster found for hcal cluster with id " << cluster->id() << " and detectorId " << cluster->detectorId() << endm;
```
nigmatkulov: How often does this call happen?

StRoot/StPicoDstMaker/StPicoDstMaker.cxx
```diff
@@ -2502,6 +2534,173 @@ void StPicoDstMaker::fillMtdHits() {
...
+        picoFwdTrack.setHCalProjection( hcalProj.mXYZ.x(), hcalProj.mXYZ.y(), hcalProj.mXYZ.z() );
+      else
+        picoFwdTrack.setHCalProjection( -999., -999., -999. );
+
+      int counter = mPicoArrays[StPicoArrays::FwdTrack]->GetEntries();
+      picoFwdTrack.setId( counter );
+      new((*(mPicoArrays[StPicoArrays::FwdTrack]))[counter]) StPicoFwdTrack(picoFwdTrack);
+    }
+  } else {
+    LOG_WARN << "Cannot get Fwd Tracks from StEvent" << endm;
```
nigmatkulov: How often does this call happen? Do you really need it? Is it possible that the event will not have FWD tracks? I suggest to move it to LOG_DEBUG

StRoot/StPicoDstMaker/StPicoDstMaker.cxx
```diff
@@ -2502,6 +2534,173 @@ void StPicoDstMaker::fillMtdHits() {
...
+  // get primary vertex from StEvent
+  LOG_DEBUG << "There are " << evt->numberOfPrimaryVertices() << " primary vertices from StEvent" << endm; 
+  for ( size_t i = 0; i < evt->numberOfPrimaryVertices(); i++ ){
+    StPrimaryVertex * evVertex = evt->primaryVertex(i);
+    if (!evVertex || !evVertex->isFwdVtx()) continue; // only save valid FwdVertex
+    StPicoFwdVertex picoFwdVertex;
+    // Set the PicoDst attributes
+    picoFwdVertex.setPosition( evVertex->position().x(), evVertex->position().y(), evVertex->position().z() );
+    LOG_INFO << "Adding StPicoFwdVertex from StPrimaryVertex" << endm;
+    picoFwdVertex.setChi2( evVertex->chiSquared() );
```
nigmatkulov, line 2635: How often does this call happen?

nigmatkulov (changes requested):
Please take a look at the comments. I'm not sure we need such an excessive logging. I suggest to revert part related to covariance matrix and dca geometry. For the FWD classes I suggest to add doxygen documentation.

StRoot/StPicoDstMaker/StPicoDstMaker.cxx
jdbrice: Yes, they are both necessary (at least in practice) because I get a set fault in the rare cases when there IS a primary vertex but no primary tracks. Also, it is necessaery because some of the "numberOf..." functions do not check for null ptr before asking collection for entries.

StRoot/StPicoDstMaker/StPicoDstMaker.cxx
jdbrice: @akioogawa can you comment - since this was an upstream PR from you into my branch

StRoot/StPicoDstMaker/StPicoDstMaker.cxx
jdbrice: this is necessary because the next line:
```
Int_t nPrimarys = mMuDst->numberOfPrimaryTracks();
```
seg faults if the primary tracks array is nullptr. This nullptr check should have always been there but was probably never an issue in practice.

StRoot/StPicoDstMaker/StPicoDstMaker.cxx
jdbrice: I would prefer to leave these at least for first productions, as it would indicate some failures of the algorithm that in theory should not be possible.

StRoot/StPicoDstMaker/StPicoDstMaker.cxx
jdbrice: Same as above, in theory this should never happen, only if the FCS db was somehow failing - but it would indicate to me that e.g. we need to stop and restart the production

StRoot/StPicoDstMaker/StPicoDstMaker.cxx
akioogawa: 1st one is 48bit counter modulo 120, and it is NOT aligned with bunch 0, but has a offset (offset does not change during a run, but change run by run). 2nd one is 7bit counter which is reset by RevTick (=bunch0 signal) and shows correct bunch Id.

StRoot/StPicoDstMaker/StPicoDstMaker.cxx
nigmatkulov: Could you please show such an event? That is really weird and must not happen. It means there is a problem at the MuDst level.

StRoot/StPicoDstMaker/StPicoDstMaker.cxx
nigmatkulov: Could you please show when it happens? MuDst must return 0 in case no primary tracks are found.

StRoot/StPicoDstMaker/StPicoDstMaker.cxx
nigmatkulov: Thank you!

StRoot/StPicoDstMaker/StPicoDstMaker.cxx
nigmatkulov: @genevb, please comment on it if this is okay. Generally speaking, I would assume there should be a small test production to check this

StRoot/StPicoDstMaker/StPicoDstMaker.cxx
nigmatkulov: Do we connect and read the FCS DB when running the conversion from MuDst to picoDst?

StRoot/StPicoDstMaker/StPicoDstMaker.cxx
nigmatkulov: Another way to think about it is that we have events without a primary vertex found. For those, we still want to store the global track covariance matrix. Not just for primary tracks.

fgeurts:
@nigmatkulov and @jdbrice : please let us know when this PR can be wrapped up.

StRoot/StPicoDstMaker/StPicoDstMaker.cxx
jdbrice: change to LOG_DEBUG

StRoot/StPicoDstMaker/StPicoDstMaker.cxx
jdbrice: changed to LOG_DEBUG

StRoot/StPicoDstMaker/StPicoDstMaker.cxx
jdbrice: yes

StRoot/StPicoDstMaker/StPicoDstMaker.cxx
jdbrice: l462 of StMuDst.h

```c++
static unsigned int numberOfPrimaryVertices()  { return arrays[muPrimaryVertex]->GetEntriesFast(); }
static unsigned int numberOfPrimaryTracks()  { return mCurrPrimaryTracks->GetEntriesFast(); }
```

these are unprotected and can set fault if the underlying collections are not initialized

StRoot/StPicoDstMaker/StPicoDstMaker.cxx
jdbrice: the fwd test suite which runs without TPC tracking will trigger this case if not protected

StRoot/StPicoDstMaker/StPicoDstMaker.cxx
nigmatkulov: Can we add protection at the MuDst level?

StRoot/StPicoDstMaker/StPicoDstMaker.cxx
genevb: Compromise: perhaps it would be better to replace this per-cluster message with something along the lines of this:

    int noFCScluster4ECALcluster = 0;
    for ( auto & cluster : evTrack->ecalClusters() ){
      if ( mMapFcsIdPairIndex.count( make_pair( cluster->detectorId(), cluster->id() ) ) == 0 ){
        if (debug()) { LOG_WARN << "No FCS cluster found for ecal cluster with id " << cluster->id() << " and detectorId " << cluster->detectorId() << endm; }
        noFCScluster4ECALcluster++;
      }
    }
    if (noFCScluster4ECALcluster > 0) { LOG_WARN << noFCScluster4ECALcluster << " ECAL clusters w/o FCS cluster" << endm; }

StRoot/StPicoDstMaker/StPicoDstMaker.cxx
genevb: Same suggestion as above: log per-cluster only in debug mode; otherwise 0 or 1 messages per event to flag occurrence.

genevb:
General comment:
If we're worried about memory build-up generated by `LOG_DEBUG` lines, then please consider using the following inside classes that inherit from StMaker:

    if (debug()) { LOG_INFO << blah << endm; }

or `LOG_WARN` as you see fit.
(NB: the curly brackets around `LOG_*` messages after an `if` _are_ _necessary_ because the `LOG_*` are actually macros that resolve to multiple lines of code.)

StRoot/StPicoDstMaker/StPicoDstMaker.cxx
nigmatkulov: Does it give the same result for the old productions? I mean, did you check that with these changes, the new files are identical to the old ones?

StRoot/StPicoDstMaker/StPicoDstMaker.cxx
jdbrice: LOG_ERROR once per event per Gene's suggestion in 77ee867

StRoot/StPicoDstMaker/StPicoDstMaker.cxx
jdbrice: LOG_ERROR once per event per Gene's suggestion in 77ee867

StRoot/StPicoDstMaker/StPicoDstMaker.cxx
jdbrice: This code simply checks a pointer for non-null before using. I removed the unnecessary check on primary vertices in 9949d29.

StRoot/StPicoDstMaker/StPicoDstMaker.cxx
jdbrice: That is opening a can of worms. Could be addressed in a different PR but would take significant time investment.

StRoot/StPicoDstMaker/StPicoDstMaker.cxx
jdbrice: This is testing a pointer for non-null before use. If this changes behavior then it was undefined behavior anyways. This is how the code should have been before. `mMuDst->primaryVertex()` in the next line will segfault (or give undefined behavior) if called when pointer is null

jdbrice: PREVIOUS code was:
```cpp
else {
      StPhysicalHelixD gHelix = dcaG->helix();
      gHelix.moveOrigin( gHelix.pathLength( mMuDst->primaryVertex()->position() ) );
      StThreeVectorF globMom =  gHelix.momentum( mBField * kilogauss );
      StThreeVectorF globOrigin = gHelix.origin();
      picoTrk->setGlobalMomentum( globMom.x(),
				  globMom.y(),
				  globMom.z() );
      picoTrk->setOrigin( globOrigin.x(),
			  globOrigin.y(),
			  globOrigin.z() );
    }
```

StRoot/StPicoDstMaker/StPicoDstMaker.cxx
jdbrice: In [edef3b5](https://github.com/star-bnl/star-sw/pull/639/commits/edef3b5b3062ddcadccf067e45db4341213dd127) I just check for nullptr before getting # of primary tracks. This way it should skip primary track loop, but continue with global tracks.

jdbrice:
Summary (Aug 28 12pm): 
- One "breaking" change was reverted - dont return early from StPicoDstMaker::fillTracks() if muDst->primaryTracks == nullptr. Instead skip only the primary track loop.
- LOG_WARN on unmatched StEvent FCS clusters moved to once per event as LOG_ERROR
- safety checks on pointers (non-null) before use were not removed. This is how the code should have been and prevents segfault / undefined behavior.
All comments adequately addressed AFAIK

jdbrice:
@jdb will check consistency of nightly test output to verify that the picots output is unchanged

nigmatkulov (approved):
Seems to be okay. @genevb will build the library for testing the changes.

genevb:
I built an SL25z library (64-bit, on NFS4 only) with this PR, and ran several nightly tests for comparison with DEV (64-bit, AFS). Some of the tests I ran only produced MuDsts, so I did not have a closer look at them. The following nightly test datasets' PicoDsts were compared:

production_pp500_2022_64bit (optimized)
AuAu200_production_2016_64bit (unoptimized) (see footnote 1)
AuAu200_production_2016_64bit (optimized) (see footnote 1)
production_AuAu_2023_64bit (optimized)
production_AuAu_2024_64bit (optimized)
pp200_production_LowLuminosity_64bit (optimized)
27GeV_production_2018_64bit (unoptimized) (see footnote 1)
27GeV_production_2018_64bit (optimized) (see footnote 1)

I found in all cases (see footnote 2), the PicoDsts were essentially identical. This evidence supports moving forward with this PR.

* Footnote 1: The jobs from years 2016 and 2018 (before the FWD subsystems existed) report these errors:

    StPicoDstMaker::ERROR - Cannot get StFcsDb object
    StPicoDstMaker::ERROR - null FwdTrackCollection

* Footnote 2: The dev unoptimzed 27GeV saw a tiny difference in the number of primary tracks from the other three 27GeV jobs. Total numbers of global tracks were the same, however. So I'm inclined to believe this is a matter of rounding (least significant bits) differences, and not related to this PR.

genevb (approved):
The change I requested was made, so I'm changing my review status to "Approve".

genevb:
...but it would be good if we can resolve the error messages. Perhaps a default DB entry for some table with `beginTime` = `entryTime` = "1996-01-01" would be appropriate?

nigmatkulov:
> ...but it would be good if we can resolve the error messages. Perhaps a default DB entry for some table with `beginTime` = `entryTime` = "1996-01-01" would be appropriate?

@jdbrice, could you please do it? It should be a simple change.

StRoot/StPicoDstMaker/StPicoDstMaker.cxx
```diff
@@ -2502,6 +2536,181 @@ void StPicoDstMaker::fillMtdHits() {
...
+  StFcsDb* fcsDb = static_cast<StFcsDb*>(GetDataSet("fcsDb"));
+  if( !fcsDb ) {
+      LOG_ERROR << "Cannot get StFcsDb object" << endm;
+      return;
+  }
+  StMuFcsCollection * muFcs = mMuDst->muFcsCollection();
+  if ( !muFcs ) {
+    LOG_ERROR << "Cannot get Fcs Collection from MuDst" << endm;
+    return;
+  }
```
genevb: I expect that a better logic is the way it was done for `fillBTowHits()`, where the function to fill isn't called unless the MuDst collection for BTow exists at all. However, getting that improvement in place can be considered a next step if we are rushing to move forward with this, and it can be considered sufficient for now to simply reverse these two checks inside `fillFcsClusters()` with the understanding that if there is no MuDst FCS collection, that's not an error...but if there is an FCS collection, *then* `StFcsDb` is necessary:
```suggestion
  StMuFcsCollection * muFcs = mMuDst->muFcsCollection();
  if ( !muFcs ) return;
  StFcsDb* fcsDb = static_cast<StFcsDb*>(GetDataSet("fcsDb"));
  if( !fcsDb ) {
      LOG_ERROR << "Cannot get StFcsDb object" << endm;
      return;
  }
```

StRoot/StPicoDstMaker/StPicoDstMaker.cxx
```diff
@@ -2502,6 +2536,181 @@ void StPicoDstMaker::fillMtdHits() {
...
+//_________________
+void StPicoDstMaker::fillFwdTracks() {
+
+  StEvent *evt = (StEvent *)GetDataSet("StEvent");
+  if ( evt ){
+    StFwdTrackCollection * evc = evt->fwdTrackCollection();
+    if ( !evc ){
+      LOG_ERROR << "null FwdTrackCollection" << endm;
+      return;
+    }
```
genevb: Here again we have the sub-optimal logic, as `fillFwdTracks()` probably should only be called when the StFwdTrackCollection exists, and it isn't necessarily a problem if that doesn't exist.
```suggestion
    if ( !evc ) return;
```

StRoot/StPicoDstMaker/StPicoDstMaker.cxx
```diff
@@ -2502,6 +2536,181 @@ void StPicoDstMaker::fillMtdHits() {
...
+  StFcsDb* fcsDb = static_cast<StFcsDb*>(GetDataSet("fcsDb"));
+  if( !fcsDb ) {
+      LOG_ERROR << "Cannot get StFcsDb object" << endm;
+      return;
+  }
+  StMuFcsCollection * muFcs = mMuDst->muFcsCollection();
+  if ( !muFcs ) {
+    LOG_ERROR << "Cannot get Fcs Collection from MuDst" << endm;
+    return;
+  }
```
genevb: See my comment for `fillFcsClusters ()`.
```suggestion
  StMuFcsCollection * muFcs = mMuDst->muFcsCollection();
  if ( !muFcs ) return;
  StFcsDb* fcsDb = static_cast<StFcsDb*>(GetDataSet("fcsDb"));
  if( !fcsDb ) {
      LOG_ERROR << "Cannot get StFcsDb object" << endm;
      return;
  }
```

genevb (commented):
The suggestions I propose here deal with the unhelpful error messages for dataset which don't have FCS and/or FWD tracks. The understanding at today's S&C Management meeting was to make these changes for now to complete this PR so that we can proceed with freezing and building a library and move to a production.... but that it would be appropriate for a follow-up PR to more optimally handle these cases. There is code redundancy, and there should be no need for the `fillFcs*()` functions to be called at all if no FCS collection exists in the MuDst.

plexoos (approved):

jdb:
> @jdb will check consistency of nightly test output to verify that the picots output is unchanged

Just a note @jdb is not the right person, I don't participate to this repo. Probably @jdbrice is.

genevb:
> > @jdb will check consistency of nightly test output to verify that the picots output is unchanged
> 
> Just a note @jdb is not the right person, I don't participate to this repo. Probably @jdbrice is.

Hahaha.... What makes that particularly funny is that @jdbrice is the one who wrote @jdb :-)  Anyhow, sorry we bothered you! The authenticity of users included in these github discussions has been a concern of mine since the start  :-/

PR #640, opened 2024-01-31
refactor: move StiCA functionality to Sti package

### LINKED ISSUE #628: BFC tests fail in StiDefaultToolkit when using ROOT 6.24
https://github.com/star-bnl/star-sw/actions/runs/6800427946?pr=619

A number of test jobs (ROOT6.24) failed with the following message:

```
StiMaker:INFO  - TAttr::Found : usetracker = 1

StiKalmanTrackFinder::StiKalmanTrackFinder() - Started
StiKalmanTrackFinder::StiKalmanTrackFinder() - Done
StiMaker:INFO  - TAttr::Found : seedfinders = CA DEF

fatal error: cannot open file 'input_line_583': No such file or directory
root4star: .sl79_gcc485/OBJ/StRoot/Sti/StiDefaultToolkit.cxx:645: virtual StiTrackFinder* StiDefaultToolkit::getTrackSeedFinderCA(): Assertion `mySeed' failed.
```

The offending code appears to be in `StRoot/Sti/StiDefaultToolkit.cxx` where a new seed finder is create via the ROOT interpreter. 

```cpp
640 StiTrackFinder   * StiDefaultToolkit::getTrackSeedFinderCA()                                                             
641 {                                                                                                                        
642   gSystem->Load("TPCCATracker");                                                                                         
643   gSystem->Load("StiCA");                                                                                                
644   auto *mySeed = (StiTrackFinder*)gROOT->ProcessLineFast("StiCALoader::New()");                                          
645   assert(mySeed);                                                                                                        
646   return mySeed;                                                                                                         
647 } 
```

It is not very clear to me why that statement was chosen to be "interpreted" by ROOT rather than simply compiled. It may help to rewrite the getters in this class. Any objections?

plexoos:
Resolves #628


klendathu2k (approved):
This looks fine to me.  It simply moves the interface between the CA track finder and the Sti track fitter into the Sti package... all code that is reasonably 'owned' by Sti.    It also removes the need for loading libraries in code, and the need to rely on root introspection for the creation of new objects.

genevb (approved):
If I understand correctly, CA codes will now get loaded unnecessarily if a chain is run with Sti but not CA (not StiCA), but that rather insignificant use case is likely worth the reduced dependencies this brings moving forward.

PR #642, opened 2024-01-31
refactor: do not clone geometry for Sti detector builder

plexoos:


klendathu2k (approved):
We will need the capability to restore the ROOT geometry after the Sti averaging procedure in order to run any VMC based simulation and tracking in the same chain.

PR #643, opened 2024-01-31
Disable DAQ logging when compiling outside of local domains

jml985:
Disable DAQ logging when compiling outside of local domains

2 defines are added which should be left ignored for most offline compiles:

RTS_ENABLE_LOG
RTS_LOG_DEFAULT_NET

The first will enable logging, and it will go to stderr
The second enables logging to the main daq server, and can not be successful outside the local network.

### DIFF
--- a/StDb/idl/fcsEcalGainOnline.idl
+++ b/StDb/idl/fcsEcalGainOnline.idl
@@ -1,0 +1,10 @@
+/*    fcsEcalGainOnline.idl
+ *
+ *  Table: fcsEcalGainOnline
+ *
+ *       description: // FCS ECal gain online
+ */ 
+
+struct fcsEcalGainOnline {   
+  float gainOnline[1496];     /* gaincorr 2*748 */
+};  
--- a/StDb/idl/fcsHcalGainOnline.idl
+++ b/StDb/idl/fcsHcalGainOnline.idl
@@ -1,0 +1,10 @@
+/*    fcsHcalGainOnline.idl
+ *
+ *  Table: fcsHcalGainOnline
+ *
+ *       description: // FCS HCal gain online 
+ */ 
+
+struct fcsHcalGainOnline {   
+  float gainOnline[520];     /* gaincorr 2*260 */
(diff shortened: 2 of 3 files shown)


plexoos:
@jml985, could you please confirm that all the submitted changes are intended? You can do this by checking the "Files changed" tab at https://github.com/star-bnl/star-sw/pull/643/files. I am asking because there appear to be modifications beyond what the title and descriptions indicate. Are these changes minor enough to be disregarded?

genevb (approved):
I am not an expert on the code, but approving to help it move quickly.

jml985:
@plexoos Yes, there were also some changes here for the daq readers of various detectors during the summer.   These could in principal be separated from this pull request if you would rather keep them independent.

plexoos:
> These could in principal be separated from this pull request if you would rather keep them independent.

Fine by me

plexoos (approved):

genevb:
> @plexoos Yes, there were also some changes here for the daq readers of various detectors during the summer. These could in principal be separated from this pull request if you would rather keep them independent.

A potential reason to separate things is that, as I understand it, we're going to want to back-propagate the DAQ-logging portion of this PR into older software libraries. The FCS, iTPC, TPC23, STGC codes did not exist in old libraries, so this PR as it stands can't simply be used as is for updating old libraries.

plexoos:
Yes, it's a very good point Gene. Jeff, can you prepare a patch which can be applied to older libraries?

jml985:
What is the best way to remove files from the pull request, do we drop the request and make a new one?

plexoos:
Yes, considering there aren't many changes, it might be easier to create a new branch from the head of main, reintroduce only the relevant modifications, and then create a pull request using that new branch. Alternatively, you can use `git revert` followed by `git add -e` or `git add -p` to selectively include the relevant changes and skip the unwanted ones.

jml985:
Ok, I'll do this after my current meeting...

-Jeff

PR #663, opened 2024-03-07
Update centrality definition for Run20 and Run21

lbavinh:
Update centrality definition for:
- FXT: Au+Au sqrt(s_NN) = 5.2, 6.2 GeV
- COL BES-II: Au+Au sqrt(s_NN) = 9.2, 11.5, 17.3 GeV

Temporarily set event weight for sqrt(s_NN) = 3.0 GeV (2018) to unity


nigmatkulov (approved):
LGTM

PR #666, opened 2024-03-13
Add Be8 decays to 2 alphas...

klendathu2k:
https://en.wikipedia.org/wiki/Beryllium-8

### DIFF
--- a/pams/sim/gstar/gstar_part.g
+++ b/pams/sim/gstar/gstar_part.g
@@ -525,8 +525,7 @@ MODULE gstar_part Is the STAR Particle Database
                              trktyp    = kGtHION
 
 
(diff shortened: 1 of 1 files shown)


plexoos (approved):

genevb (commented):
I have no objection, though the words "Alpha clusters decay into Alpha" seem a little odd. Maybe "Alpha clusters decay into individual Alphas"?

Do we know in which (old) libraries this will be needed?

-Gene

klendathu2k:
Updated the comment.  I'll follow up with Jie about which libraries this
commit will need to be added to.  (It should be easier than the number 
of
the PR implies...)

Cheers,
Jason

akioogawa (approved):

PR #668, opened 2024-03-14
Adding FCS online gain used on DEP in offline DB

akioogawa:
Adding FCS online gain used on DEP in offline DB which will be used for offline trigger emulation (StFcsTriggerSimMaker). Offline DB already has tables pushed. Here is code needed to reading of those tables, and using them. Assorted macros to push those into DB and input files are also in here for future reference/updates. Those gainOnline tables will be only useful for trigger emulation in offline, and other analysis only uses gain and gaincorr for offline energy gain calibrations.


StRoot/StFcsDbMaker/StFcsDb.cxx
```diff
@@ -354,8 +373,31 @@ int StFcsDb::InitRun(int runNumber) {
...
+	    getDepfromId(det,id,ehp,ns,crt,slt,dep,ch);
+	    mGainOnline[ehp][ns][dep][ch]=mFcsPresThreshold.threshold[ip];
+	    ip++;
+        }
+    }    
+
     // Get beamline 
-    TDataSet* dbDataSet = StMaker::GetChain()->GetDataBase("Calibrations/rhic/vertexSeed");
+    //TDataSet* dbDataSet = StMaker::GetChain()->GetDataBase("Calibrations/rhic/vertexSeed");
+    TDataSet* dbDataSet = 0;
```
mrosales2016, line 400: Hi Akio, could you please explain the redefinition of the dbDataSet variable here. With this definition the block of code below would not execute. I also don't see this in the code that I copied over from your local directory so I was curious about this change.

akioogawa:
Please someone review this?

veprbl (approved):
Manny says this is good. The question about vertex db unset was marked as resolved by @akioogawa without a response. I assume that is okay, and the change acceptable.

mrosales2016 (approved):

StRoot/StFcsDbMaker/StFcsDb.cxx
akioogawa: This is not currenty used but here as placeholder for when significant crossing angle exist (like pAu, planned but not happening in run24). This cuased some crashes and this is walk around. Yes this looks bad but we need to resolve this when we needed.

akioogawa:
Sorry somehow my comment was not recorded and resolved without comments. I now added back my comment. Need one more reviwer please.

StRoot/StFcsDbMaker/StFcsDb.cxx
mrosales2016: Great that addresses my question thank you for the update.

veprbl:
@plexoos , please

zlchang (approved):

PR #669, opened 2024-03-20
Proper dependencies for FCS Pi0 used in QA

genevb:
This resolves the issue brought in PR #437 :
St_QA_Maker should not directly depend on MuDst (a solution proposed in #437 ), but instead should depend on the StFcsPi0FinderForEcal, which itself has a dependency on MuDst.

### DIFF
--- a/StRoot/StBFChain/BigFullChain.h
+++ b/StRoot/StBFChain/BigFullChain.h
@@ -1639,6 +1639,7 @@ Bfc_st BFC[] = { // standard chains
    "StFcsClusterMaker","StFcsClusterMaker","Fill FCS clusters",                                     kFALSE},
   {"fcsPoint"   ,"","fcsChain", "StEvent,fcsDb",
    "StFcsPointMaker","StFcsPointMaker,libMinuit","Fill FCS points",                                 kFALSE},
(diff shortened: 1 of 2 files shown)


StRoot/StBFChain/BigFullChain.h
```diff
@@ -1639,6 +1639,7 @@ Bfc_st BFC[] = { // standard chains
    "StFcsClusterMaker","StFcsClusterMaker","Fill FCS clusters",                                     kFALSE},
   {"fcsPoint"   ,"","fcsChain", "StEvent,fcsDb",
    "StFcsPointMaker","StFcsPointMaker,libMinuit","Fill FCS points",                                 kFALSE},
+  {"fcsPi0Libs","", "", "MuDst", "", "StFcsPi0FinderForEcal", "Libs for FCS Pi0 Finder",            kFALSE},
```
klendathu2k, line 1651: FCS pi0 finder uses the MuDst for input rather than StEvent.

genevb: It seems an odd mix of MuDst and StEvent: it appears to require StEvent for its Make() member function, but then may call some MuDst static functions later in that member function. It's likely that these dependencies can be further reduced/improved, but for now I'm trying to ensure St_QA_Maker's dependencies, as that's my responsibility.

plexoos (approved):

akioogawa (approved):

PR #671, opened 2024-03-22
First chains for 2024 (no ETOF)

genevb:
Initial chains for pp200 in 2024, needed for FastOffline coming in just a few weeks. Another update with a chain for AA processing may be needed later if it becomes certain that we will acquire AuAu200 data.


klendathu2k (approved):
LGTM

plexoos (approved):

PR #644, opened 2024-02-01
Disable logging outside of local domain, with no other changes

jml985:
Clean version of the pull request disabling DAQ logging when compiling outside of the local domain.

PR #673, opened 2024-03-25
RtsLoggingUpdate

jml985:
updates to allow RTS logging in Jevp while leaving it disabled in general offline computing

### DIFF
--- a/OnlTools/Jevp/level.source
+++ b/OnlTools/Jevp/level.source
@@ -1,4 +1,7 @@
 # STAR LEVEL for EVP code
 
-starver SL23b
+starver SL23c 
 #starver SL21b
+
+
+unsetenv DB_SERVER_LOCAL_CONFIG
--- a/mgr/Conscript-standard
+++ b/mgr/Conscript-standard
@@ -197,7 +197,8 @@ if ( $pkg eq "Jevp") {
 	      .   $main::PATH_SEPARATOR . "#OnlTools/Jevp/StJevpBuilders"
 	      .   $main::PATH_SEPARATOR . "#OnlTools/Jevp/StJevpViewer"
 	      .   $main::PATH_SEPARATOR .  $CPPPATH;
-    $CPPFLAGS .= " -DQT_THREAD_SUPPORT -DQT_SHARED -DQT_NO_DEBUG ";
+    $CPPFLAGS .= " -DQT_THREAD_SUPPORT -DQT_SHARED -DQT_NO_DEBUG -DRTS_ENABLE_LOG ";
+
 }
 
 #$CPPFLAGS .= " -DNEW_DAQ_READER -D__NO_STRANGE_MUDST__  ";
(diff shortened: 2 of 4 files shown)


mgr/Conscript-standard
```diff
@@ -1432,7 +1433,7 @@ if ( $#src > -1 ) {
       }
     }
     if ($pkg eq "RTS") {
-      my $cppflags = "-DRTS_PROJECT_STAR -DTPXREADER -DRTS_LITTLE_ENDIAN";
+      my $cppflags = "-DRTS_PROJECT_STAR -DTPXREADER -DRTS_LITTLE_ENDIAN ";
```
klendathu2k, line 1436: Just making a note that the only difference in this line appears to be whitespace at the end of `$cppflags`.

klendathu2k (approved):
I don't see any issues with cons.  Internal changes to the logger look to me like code cleanup, so no comments there.

OnlTools/Jevp/level.source
```diff
@@ -1,4 +1,7 @@
 # STAR LEVEL for EVP code
 
-starver SL23b
+starver SL23c
```
genevb, line 3: SL23b and SL23c are equivalent libraries, so this change has no effect.

mgr/Conscript-standard
```diff
@@ -197,7 +197,8 @@ if ( $pkg eq "Jevp") {
 	      .   $main::PATH_SEPARATOR . "#OnlTools/Jevp/StJevpBuilders"
 	      .   $main::PATH_SEPARATOR . "#OnlTools/Jevp/StJevpViewer"
 	      .   $main::PATH_SEPARATOR .  $CPPPATH;
-    $CPPFLAGS .= " -DQT_THREAD_SUPPORT -DQT_SHARED -DQT_NO_DEBUG ";
+    $CPPFLAGS .= " -DQT_THREAD_SUPPORT -DQT_SHARED -DQT_NO_DEBUG -DRTS_ENABLE_LOG ";
```
genevb, line 200: Jevp gets compiled when anyone checks out and compiled this `star-sw` repository (including official library compilations at SDCC currently, e.g. $STAR/.sl73_gcc485/LIB/libJevp.so). Does this mean the logging will be enabled within Jevp for everyone? Is that the desired effect?

jml985: regarding SL23b/SL23c I don't mind if it has no effect.  it works for SL23c and it was the newest so its what I chose.

regarding the desired effect.  Yes, that is the desired effect.   The server is a rather special case, I wouldn't expect anyone to attempt to run it other than on the official servers.    For the builders, which users do run while debugging, the default is set to write to stderror, so in order to see any output the users need to have logging enabled.

jml985: Is there a way to set a define easily at the compile step?   via a cons flag or environment variable?   If so, I would be happy to use such a mechanism, rather than set it globally.

plexoos (approved):
Hopefully, these changes were tested.

Seems like the new flag enables the logging logic introduced in https://github.com/star-bnl/star-sw/pull/644

PR #682, opened 2024-04-15
StFttDbMaker: Warning when DB not found

jdbrice:
Checks if databases are returned and issues a warning if not (instead of set faulting)
Warning message provides helpful info about the name of the local file (debug only) to be used instead of the DB


plexoos (approved):
LGTM

veprbl (approved):

PR #686, opened 2024-05-16
Provide manual timecut control for fast offline production and calibr…

jdbrice:
…ation

This allows the StFttClusterMaker to be set to manually override DB timeline cuts for processing fast offline and for running calibrations

### DIFF
--- a/StRoot/StFttClusterMaker/StFttClusterMaker.cxx
+++ b/StRoot/StFttClusterMaker/StFttClusterMaker.cxx
@@ -86,6 +86,7 @@ StFttClusterMaker::Finish()
 Int_t
 StFttClusterMaker::Make()
 { 
+    LOG_INFO << "StFttClusterMaker::Make" << endm;
     mEvent = (StEvent*)GetInputDS("StEvent");
     if(mEvent) {
     } else {
@@ -210,16 +211,26 @@ void StFttClusterMaker::InjectTestData(){
 
 
 bool StFttClusterMaker::PassTimeCut( StFttRawHit * hit ){
-    int time_cut0 = -999;
-    int time_cut1 =  999;
-    int time_cutm = 0;
-    //  in principle it could vary VMM to VMM;
-    mFttDb->getTimeCut(hit, time_cutm, time_cut0, time_cut1);
-    if ( time_cutm == 0 ) // default, cut on bunch crossing
-        return (hit->time() <= time_cut1 && hit->time() >= time_cut0); 
+    
+    if ( mTimeCutMode == -1 ){
+        int time_cut0 = -999;
+        int time_cut1 =  999;
+        int time_cutm = 0;
+        //  in principle it could vary VMM to VMM;
+        mFttDb->getTimeCut(hit, time_cutm, time_cut0, time_cut1);
+        if ( time_cutm == 0 ) // default, cut on bunch crossing
+            return (hit->time() <= time_cut1 && hit->time() >= time_cut0); 
 
-    // cut on timebin
-    return (hit->tb() <= time_cut1 && hit->tb() >= time_cut0);
+        // cut on timebin
+        return (hit->tb() <= time_cut1 && hit->tb() >= time_cut0);
+    } else { // manually set, use that
+        if ( mTimeCutMode == 0 ){
+            return (hit->time() <= mTimeCutMax && hit->time() >= mTimeCutMin); 
+        }
+        else if ( mTimeCutMode == 1 ){
+            return (hit->tb() <= mTimeCutMax && hit->tb() >= mTimeCutMin); 
+        }
+    }
 }
 
 
--- a/StRoot/StFttClusterMaker/StFttClusterMaker.cxx
+++ b/StRoot/StFttClusterMaker/StFttClusterMaker.cxx
@@ -86,7 +86,6 @@ StFttClusterMaker::Finish()
 Int_t
 StFttClusterMaker::Make()
 { 
-    LOG_INFO << "StFttClusterMaker::Make" << endm;
     mEvent = (StEvent*)GetInputDS("StEvent");
     if(mEvent) {
     } else {
--- a/StRoot/StFttClusterMaker/StFttClusterMaker.cxx
+++ b/StRoot/StFttClusterMaker/StFttClusterMaker.cxx
@@ -14,7 +14,7 @@
 #include <map>
 #include <array>
 #include <algorithm>    // std::is_sorted
-
+#include <climits>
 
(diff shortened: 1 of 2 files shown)


StRoot/StFttClusterMaker/StFttClusterMaker.h
```diff
@@ -45,12 +51,18 @@ class StFttClusterMaker: public StMaker {
     float GetThresholdFor( StFttRawHit * hit ) { return 0.0;}
     bool PassTimeCut( StFttRawHit * hit );
 
+
+
```
plexoos:
```suggestion
```

StRoot/StFttClusterMaker/StFttClusterMaker.h
```diff
@@ -45,12 +51,18 @@ class StFttClusterMaker: public StMaker {
...
     StEvent*             mEvent;
     StFttCollection*     mFttCollection;
     int                  mRunYear;
     bool                 mDebug;
     StFttDb*             mFttDb;
 
+    int mTimeCutMin  = -999;
+    int mTimeCutMax  =  999;
+    int mTimeCutMode = -1; //default - lookup from DB
+
```
plexoos:
```suggestion
```

StRoot/StFttClusterMaker/StFttClusterMaker.h
```diff
@@ -30,6 +30,12 @@ class StFttClusterMaker: public StMaker {
     int  Finish();
     int  Make();
 
+    void SetTimeCut( int mode, int min, int max ) {
```
genevb, line 33: If one sets the mode value to -1 via this function, then min and max are meaningless. I suggest adding some defaults for those, e.g. min=-999, max=999.

StRoot/StFttClusterMaker/StFttClusterMaker.cxx
```diff
@@ -210,16 +210,26 @@ void StFttClusterMaker::InjectTestData(){
...
+        // cut on timebin
+        return (hit->tb() <= time_cut1 && hit->tb() >= time_cut0);
+    } else { // manually set, use that
+        if ( mTimeCutMode == 0 ){
+            return (hit->time() <= mTimeCutMax && hit->time() >= mTimeCutMin); 
+        }
+        else if ( mTimeCutMode == 1 ){
+            return (hit->tb() <= mTimeCutMax && hit->tb() >= mTimeCutMin); 
+        }
+    }
```
genevb, line 255: This function has no code to handle an `mTimeCutMode` value outside of {-1,0,1}. At the very least, there should be a `LOG_WARN` or `LOG_ERROR` for other values, though I'm not sure what the function return value should be in that case (false?).

StRoot/StFttClusterMaker/StFttClusterMaker.h
```diff
@@ -45,12 +51,18 @@ class StFttClusterMaker: public StMaker {
...
     int                  mRunYear;
     bool                 mDebug;
     StFttDb*             mFttDb;
 
+    int mTimeCutMin  = -999;
+    int mTimeCutMax  =  999;
+    int mTimeCutMode = -1; //default - lookup from DB
+
 
     ClassDef( StFttClusterMaker, 1 )
```
genevb: This should probably have a version = 0, as the class isn't meant to be persistent. Otherwise we'd be incrementing from version 1 to 2 because you've added new data members. But 0 is better in such a case as this.

genevb (commented):
A few simple fixes suggested.

StRoot/StFttClusterMaker/StFttClusterMaker.cxx
plexoos: Here is a slightly improved version of this method (as suggested by a GPT). It gets rid of the nested if-statements and introduces a default `return`. Feel free to refactor as needed:

```cpp
bool StFttClusterMaker::PassTimeCut(StFttRawHit* hit) {
    const int defaultTimeCutMin = -999;
    const int defaultTimeCutMax = 999;

    if (mTimeCutMode == -1) {
        int timeCutMin = defaultTimeCutMin;
        int timeCutMax = defaultTimeCutMax;
        int timeCutMode = 0;

        mFttDb->getTimeCut(hit, timeCutMode, timeCutMin, timeCutMax);

        if (timeCutMode == 0) {
            return (hit->time() >= timeCutMin && hit->time() <= timeCutMax);
        } else {
            return (hit->tb() >= timeCutMin && hit->tb() <= timeCutMax);
        }
    } else if (mTimeCutMode == 0) {
        return (hit->time() >= mTimeCutMin && hit->time() <= mTimeCutMax);
    } else if (mTimeCutMode == 1) {
        return (hit->tb() >= mTimeCutMin && hit->tb() <= mTimeCutMax);
    } else {
        return false; // Default return value if no conditions are met
    }
}
```

StRoot/StFttClusterMaker/StFttClusterMaker.h
```diff
@@ -45,12 +51,18 @@ class StFttClusterMaker: public StMaker {
...
+
+
     StEvent*             mEvent;
     StFttCollection*     mFttCollection;
     int                  mRunYear;
     bool                 mDebug;
     StFttDb*             mFttDb;
 
+    int mTimeCutMin  = -999;
+    int mTimeCutMax  =  999;
```
plexoos: You may want to introduce named constants at the class level which would also help to document your intention:

```cpp
const int defaultTimeCutMin = -999;
const int defaultTimeCutMax = 999;
```

There are standard named constants `INT_MIN` and `INT_MAX` defined by the language https://en.cppreference.com/w/cpp/types/climits They should be preferred over magic numbers.

StRoot/StFttClusterMaker/StFttClusterMaker.h
jdbrice: @genevb - fixed in [aa226e4](https://github.com/star-bnl/star-sw/pull/686/commits/aa226e47c51be7cc84e9244929888f382c26deeb)

StRoot/StFttClusterMaker/StFttClusterMaker.h
jdbrice: @plexoos Thanks for all of the suggestions. I have made the code (IMHO) much clearer both with your suggestions and a few other changes. 
please look at [aa226e4](https://github.com/star-bnl/star-sw/pull/686/commits/aa226e47c51be7cc84e9244929888f382c26deeb)

akioogawa (approved):

plexoos (approved):

PR #688, opened 2024-06-20
updates to reader --- mainly to fix iTPC fee_words=0 problem

jml985:


StRoot/RTS/src/DAQ_TPC23/itpc23.cxx
```diff
@@ -1371,6 +1391,8 @@ itpc23::itpc23()
 
 	fmt = 0 ;
 
+	fee_words = 0 ;
```
genevb, line 1401: This initialization is for me the critical part as an uninitialized `fee_words` was causing random issues for reconstructing iTPC data.

StRoot/RTS/src/DAQ_TPC23/itpc23.cxx
```diff
@@ -718,7 +719,7 @@ u_int *itpc23::fee_non_trgd(u_int *start)
 	int fee_words = 0 ;
```
genevb, line 726: Jeff and I have noted to Tonko that this local `fee_words` should perhaps not be local, but we believe this is not worth holding up this commit as even without any change to this line, this commit is an important step forward.

genevb (approved):
This commit is important for FastOffline for the current data-taking at the experiment. There are many modifications I know nothing about, but I have commented on the items I identified which spurred Tonko to change the code, and consequentially Jeff to make this commit. I will approve from that perspective. Thanks, Jeff, for helping get this in.

StRoot/RTS/src/TRG_FCS/fcs_trg_base.cxx
akioogawa, line 1: I want to acceess this dsmout later in Trigger QA, so added to .h file, together with all intermediate sums. Can this stay to be none-local variable defined at bottom of fcs_trg_base.h?

StRoot/RTS/src/TRG_FCS/fcs_trg_base.cxx
```diff
@@ -540,7 +540,7 @@ int fcs_trg_base::end_event()
 
 	verify_event_io() ;	// verify interconnectivity 
 
-	//int dsmout = 0;  moved to .h file
+	int dsmout = 0;
```
plexoos: Akio's concern seems legitimate https://github.com/star-bnl/star-sw/pull/688/files#r1651407708
This new definition should not be added since it shadows a class member with the same name that was introduced about 4 years ago: https://github.com/star-bnl/star-sw/commit/4f5a8041d1f5a20c62042bce00c0c976a2765ae7?diff=unified&w=0#diff-dd89e161b0be358807922c6c11eac6634ce3fcbfbb642e31f914a88942cd28baR257

plexoos (approved):

StRoot/RTS/src/TRG_FCS/fcs_trg_base.cxx
plexoos: @jml985 @tonko-lj can you follow up?

jml985: Nobody seems to know what's going on here.   Obviously, this code is wrong because the local variable shadows the class variable.    As far as I can tell this has been true since the FCS came online.   It runs in the FCS though, and I think its Akio's code.   Akio, are you asking me to change this?   Are you asking Tonko?

jml985:
I talked to Akio and resolved the dsmout in the fcs_trg_base,    This should now be ready to go.

PR #689, opened 2024-06-24
patched FTT table for Daniel

dmarkh:
The patched version of the FTT table as requested by Daniel. Note that the table name is "fttDataWindowsB" now, please update FTT codes to use it instead of "fttDataWindows" one.


akioogawa (approved):

jdbrice (approved):
This is due to my mistake in original table - octet did not have enough range to hold the uuid values of 1-384 

Thanks for quick response Dmitri

PR #693, opened 2024-06-28
No StMaker functions if no StMaker loaded

genevb:
In ROOT6, it isn't enough to just hide a class's functions inside an `if`
clause checking the corresponding TClass's existence when the class isn't
loaded. This can be demonstrated currently in SL24x by simply executing:
```
root4star -b -l -q
```
...which produces an error when executing `rootlogoff.C` at quit. Maybe there
are other solutions, but this simple proposal breaks `rootlogoff.C` into two parts,
not even loading the second part unless the first part is true
- test the existence of the StMaker class
- execution of StMaker functions

I welcome other proposals.


StRoot/macros/rootlogoff.C
```diff
@@ -1,10 +1,7 @@
 {
   if (TClassTable::GetDict("StMaker"))
   {
-    StMaker* mk = StMaker::GetChain();
-    if (mk) {
-      mk->Finish();
-    }
+    gROOT->Macro("rootlogoff2.C");
```
genevb, line 4: This (`gROOT->Macro()`) seems to be a more protective wrapper around `gInterpreter->ExecuteMacro()`, but I don't know that it's really any better in this circumstance. But if anyone else has a perspective on which is better to use, please chime in.

klendathu2k: I can't think of a simpler solution here, short of modifying our root4star application.  But that means we lose this functionality when bfc is run under root.exe.  If we can register a callback with the root ".q" command, then we would be in business.  But a very quick search didn't turn up anything.

klendathu2k (approved):
Bottom line... LGTM.

genevb:
A ping on this PR. Not urgent, but if no one has another idea, it would be good to get a second approval and get this out of the way.

iraklic (approved):

PR #698, opened 2024-07-29
Update StEvent structs for StFttPoint StFttCluster and update StFttFastSimMaker

jdbrice:
1. Updates StEvent containers for Ftt data (clusters and points) to include idTruth (mirror FST and FCS)
2. Update StFttFastSimMaker to copy Ftt GEANT hits to StFttCollection as StFttPoints

The Fast sim just copies GEANT hits from front modules and writes them into the StFttCollection as StFttPoints (skips the entire reco chain). The "slow" sim (still WIP) will write clusters and therefore idTruth is needed there.
Removed old code based on square modules (RnD geometry) and different strip layout. This also simplifies the Fwd tracking because it no longer needs to read directly from GEANT

### DIFF
--- a/StRoot/StEvent/StFttCluster.h
+++ b/StRoot/StEvent/StFttCluster.h
@@ -27,6 +27,7 @@ public:
     float sumAdc() const;
     float x() const;  // Mean x ("center of gravity") in local grid coordinate (1st moment).
     float sigma() const; // Maximum 2nd moment (along major axis).
+    UShort_t idTruth() const { return mIdTruth; } // Get the truth ID
 
     void setId(int cluid);
     void setPlane(UChar_t plane);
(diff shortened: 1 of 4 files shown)


plexoos (commented):
I see lots of code removed in this PR. Has it been moved somewhere or not needed anymore?

jdbrice:
@plexoos Good question. The original StFttFastSimMaker was created before we had finalized the design of the detector. It was no longer useful with the final design so I deleted it. 
We are working on a "slow" simulator (just means more realistic than the fast-simulator) which accomplishes the same thing as the deleted code - to simulate the detector at a lower level approximating real data.

klendathu2k (approved):
Looks good to me.

akioogawa (approved):

PR #700, opened 2024-07-30
Update StFstFastSimMaker to write hits into the StFstHitCollection

jdbrice:
as is done for the real data reco chain. I continue to write hits into the StRnDCollection, but this is no longer used by downstream code. 

Also add a function to StFstHit to allow setting the disk, wedge, and sensor for sim hits. Does not change the StFstHit data structure.

### DIFF
--- a/StRoot/StFstSimMaker/StFstFastSimMaker.cxx
+++ b/StRoot/StFstSimMaker/StFstFastSimMaker.cxx
@@ -59,7 +59,6 @@ StFstFastSimMaker::StFstFastSimMaker(const Char_t *name)
     mNumR{8},
     mNumPHI{128},
     mNumSEC{12},
-    mRaster{0},
     mInEff{0},
     mHist{false},
 	mGEANTPassthrough{false},
@@ -154,19 +153,9 @@ void StFstFastSimMaker::FillSilicon(StEvent *event) {
 	const int MAXR = mNumR;
 	const int MAXPHI = mNumPHI * mNumSEC;
 
-	float X0[] = {0, 0, 0, 0, 0, 0};
-	float Y0[] = {0, 0, 0, 0, 0, 0};
-
 	if ( mGEANTPassthrough ){
 		LOG_INFO << "FST Hits using GEANT xyz directly (no raster etc.)" << endm;
 	}
-	
-	if (mRaster > 0)
-		for (int i = 0; i < 6; i++) {
-			X0[i] = mRaster * TMath::Cos(i * 60 * TMath::DegToRad());
-			Y0[i] = mRaster * TMath::Sin(i * 60 * TMath::DegToRad());
-		}
-
 	
 	// maps for hit and energy for each disk's r-phi strip
 	std::map< FstGlobal::FstKeyTriple, StRnDHit* > hitMap;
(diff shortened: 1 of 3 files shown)


plexoos:
The test error does not seem to be related to the code changes:

```
docker: Error response from daemon: Get "https://ghcr.io/v2/": net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers).
See 'docker run --help'.
Error: Process completed with exit code 125.
```

Looks like a network glitch. I just restarted the failed job to see if it passes this time

StRoot/StFstSimMaker/StFstFastSimMaker.cxx
```diff
@@ -221,10 +237,13 @@ void StFstFastSimMaker::FillSilicon(StEvent *event) {
 		double xc = X0[disk_index];
 		double yc = Y0[disk_index];
 
+		// This z-offset is used to shift the hits
+		// to the center of the FST where the tracking planes are defined
+		const double z_delta = 1.755;
```
plexoos, line 227: Any chance this value can be extracted from the geometry?

plexoos (approved):

StRoot/StFstSimMaker/StFstFastSimMaker.cxx
```diff
@@ -144,6 +156,10 @@ void StFstFastSimMaker::FillSilicon(StEvent *event) {
 
 	float X0[] = {0, 0, 0, 0, 0, 0};
 	float Y0[] = {0, 0, 0, 0, 0, 0};
+
+	if ( mGEANTPassthrough ){
+		LOG_INFO << "FST Hits using GEANT xyz directly (no raster etc.)" << endm;
+	}
 	
 	if (mRaster > 0)
```
klendathu2k: Is there still a need to randomize the X,Y center of each plane?

jdbrice: I am not sure what you mean? Maybe you mean blur the position of the hits according to a resolution - thats not a bad idea. I can add that, but please let me know if you meant something different

klendathu2k: It looks like I commented on the wrong line of code.  Just so that we're on the same page, I'm referring to lines 157 and 158, and the if ( mRaster>0 ) block on line 164.  

This code appears suspect to me.  It doesn't smear the hit positions.  It systematically shifts the (x,y) center of each disk to the vertices of a hexagon of 'radius' mRaster.  As I recall, the idea was to provide a simple way to introduce a small stereo angle between the different planes, to improve tracking resolution during design studies.  The class defaults to mRaster=0, so... it probably hasn't caused a problem.  But it looks to me to be dead code that no longer serves a purpose, and thus should be cleaned up.  Time being at a premium... at a minimum I would eliminate any setter methods.

jdbrice: Thanks - this is taken care of in d8801c1. I removed the options to set Raster and removed the additional logic in the main code.

jdbrice: @klendathu2k if you are good with this can you approve?

jdbrice:
@klendathu2k - does this look good to you now?
If so please approve and I'll merge

akioogawa (approved):

PR #701, opened 2024-08-02
Prevent ROOT6/CLING from loading all TROOT exposed classes twice... 

klendathu2k:
… (which couldn't possibly cause any side effects, but lets be safe about it shall we?)


iraklic (approved):
I am more of a `#ifndef` person but `#if !defined` works too :)

plexoos (approved):

genevb (approved):

PR #702, opened 2024-08-08
💥DO NOT MERGE💥 ❗ add new TPC alignment, hit errors and slewing corrections for 2024

fisyak:
new TPC Alignment 24
new TPC hit errors and slewing correction
The detailed description can be found at 
https://drupal.star.bnl.gov/STAR/system/files/TPC24c.Export.Mini_.pdf


StRoot/StBFChain/BigFullChain.h
```diff
@@ -213,9 +213,9 @@ Bfc_st BFC[] = { // standard chains
    ,                                                                                       "","","",kFALSE},
   {"RC.pp.y2005"     ,"","","pp2005a,tofdat,OSpaceZ2,OGridLeak3D"                         ,"","","",kFALSE},
   {"RC.pp.y2006"     ,"","","pp2006b,OSpaceZ2,OGridLeak3D"                                ,"","","",kFALSE},
-  {"RC.y2007"        ,"","","DbV20080418,B2007g,IAna,KeepSvtHit,hitfilt,VFMinuit3,emcDY2,ftpc,trgd,"
```
genevb: Many old chains are modified in this PR. Please leave old chains as they are.

StRoot/StBFChain/BigFullChain.h
```diff
@@ -1192,6 +1196,7 @@ Bfc_st BFC[] = { // standard chains
   {"useLDV" ,""  ,"","",""                                   ,"","... uses laserDV database flavor",kFALSE},
   {"useCDV" ,""  ,"","",""                                       ,"","... uses ofl database flavor",kFALSE},
   {"useNewLDV" ,""  ,"","",""                                    ,"","... uses ofl database flavor",kFALSE},
+  {"TFGdbOpt" ,""  ,"","",""                ,"","... uses TFG database flavor for alignemnt tables",kFALSE},
```
genevb, line 1247: Does this imply that TFG now has distinct database entries in addition to distinct software? It is unclear to me why we need this.

StRoot/StBFChain/BigFullChain.h
```diff
@@ -1134,8 +1135,11 @@ Bfc_st BFC[] = { // standard chains
    ,                                                     "... Corr4+SCEbyE,OGridLeak3D,OSpaceZ2...",kFALSE},
   {"CorrX"       ,"" ,"","ExB,OShortR,OBmap,OPr13,OIFC,OSectorAlign,NewTpcAlignment"
    ",-AlignSectors,-OBmap2D,-OClock,-OTwist"                             ,"","","New Tpc Alignment",kFALSE},
-  {"CorrY"       ,"" ,"","ExB,OShortR,OBmap,OPr40,OIFC,OSectorAlign,NewTpcAlignment,OSpaceZ2,OGridLeakFull"
+  {"CorrY"       ,"" ,"","ExB,OShortR,OBmap,OPr40,-OPr13OIFC,OSectorAlign,NewTpcAlignment,OSpaceZ2,OGridLeakFull"
    ",-AlignSectors,-OBmap2D,-OClock,-OTwist"                             ,"","","New Tpc Alignment",kFALSE},
+  {"Alignment2024","" ,"","NewTpcAlignment"                         ,"","","Turn on Alignment 2024",kFALSE},
+  {"CorrZ","" ,"","Alignment2024,TFGdbOpt,-CorrY,-CorrX,ExB,OShortR,OBmap,OPr40,-Opr13,OIFC,OSpaceZ2"
```
genevb, line 1189: We can introduce a new `CorrZ` that is different from `CorrY` in that it uses `Alignment2024`, but as far as I know, that should be the only difference.

genevb (changes requested):
I only looked at the BFC-related changes, but I think there are some unrelated inclusions from FWD updates.

What is `Bfc.xml`?

If this PR gets through approval, we may want it on a different branch from Main initially so that it can be further vetted before exposing FastOffline to it.

StRoot/StBFChain/BigFullChain.h
genevb: I see this note in the documentation:
"TFGdbOpt to provide a switch between new and old Alignment DB tables."

If this is the purpose, then we should not be using "TFG" in the chain option, nor in the database entries' flavors (production should never depend on anything labeled as developmental). Something more similar/related to "Alignment2024" is more appropriate. And if this is only meant to pair with the new `Alignment2024` chain option, why not directly include the actions as all part of `Alignment2024`?

Bfc.xml
plexoos, line 1: This file should not be placed in the root of this repo. Consider deleting it or moving to StRoot/StStarLogger/

plexoos (changes requested):

StRoot/StBFChain/BigFullChain.h
fgeurts: Hi, I agree that we should not modify old chains. Yuri, can you correct this?

fisyak:
Hi Frank, because for last ~10 year we don’t store StEvent on HPSS the “hitfilt” option has no sense but just create confusion.
I insist to remote it at all.


                                                              Yuri Fisyak

STAR                                           Phone: +1 631 344 3913
Brookhaven National Laboratory  Fax:    +1 631 344 4206
510A/1-161
http://www.star.bnl.gov/~fisyak       E-mail: ***@***.******@***.***>



From: Frank Geurts ***@***.***>
Date: Tuesday, August 20, 2024 at 11:56 AM
Cc: Fisyak, Yuri V ***@***.***>, Author ***@***.***>
Subject: Re: [star-bnl/star-sw] Tfg24c.export.mini (PR #702)

@fgeurts commented on this pull request.

In StRoot/StBFChain/BigFullChain.h<https://github.com/star-bnl/star-sw/pull/702#discussion_r1723562994>:

> @@ -213,9 +213,9 @@ Bfc_st BFC[] = { // standard chains

    ,                                                                                       "","","",kFALSE},

   {"RC.pp.y2005"     ,"","","pp2005a,tofdat,OSpaceZ2,OGridLeak3D"                         ,"","","",kFALSE},

   {"RC.pp.y2006"     ,"","","pp2006b,OSpaceZ2,OGridLeak3D"                                ,"","","",kFALSE},

-  {"RC.y2007"        ,"","","DbV20080418,B2007g,IAna,KeepSvtHit,hitfilt,VFMinuit3,emcDY2,ftpc,trgd,"

Hi, I agree that we should not modify old chains. Yuri, can you correct this?

StRoot/StDbUtilities/StTpcCoordinate.cxx
```diff
@@ -7,6 +7,7 @@
 #include "StTpcLocalSectorAlignedCoordinate.hh"
 #include "StTpcLocalSectorDirection.hh"
 #include "StTpcLocalSectorCoordinate.hh"
+#include "Stiostream.h"
```
plexoos, line 10: Why do we need to include this extra header file?

It essentially includes the following three header files:

```cpp
#include <fstream>
#include <iostream>
#include <iomanip>
```

StRoot/StDbUtilities/StTpcCoordinate.cxx
```diff
@@ -21,6 +22,8 @@ ostream& operator<<(ostream& os, const StTpcCoordinate& a) {
   return os << OS;
 }
 //________________________________________________________________________________
+void StTpcCoordinate::Print(Option_t *option) const {cout << *this << endl;}
```
plexoos, line 25: Does the use of cout and endl prompt the inclusion of `Stiostream.h`? If so, just add `using std::cout; using std::endl;` somewhere in this file

StRoot/StBFChain/BigFullChain.h
plexoos: These changes have been merged in #709. Does this imply that `StHitFilterMaker` was never actually amending the output?

StRoot/Sti/StiHit.cxx
```diff
@@ -147,15 +148,20 @@ static int nCall =0; nCall++;
...
          LOG_ERROR <<
            Form("**** StiHit.%s outside (%d) by %g (%g) ****"
            ,detector->getName().c_str()
 	   ,StiDetector::mgIndex
 	   ,StiDetector::mgValue[0],StiDetector::mgValue[1])
          << endm;
-          assert( detector->insideL(dd,7,1.5));     // 
+	 //	 assert( detector->insideL(dd,7,1.5));     // 
+	 static Int_t ibreak = 0;
+	 ibreak++;
```
plexoos: These lines don't do anything.

StRoot/Sti/StiHit.cxx
```diff
@@ -147,15 +148,20 @@ static int nCall =0; nCall++;
       mz = gz;
       double dd[3]={mx,my,mz};
 
-
-      if (!detector->insideL(dd,7,1.2)) {
+//		Special TPC case. Pileup tracks could be very far
+//              from volume in Z for splitted sectors
+      int mask = 7;
+      if (detector->getGroupId()==kTpcId) mask = 3;
+      if (!detector->insideL(dd,mask,1.2)) {
```
plexoos, line 155: May affect past productions.

StRoot/Sti/StiHit.cxx
genevb: ...except remove the assert?

plexoos: GitHub's UI allows comments on specific lines by selecting them, so the two lines I selected do not remove the assert. My suggestion is to remove the two lines with the `ibreak` variable.

```suggestion
```

Regarding the assert, I think it was removed because it is not compatible with the new logic in the preceding `if` statement.

fisyak: commented out

StRoot/StDbUtilities/StTpcCoordinate.cxx
fisyak: it is matter of taste. Ignored.

plexoos: Not quite. Adding unused headers will increase compilation time."

mgr/ConsDefs.pm
```diff
@@ -801,6 +801,7 @@
 	# restore  
 	($CFLAGS,$CXXFLAGS) = @ARGS;        # <--- same order than above
     }
+ $CXXFLAGS .= " -fdiagnostics-color=always";#  -Wno-pedantic -Wno-shadow -Wno-unused-parameter -Wno-extra -Wno-shadow";
```
klendathu2k, line 804: This should be `-fdiagnostics-color=auto`, so it will only be enabled when gcc recognizes it is outputting to a terminal.

StarDb/tpc/tsspars/tsspar.year_1a.C
```diff
@@ -1,58 +0,0 @@
-#include "tables/St_tss_tsspar_Table.h"
```
klendathu2k, line 1: Obsolete files should be deleted in a separate pull request to minimize the amount of noise reviewers need to sift through.

StarDb/RunLog/onl/starClockOnl.C
```diff
@@ -7,7 +9,7 @@ TDataSet *CreateTable() {
   memset(&row,0,tableSet->GetRowSize());
   row.runNumber	 =    2214033; // run number  ;
   row.time	 =  996787667; // unix time of entry  ;
-  row.frequency	 =    9383160; // frequency in Hz  ;
+  row.frequency	 =    9399962; // local clock
```
klendathu2k, line 12: Confused by this change.  There are `starClockOnl.y2001.C` ... `starClockOnl.y2014.C` as well, which I though take precedence over this file.  The frequency is set to 9399970 there.   

Or is this change in effect for all years that don't have a corresponding y20xx file?  If so, why is it appropriate to change those past runs?

StarDb/RunLog/MagFactor.y2016.C
```diff
@@ -0,0 +1,16 @@
...
+// StarDb/.data/StarDb/RunLog/MagFactor Allocated rows: 1  Used rows: 1  Row size: 4 bytes
+//  Table: MagFactor_st[0]--> MagFactor_st[0]
+// ====================================================================
+// ------  Test whether this table share library was loaded ------
+  if (!gROOT->GetClass("St_MagFactor")) return 0;
+MagFactor_st row;
+St_MagFactor *tableSet = new St_MagFactor("MagFactor",1);
+//
+memset(&row,0,tableSet->GetRowSize());
+    row.ScaleFactor	 =        -1.002; // ;
```
klendathu2k, line 12: Why are we changing the magnetic field in 2016?

StRoot/Sti/StiTrackNodeHelper.cxx
```diff
@@ -465,16 +477,8 @@ int StiTrackNodeHelper::join()
...
-   if (std::fabs(mJoinPars.hz() - mTargetHz) > 1e-10)
-   {
-     LOG_WARN << "Expected |mJoinPars.hz() - mTargetHz| <= 1e-10 "
-              << "instead |" << mJoinPars.hz() << " - " << mTargetHz << "| = "
-              << std::fabs(mJoinPars.hz() - mTargetHz) << ". "
-              << "Will set mJoinPars.hz to " << mTargetHz << endm;
-     mJoinPars.hz() = mTargetHz;
-   }
-
+   mJoinPars.hz()=mTargetHz;//////////////////////???????????????????????????????????????????????????
```
klendathu2k, line 480: `//////////////////////???????????????????????????????????????????????????`

Not very informative comment here.  This change can impact past productions.

StRoot/Sti/StiTrackNodeHelper.cxx
```diff
@@ -4,10 +4,17 @@
 #include "StiTrackNodeHelper.h"
 #include "StiElossCalculator.h"
 #include "StDetectorDbMaker/StiHitErrorCalculator.h"
+#include "StDetectorDbMaker/StiTpcHitErrorMDF4.h"
+#include "StEvent/StTpcHit.h"
```
klendathu2k, line 8: We probably don't need a TPC hit dependency here...

StRoot/Sti/StiTrackNodeHelper.cxx
```diff
@@ -77,7 +88,7 @@ void StiTrackNodeHelper::set(StiKalmanTrackNode *pNode,StiKalmanTrackNode *sNode
 #ifdef __CHECKIT__
     mTargetNode->mFP.check("1StiTrackNodeHelper::set");
 #endif
-    assert(fabs(mTargetHz-mTargetNode->mFP.hz()) < EC*0.1);
+    //yf    assert(fabs(mTargetHz-mTargetNode->mFP.hz()) < EC*0.1);
```
klendathu2k, line 91: Why is this assert no longer necessary?

StRoot/Sti/StiDetector.h
```diff
@@ -115,6 +117,7 @@ class StiDetector : public Named
 
     /// Hit Error Calculator for this detector
     const StiHitErrorCalculator * _hitErrorCalculator;
+    const StiTpcHitErrorMDF4        *_hitErrorCalculatorMDF4;
```
klendathu2k, line 120: This changes the design in Sti.  Each detector has one hit error caclulator.   It should be an instance of StiHitErrorCalculator or a derived class.  [StiTpcHitErrorMDF4](https://github.com/star-bnl/star-sw/blob/b06b780de6384f87c15282d7b1a139c8cbf7a0ff/StRoot/StDetectorDbMaker/StiTpcHitErrorMDF4.h#L9) inherits from something else.  The consequence of that choice is that StiTrackNodeHelper has to have business logic added to decide which error calculator should be used on a given detector.  It also introduces a dependency between StEvent and StiTrackNodeHelper, that should be avoided.

Recommendation is to either refactor StiTpcHitErrorMDF4 so that it inherits from StiHitErrorCalculator, or create 
facade class that inherits from it.

You can modify the StiHitErrorCalculator class's calculateError method to forward the StiHit to your class.  It can then handle all of the logic that was added to StiTrackNodeHelper.

StRoot/Sti/StiTrackNodeHelper.cxx
```diff
@@ -1102,9 +1112,28 @@ int StiTrackNodeHelper::getHitErrors(const StiHit *hit,const StiNodePars *pars,S
 {
   hrr->reset();
   const StiDetector *det = hit->detector();
-  const StiHitErrorCalculator *calc = (det)? det->getHitErrorCalculator():0;
-  if (calc) {//calculate it
-     calc->calculateError(pars,hrr->hYY,hrr->hZZ);
+  const StiTpcHitErrorMDF4 *calcMDF4 = (det)? det->getHitErrorCalculatorMDF4() : 0;
+  const StiHitErrorCalculator  *calc = (det)? det->getHitErrorCalculator()     : 0;
+  if (calc || calcMDF4) {//calculate it
```
klendathu2k, line 1117: All of the logic dealing with the MDF4 model can be moved into the `StiTpcHitErrorMDF4` class, if we modify the `calculateError` function in StiHitErrorCalculator to accept the StiHit.  See comments around StiDetector.cxx line 120.

StRoot/StDbUtilities/StTpcCoordinateTransform.cc
```diff
@@ -424,12 +409,12 @@ Double_t StTpcCoordinateTransform::zFromTB(Double_t tb, Int_t sector, Int_t row,
   if (! St_starTriggerDelayC::instance()->Table()->IsMarked()) {// new scheme: offset  = clocks*timebin + t0
     trigT0 = St_starTriggerDelayC::instance()->clocks()*mTimeBinWidth;
     elecT0 = St_starTriggerDelayC::instance()->tZero();
+    if (StTpcDb::IsCosmics()) {
```
klendathu2k, line 412: Not sure what the relevance for this PR is.

StRoot/StDbUtilities/StTpcCoordinateTransform.cc
```diff
@@ -521,25 +501,113 @@ void  StTpcCoordinateTransform::operator()(const        StTpcLocalSectorCoordina
...
+    StTpcDb::instance()->Pad2Tpc(sector,row).LocalToMasterVect(a.position().xyz(),xGG.xyz()); 
+    const Double_t *trans = StTpcDb::instance()->Pad2Tpc(sector,row).GetTranslation(); // 4
+    TGeoTranslation GG2TPC(trans[0],trans[1],trans[2]);
+    GG2TPC.LocalToMaster(xGG.xyz(), b.position().xyz());
+  } else {// new alignment with drift along z in Super sector coordinate  system
+    Double_t drift = a.position().z() +  StTpcDb::instance()->SubS2Sup12S(sector,row).GetTranslation()[2];
+    TGeoHMatrix &rotA = Sub2Tpc(sector, row, drift);
+    rotA.LocalToMaster(a.position().xyz(), b.position().xyz());
+  }
+  b.setSector(sector); b.setRow(row);
```
klendathu2k, line 514: Am I missing something, or does this introduce a change in the old behavior?   Looking at the code on the left...
```
b.setSector(a.sector()); b.setRow(row);
```

From this version
```
Int_t sector = a.fromSector();
...
b.setSector(sector); b.setRow(row);
```

StRoot/StEvent/StTpcHit.cxx
```diff
@@ -89,14 +94,59 @@ void StTpcHit::setExtends(Float_t cl_x, Float_t cl_t, Short_t mnpad, Short_t mxp
...
+	      << Form(" pmin/max %3i/%3i np %2i tmin/max %3i/%3i nt %2i tm %6.2f pad %6.2f adc %5i",
 		      (Int_t)  v.minPad(), (Int_t)  v.maxPad(),(Int_t) v.padsInHit(), 
 		      (Int_t) v.minTmbk(), (Int_t) v.maxTmbk(),(Int_t) v.timeBucketsInHit(),
 		      v.timeBucket(),v.pad(), v.adc());  
 }
+//________________________________________________________________________________
 void   StTpcHit::Print(Option_t *option) const {cout << *this << endl;}
+//________________________________________________________________________________
+#ifdef __TFG__VERSION__
+ const Char_t *StTpcHit::GetPath() const {
```
klendathu2k, line 112: No.  This does not belong in the event model.  Define a helper function and include it in your code to convert a hit to a geometry path.

StRoot/StEvent/StTpcHit.h
```diff
@@ -153,38 +157,46 @@ class StTpcHit : public StHit {
...
+    StThreeVectorF mPositionU;  //  upper position = y_local + padlength/2.
+    StThreeVectorF mPositionL;  //  lower position = y_local - padlength/2.
+    Float_t        mdX;         //  estimated dX from StdEdxY2Maker
+    Float_t        mTimeBucket; // average timebucket corrected
+    Float_t        mdY;         // Z correction from Hit error parameterization
+    Float_t        mdZ;         // Z correction from Hit error parameterization
+#ifdef __TFG__VERSION__
+    static TString fgFMT;        
+#endif /*  __TFG__VERSION__ */
+    ClassDef(StTpcHit,13)
```
klendathu2k, line 199: Should be 11, not 13.

StRoot/Sti/StiNodePars.h
```diff
@@ -1,7 +1,7 @@
 #ifndef __StiNodePars_h_
 #define __StiNodePars_h_
 #include "Rtypes.h"
-
+#include "TMath.h"
```
klendathu2k, line 4: I don't see any clear advantage to switching to TMath here.  Keep fabs.

StRoot/Sti/StiHit.h
```diff
@@ -110,7 +111,7 @@ class StiHit
 
     ///Return a const reference to a StThreeVectorF that denotes the position
     ///of the hit in global STAR coordinates.
-    const StThreeVectorF globalPosition() const;
+    StThreeVectorF globalPosition() const;
```
klendathu2k, line 114: Why is it necessary to drop const here?  Of course, the comment implies that at one time this was returning a reference.  Leading to another question of why did we de-optimize this by returning a copy in the first place I suppose.

StRoot/StiTpc/StiTpcDetectorBuilder.cxx
```diff
@@ -236,11 +237,17 @@ StiDetector* StiTpcDetectorBuilder::constructTpcPadrowDetector(StiLayer stiLayer
   pDetector->setShape(pShape);
   pDetector->setPlacement(pPlacement);
 
-  if (tpc_padrow_id <= nInnerPadrows)
+  if (tpc_padrow_id <= nInnerPadrows) {
+#ifdef StiTpcHitErrorMDF4_h
```
klendathu2k: This should be controllable via chain option, not at compile time.

StRoot/Sti/StiNodePars.h
klendathu2k: Likewise I am not sure what problems are solved on lines 24 and 25.

StRoot/StTpcDb/StTpcDbMaker.cxx
```diff
@@ -255,7 +260,17 @@ Int_t StTpcDbMaker::InitRun(int runnumber){
     gMessMgr->QAInfo() << "ReSet environment NewTpcAlignment (year >= 2013)  from " << iNewTpcAlignmentOld << " to " << iNewTpcAlignment << endm;
     gEnv->SetValue("NewTpcAlignment", iNewTpcAlignment);
   }
- 
+#if 0 /* move to  StBFChain::SetDbOptions */
```
klendathu2k, line 263: Remove inop code.

StRoot/Sti/StiTrackNodeHelper.cxx
```diff
@@ -65,6 +72,10 @@ void StiTrackNodeHelper::set(StiKalmanTrackNode *pNode,StiKalmanTrackNode *sNode
   mParentNode = pNode;
   mTargetNode = sNode;
   mTargetHz = mTargetNode->getHz();
+assert(mTargetHz);
+ if (! sNode->fitPars().hz()) sNode->fitPars().hz() = mTargetHz;
```
klendathu2k, line 76: Wondering what condition leads to this...  Does this mean that in fixed target we are seeing nodes w/out a magnetic field?

StRoot/Sti/StiTrackNodeHelper.cxx
```diff
@@ -667,7 +671,9 @@ double StiTrackNodeHelper::joinVtx(const double      *Y,const StiHitErrs  &B
 //______________________________________________________________________________
 int StiTrackNodeHelper::save()
 {
+mPredPars.hz()=mTargetHz;///????????????????????????????
```
klendathu2k, line 674: Is this something we have been missing for ~20 years?

klendathu2k (changes requested):
1) Several MagFactor tables added and/or modified.   It looks like most just implement the ideal magnetic field for a given year.  But MagFactor.y2016.C sets a non-ideal value of -1.002.  Why this value when everything else is an ideal +/-1 ?

2) StEvent/StTpcHit.h / cxx  extends the data stored and should result in a ID passed to the ClassDef macro.   Unless there is a good reason to, we should keep these IDs consecutive, so the next value is 11 not 13 as per the PR.

There is a helper function defined in the proposed StTpcHit to return a format string for building a ROOT/TGeo of one of our geometry models of the TPC.  This should go into a helper function somewhere other than StEvent.

3) MDF4 Hit Error Model

This should be enabled / disabled by a BFC option, not a compile time flag.

This should use the standard Sti interface... you should not need to modify StiDetector to apply this.  A wrapper class which inherits from StiHitErrorCalculator can forward the calculateError call to your StiHitErrorMDF4 class.  (calculateError will need to be modified to accept an StiHit).  

With that change, I think that the changes to StiTrackNodeHelper lines 1117 to 1136 can be moved into the wrapper class's calculateError function.



The easiest way to do this is with a wrapper function which inherits from StiHitErrorCalculator (or maybe StiTpcHitErrorCalculator).   The wrapper class will forward the calculateError call to the MDF4 model.

fgeurts:
Hi Yuri, I see several comments were made by STAR software experts. Can you follow up on these as your earliest convenience so that we can get the merger and new library set up by early next week? Thanks!!

StRoot/Sti/StiTrackNodeHelper.cxx
fisyak: Hi Frank and Jason, I am working on the comments.

StRoot/StBFChain/BigFullChain.h
fisyak: Ignored

StRoot/StBFChain/BigFullChain.h
fisyak: Ignored

StRoot/StDbUtilities/StTpcCoordinate.cxx
fisyak: Ignored

mgr/ConsDefs.pm
fisyak: Ignored

StarDb/tpc/tsspars/tsspar.year_1a.C
fisyak: This will be issue for another PR

mgr/ConsDefs.pm
klendathu2k: The concern is that the terminal color escape characters used by colorizing the gcc warnings will propagate to our 
[autobuild](https://www.star.bnl.gov/public/comp/prod/Sanity/AutoBuild.html) status page.  My reading of the documentation is that "auto" will avoid that problem.

StRoot/StEvent/StTpcHit.h
plexoos: I second Jason's comment. Why skip versions?

StRoot/StBFChain/BigFullChain.h
genevb: `CorrZ`, as implemented, removes the `OSectorAlign` chain option that implements a TPC distortion correction I implemented. I cannot remember a presentation justifying specifically the removal of this distortion correction, but I welcome being pointed to such.

StarDb/RunLog/onl/starClockOnl.C
fgeurts: It would be good to understand this. @fisyak, can you comment on this question?

StarDb/RunLog/MagFactor.y2016.C
fgeurts: @fisyak, can you answer this question? The proposed changes should only affect 2019+ productions.

fgeurts:
> Hi Yuri, I see several comments were made by STAR software experts. Can you follow up on these as your earliest convenience so that we can get the merger and new library set up by early next week? Thanks!!

Thanks, @fisyak. Aside from some programming related comments there were a few more specific comments regarding functionality changes that may be beyond the 2019+ alignment. It would be good to understand the motivation.

StRoot/StDbUtilities/StTpcCoordinate.cxx
fisyak: Ignored

StarDb/RunLog/onl/starClockOnl.C
fisyak: We start run with cosmics (local clock) and sometime data base is not populated. This allows to use proper clock for laser runs at the very begin on run.

StRoot/Sti/StiTrackNodeHelper.cxx
fisyak: Ignred

StRoot/Sti/StiHit.h
fisyak: globalPosition object  is created on call and does and return value. This is compilator waring suppression.

StRoot/Sti/StiTrackNodeHelper.cxx
fisyak: Some times node is created with zero filed. This is check to avoid assert.

StRoot/StTpcDb/StTpcDbMaker.cxx
fisyak: removed

mgr/ConsDefs.pm
genevb, line 804: Compilation error on SDCC (I'm trying to build an SL24y test library with this branch):
```
g++ -m32 -fPIC -pipe -Wall -Woverloaded-virtual -std=c++0x -Wno-long-long -pthread -Wno-deprecated-declarations -fdiagnostics-color=always-DEVTGEN_EXTERNAL=1 -DEVTGEN_PYTHIA=1 -DEVTGEN_PHOTOS=1 -DEVTGEN_TAUOLA=1 -DEvtGen_version=\"EvtGen1_06_00\" -DPythia8_version=\"Pythia8_1_86\" -g -Dsl73_gcc485 -D__ROOT__ -DNEW_DAQ_READER -I. -IStRoot -I.sl73_gcc485/include -IStRoot/StarGenerator/EvtGen1_06_00 -IStRoot/StarGenerator/HepMC2_06_09 -IStRoot/StarGenerator/Pythia8_1_86/include -IStRoot/StarGenerator/Photos3_61/include -IStRoot/StarGenerator/Tauola1_1_5/include -I/afs/rhic.bnl.gov/star/packages/SL24y -I/afs/rhic.bnl.gov/star/packages/SL24y/StRoot -I/afs/rhic.bnl.gov/star/packages/SL24y/.sl73_gcc485/include -I/cvmfs/star.sdcc.bnl.gov/star-spack/spack/opt/spack/linux-rhel7-x86/gcc-4.8.5/root-5.34.38-llsepmmfuwlsucogcwbjiodncxanoudt/include -I/afs/rhic.bnl.gov/star/packages/SL24y/StRoot/StarGenerator/EvtGen1_06_00 -I/afs/rhic.bnl.gov/star/packages/SL24y/StRoot/StarGenerator/HepMC2_06_09 -I/afs/rhic.bnl.gov/star/packages/SL24y/StRoot/StarGenerator/Pythia8_1_86/include -I/afs/rhic.bnl.gov/star/packages/SL24y/StRoot/StarGenerator/Photos3_61/include -I/afs/rhic.bnl.gov/star/packages/SL24y/StRoot/StarGenerator/Tauola1_1_5/include -c .sl73_gcc485/obj/StRoot/StarGenerator/EvtGen1_06_00/StarEvtGenDecayer.cxx -o .sl73_gcc485/obj/StRoot/StarGenerator/EvtGen1_06_00/StarEvtGenDecayer.o
g++: error: unrecognized argument in option '-fdiagnostics-color=always-DEVTGEN_EXTERNAL=1'
g++: note: valid arguments to '-fdiagnostics-color=' are: always auto never
cons: *** [.sl73_gcc485/obj/StRoot/StarGenerator/EvtGen1_06_00/StarEvtGenDecayer.o] Error 1
cons: errors constructing .sl73_gcc485/obj/StRoot/StarGenerator/EvtGen1_06_00/StarEvtGenDecayer.o
```

fisyak: The problem is missing space between "-fdiagnostics-color=always" and "-DEVTGEN_EXTERNAL=1' coming from mgr/Conscript-standard:608:        $CXXFLAGS .= "-DEVTGEN_EXTERNAL=1 -DEVTGEN_PYTHIA=1 -DEVTGEN_PHOTOS=1 -DEVTGEN_TAUOLA=1 -DEvtGen_version=\\\"$pkg\\\" ";

fisyak: I have committed the " " before -DEVTGEN_EXTERNAL=1 -...

genevb:
I ran the nightly test suite on SDCC using this `TFG24c.Export.Mini` branch. The first finding is that all the `*_embed` jobs failed with the following assertion:
```
StChain:WARN  - StDbSql::QueryDb(table,time) line=439 tpcTimeBucketCor  has No data for query
root4star: .sl73_gcc485/obj/StRoot/StDetectorDbMaker/StDetectorDbChairs.cxx:530: static St_tpcTimeBucketCorC* St_tpcTimeBucketCorC::instance(): Assertion `table' failed.
```
An example such log file is here:
```
/star/rcf/test/gitdev/daq_sl302.ittf/Wed/year_2012/UU193_embed/st_physics_adc_13115004_raw_2510001.log
```

The 64-bit jobs also all failed, unable to open `libCore.so.5`. That may have been something I missed, so I may try it again for tomorrow if I spot my mistake.

Other jobs generally ran (a few are still running), and people can compare the output between yesterday's dev jobs at:
```
/star/rcf/test/dev/*/Tue/
```
and today's jobs with this PR's code branch:
```
/star/rcf/test/gitdev/*/Wed/
```

-Gene

fisyak:
To use nightly test is very good idea. To make track by track comparison I need event.root files in both dev  and girder.

fisyak:
Event root files are needed with hits, i.e. without "hitfilt".

genevb:
> Event root files are needed with hits, i.e. without "hitfilt".

Many such event.root files are available from the nightly test suite I ran...at least 48 of them. See the outputs of:
```
ls -l /star/rcf/test/gitdev/*/Wed/year*/*/*.event.root | grep "Sep 11"
ls -l /star/rcf/test/dev/*/Tue/year*/*/*.event.root | grep "Sep 1"
```
...and pretty much all the chains have `-hitfilt`, so they are generally without `hitfilt`.

fisyak:
I have posted comparison of dev and SL24y productions to answer the question on the effect of code modification on production of old (before iTPC) data with old chains.
https://drupal.star.bnl.gov/STAR/blog/fisyak/Track-track-and-hit-hit-analyses-dev-and-SL24y-Alignment2024-AuAu2002016-sample

I don't see any show stoppers to start the calibration production with SL24l for all fixed target samples from 2019-2021 with new alignment (CorrZ).

StRoot/Sti/StiTrackNodeHelper.cxx
```diff
@@ -1102,9 +1112,28 @@ int StiTrackNodeHelper::getHitErrors(const StiHit *hit,const StiNodePars *pars,S
...
+  if (calc || calcMDF4) {//calculate it
+    Double_t fudgeFactor = 1;
+    StTpcHit *tpcHit = (StTpcHit*) hit->stHit();
+    if (tpcHit) {
+      if ((tpcHit->detector() == kTpcId || tpcHit->detector() == kiTpcId)) {
+	if (tpcHit->flag() == 2) {
+	  fudgeFactor = 16.;
+	}
+      }
+    }
```
genevb, line 1126: Yuri suggested that these lines of code in `Sti/StiTrackNodeHelper.cxx`, modifying some TPC hit errors, might be responsible for the differences seen in nightly tests with this PR vs. DEV for pre-iTPC datasets where the alignment should be unaffected. Processing a few events with and without this bit of code is quite easy, and comparing the log files obviates that the bulk of the difference in track counts does disappear when removing this section of code.

However....from looking at the log file, I see that the new alignment scheme appears to alter TPC distortion corrections for these old, pre-iTPC datasets without any change to the reconstruction chain. This is in addition to the modification to TPC distortion corrections ("OSectorAlign") that Yuri wants to impose with the "CorrZ" chain option he has proposed for iTPC-era datasets for which I have expressed an interest in seeing a justification (through data, not anecdotally).

    < StMagUtilities::XTWIST        =  -0.38 mrad
    < StMagUtilities::YTWIST        =  0.25 mrad
    ---
    > StMagUtilities::XTWIST        =  0 mrad
    > StMagUtilities::YTWIST        =  0 mrad

    < StMagUtilities::WestClock     =  -0.43 mrad
    ---
    > StMagUtilities::WestClock     =  0 mrad

I do not recall this being discussed before, neither in the context of pre-iTPC datasets, nor with the new alignment. If the TPC alignment calibration was truly performed without any correction for the distortions due to the misalignment of the E and B fields (represented by the XTWIST & YTWIST numbers), I have significant concerns about this TPC alignment. Hopefully I am missing something and this is not the case.

____________

I have placed output of the nightly test without this particular code (but including the rest of this PR) at this location:

This PR minus hit error change :
    /star/rcf/test/gitdev/daq_sl302.stica/Mon/year_2016/AuAu200_production_2016_64bit/
    
This is for comparison with other iterations:

DEV :
    /star/rcf/test/dev/daq_sl302.stica/Sat/year_2016/AuAu200_production_2016_64bit/
This PR :
    /star/rcf/test/gitdev/daq_sl302.stica/Tue/year_2016/AuAu200_production_2016_64bit/
This PR before the Sept. 11th commits :
    /star/rcf/test/gitdev/daq_sl302.stica/Wed/year_2016/AuAu200_production_2016_64bit/

I am making note of the last two in particular because Yuri's study involved the output from the last one (those jobs were started on September 11th before the time of Yuri's last commits), even though data with those commits was available. (Side note, please pay no heed to the "Tue" or "Mon" in the gitdev paths, as I moved files there from other days to avoid them being over-written.) I will note that I didn't see any effective impact of those commits on this particular test's track counts (i.e. the last two outputs may be identical).

fisyak: https://www.star.bnl.gov/~fisyak/star/TbyT/AuAu200_2016_HitError/

genevb: Yuri has corrected me:
The "Twist" and "Clock" distortion corrections have not actually been used for the past decade, beginning with the prior "New Tpc Alignment" that was introduced around that time. StMagUtilities prints these numbers, but there has been no distortion correction applied. The alignment work introduced in this PR has resulted in the printed values being zero, but no actual effect on the applied distortion corrections.

PR #704, opened 2024-08-12
Add EPDX to RICH scalers

genevb:
The EPD coincidence rate began filling element `rs14` of the RICH scaler struct in the DAQ stream and online databases at the end of March 2018. This needs to be propagated into the `trigDetSums` table in the offline database ([now done](https://www.star.bnl.gov/cgi-bin/protected/viewvc.cgi/cvsroot/online/DataBase/dbcron/macros/FillTD.C?r1=1.2&r2=1.3)), and when reading the DAQ structures (this PR). As was previously done for the no-killer ZDCs, this makes use of a no-longer needed scaler in the trigDetSums struct (TOFp in this case), while retaining backward compatibility with older data in which that now-defunct scaler was actually used.

### DIFF
--- a/StRoot/StDAQMaker/StSCReader.h
+++ b/StRoot/StDAQMaker/StSCReader.h
@@ -36,6 +36,7 @@ class StSCReader
     double getZDCWestNoKiller();
     double getZDCEastNoKiller();
     double getZDCXNoKiller();
+    double getEPDX();
     double getMult();
     double getL0();
     double getBBCX();
@@ -57,6 +58,7 @@ class StSCReader
     sc_t *fSC;
     short flipBBCBkg;
     short useNoKillers;
+    short useEPD;
 };
 
 #endif
--- a/StRoot/StDAQMaker/StSCReader.h
+++ b/StRoot/StDAQMaker/StSCReader.h
@@ -57,8 +57,8 @@ class StSCReader
 
     sc_t *fSC;
     short flipBBCBkg;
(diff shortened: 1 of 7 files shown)


genevb:
This PR also includes propagation of the EPDX scaler into the ntuple used for TPC SpaceCharge calibrations.

StRoot/StDAQMaker/StSCReader.h
```diff
@@ -57,6 +58,7 @@ class StSCReader
     sc_t *fSC;
     short flipBBCBkg;
     short useNoKillers;
+    short useEPD;
```
plexoos: There are a few good reasons to use the `bool` type for a variable intended to represent a true/false value in c++.

plexoos (commented):
I have no comments on the changes per se, but it would be helpful to ensure everything works as expected before merging.

StRoot/StDAQMaker/StSCReader.h
genevb: Agreed, and patched.

genevb:
> I have no comments on the changes per se, but it would be helpful to ensure everything works as expected before merging.

Tested on data from before and after the EPD introduction, which was indeed worth doing as I had a typo (`=` instead of `==`). Functions properly now. Thanks!

iraklic (approved):

plexoos (approved):

PR #707, opened 2024-08-18
Update Stgm geometry z-plane locations according to survey data

jdbrice:
Move sTGC planes to match the z-locations measured as part of the precision position survey. Use the z coord for quadrant A on each plane as the z-location for the entire plane

### DIFF
--- a/StRoot/StFttSimMaker/StFttFastSimMaker.cxx
+++ b/StRoot/StFttSimMaker/StFttFastSimMaker.cxx
@@ -65,7 +65,7 @@ Int_t StFttFastSimMaker::Make() {
         
         StFttPoint *point = new StFttPoint();
         point->setPlane(plane_id);
-        point->setQuadrant(0); // TODO this could be improved, but it is not used in the current implementation
+        point->setQuadrant(0);
         StThreeVectorD xyz;
         xyz.set(x, y, z);
         point->setXYZ( xyz );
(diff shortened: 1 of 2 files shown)


klendathu2k (commented):
First, do we need to create an StgmGeo2 geometry and define new geometry tag y2024b?
In other words, do we have any existing data productions that depend on this geometry?  And if so, are those productions valid?

Second, it looks like we are hard coding misalignments into the geometry.  Is there a reason for doing it this way?  I would strongly prefer the approach where the geometry model reflects the ideal (as designed) detector positions, and misalignments are applied through misalignment tables.

jdbrice:
@klendathu2k good questions:
- No: the previous z-positions were never used. They were put in because thats where we intended to have the detector but made a change rather late. No productions depend on this, and all simulations are private (no official embedding etc.)
- I am not 100% sure what you mean. The z-locations are the installed (ideal) positions shifted by just a few mm since we have the survey data. However, we are using the misalignment tables correctly aren't we (see below)


```xml
<Placement block="STGM" in="CAVE" konly="ONLY" x="0" y="5.9" z="338.8385" >
            <Misalign table="Geometry/stgc/stgcOnTpc" row="0"/>
</Placement>
```

klendathu2k:
Thanks for the clarification.

> @klendathu2k good questions:
> 
>     * No: the previous z-positions were never used. They were put in because thats where we intended to have the detector but made a change rather late. No productions depend on this, and all simulations are private (no official embedding etc.)
> 
>     * I am not 100% sure what you mean. The z-locations are the installed (ideal) positions shifted by just a few mm since we have the survey data. However, we are using the misalignment tables correctly aren't we (see below)
> 
> 
> ```
> <Placement block="STGM" in="CAVE" konly="ONLY" x="0" y="5.9" z="338.8385" >
>             <Misalign table="Geometry/stgc/stgcOnTpc" row="0"/>
> </Placement>
> ```

klendathu2k (approved):
LGTM

plexoos (approved):

PR #709, opened 2024-08-21
Eliminte HitFilt from production chains aas obsolete

fisyak:
By Frank request create a separate request for eliminate obsolete option HitFilt from production chains.

### DIFF
--- a/StRoot/StBFChain/BigFullChain.h
+++ b/StRoot/StBFChain/BigFullChain.h
@@ -61,10 +61,10 @@ Bfc_st BFC[] = { // standard chains
   {"Geometry    ","-----------","-----------","------------------------------------------","","","",kFALSE},
   {"------------","-----------","-----------","------------------------------------------","","","",kFALSE},
 #ifdef __AgMLonFly__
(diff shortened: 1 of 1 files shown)


klendathu2k (approved):
I note that the KeepSvtHit option is removed as well, presumably b/c it only makes sense when the hitfilt option is in place.  I don't see any issues here.

genevb (commented):
I maintain that it's inappropriate to change old chains, as our practice has been to create new versions of old chains (e.g. increment a letter at the end of the chain name) when we modify them. The changes in this PR are nevertheless rather inconsequential, and the old chains are on record if they need resurrected, so I will not block this if it allows the Collaboration to move forward in other ways.

I also appreciate that this change was, appropriately, isolated in its own pull request.

iraklic (approved):

iraklic:
I understand Gene's point, but just to add my two cents, I multiple times forgot to explicitly turn off hitfilt  and have run reco without a desired output, this would have saved me a lot of time :)

plexoos:
Can someone explain how this option is obsolete if it is a dependent option in so many chains, e.g. `pp2024a`?

PR #711, opened 2024-09-22
Clock jump calibration

YannickSoehngen:
Methods for retrieving get4 state from db and flagging calibrated hits

PR #712, opened 2024-10-02
new ETOF table for Yannick

dmarkh:
Subj. A new Offline DB table was created for Yannick. Let's see if 4MB records are manageable...
https://drupal.star.bnl.gov/STAR/subsys/etof/etof-database-tables/etof-get4state

### DIFF
--- a/StDb/idl/etofGet4State.idl
+++ b/StDb/idl/etofGet4State.idl
@@ -1,0 +1,16 @@
+/* etofGet4State.idl
+*
+* table: etofGet4State
+*
+* description:Get4States and state changes dealing with "Clock-Jumps"
+*             0 - good,  1 - too early by 6.25ns ,2 too late by 6.25 ns,3 - bad
+*
+* author: Yannick Söhngen ( PI Heidelberg )
+*
+*/
+
+struct etofGet4State {
+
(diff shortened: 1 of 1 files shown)


plexoos (approved):

genevb (approved):
Pretty straightforward.

Considering that I get a single instance of the `Calibrations_tof / tofINLSCorr` table is approximately 1 GB, it seems likely to me that 4 MB entries should not be an issue.

Note: we expect PR #711 is dependent on this PR #712 , so it would be good to get this one in first.

dmarkh:
If my math is correct, tofINLSCorr is ~60MB per record: 30000 * 1024 * 2 =
61,440,000 (and only 2048 bytes per db row).
If I remember correctly, that table from 2009 was replaced with the
optimized version because it was too slow.

Now, eTOF has ~4MB per entry (equals to db row) = 1,000,000 * 4, which is a
single 4MB chunk and is going to be handled very differently at the DB
level... Let's see how it goes.

-Dmitry

genevb:
Sorry, you are correct: 60 MB for the tofINLSCorr table (1 GB was after this converted into a CINT text file, so lots of bytes unassociated with the actual payload). Thanks for the correction and additional comments, Dmitry.

-Gene

PR #714, opened 2024-11-13
Geometry y2021b

klendathu2k:

PR #715, opened 2024-11-15
chains for b versions of 2019-2021 geometries

genevb:
Goes along with PR #714 and commit 76329520deea6c856fd0304858ae44004750156f from several years ago.


klendathu2k (approved):
Looks good to me.

iraklic (approved):
Why there is no B2018b? why straight to c?

genevb:
> Why there is no B2018b? why straight to c?

I considered this. There was a 2018b geometry, but I never produced any data with it. And because I will use "c" the next time, there will never be a need for a chain that uses the 2018b geometry. No need to create a chain that will never be used.

PR #716, opened 2024-11-20
RHICf Filter for Event generator

ggfdsa10:
This is a filter for pi0 and neutron which pass through RHICf geometrical acceptance.
It is necessary to generate the RHICf+STAR simulation.

This RHICf filter was also discussed with Jason before requesting the simulation.

Thanks


ggfdsa10:
Oops... Sorry. BBC simulation and RHICf filter commit are duplicated. I will separate it.
and close this PR.

PR #717, opened 2024-11-20
RHICf Filter for event generator

ggfdsa10:
This is a filter for pi0 and neutron which pass through RHICf geometrical acceptance.
It is necessary to generate the RHICf+STAR simulation.

This RHICf filter was also discussed with Jason before requesting the simulation.

Thanks


plexoos (approved):
Unblocking for Jason to make the final decision.

ggfdsa10:
We decided to use the private RHICf filter and its combined simulation
If we need to public, we will do the pull request again. 
I will close this PR.

Thank you

PR #614, opened 2023-10-25
Tfg.export.102423b

fisyak:
The commit is a result of adjustment TPC gating grid for 3p85GeV_fixedTarget_2021 (May) data https://drupal.star.bnl.gov/STAR/system/files/TPC%20Gaiting%20Grid.pdf
There  also are a few modifications for STAR dE/dx model required by TpcRS.

PR #723, opened 2025-01-15
Zeros out the dead zones in the TPC

klendathu2k:
This PR removes the deadzone in the TPC.  Responsibility for simulating the deadzone is then shifted to the gating grid simulation.

This PR impacts all geometry tags that use the TpceGeo3a, 4, 5(*) and 6 models.

This PR should be tested prior to merging.  Note that this implementation would have the following consequences...
 ... the correct version of the TPC slow simulator must be used from this point forward ... 
 ... any backport of the tpc geometry into an earlier embedding library requires backport of the slow simulator and vice versa ...

Should have a discussion about this before mergeing.

----
(*) No tag uses the TpceGeo5 model.  Listed here for completeness.


genevb (commented):
Do these need to be in new Subversions of the geometry volumes, or is it ok to retain (modify) the existing Subversion?

Other than that, I don't see any further concerns.

klendathu2k:
This is a shift of responsibility of simulating hits in the deadzone from the geant simulation to the slow simulator.  So I think any versioning scheme would need to apply to both.

Absent that... 

We have to be careful when back porting into embedding libraries.  The slow simulator and TPC geometry must be consistent with each other.

genevb (approved):
OK, though it would be good if somewhere a link is made between this PR that hands control over to TpcRS (I assume that's where control is meant to be now), and the complementary change in TpcRS where control of this dead region was taken.

fisyak:
TpcRS took over for Gating Grid dead region accounting via StRoot/StdEdxY2Maker/StTpcdEdxCorrection in PR
Tfg.export.102423b #614
PR# 614 has explanation for these modifications,

fisyak (approved):

genevb:
@plexoos : I understand this is needed for embedding for FXT in SL24y. Can you please tag and I'll re-build?

plexoos:
All set! The new tag is SL24y_2.

PR #571, opened 2023-07-25
New FWD tracking mode based on FST-track finding

jdbrice:
These updates provide a new FWD tracking mode in which the track finding is one with FST information.
Tracks are then fit to valid Seeds and then (optionally) any FTT hits are added to the track if found along the trajectory. 

This mode is needed for any analysis of real data at this point. That would include fast offline for ongoing data taking.

PR #724, opened 2025-01-15
Add support in StVertex for "Fwd" vertex

jdbrice:
Minimal updates to enum StVertexId and StVertex class to allow primary vertices to be assigned as FWD - so that we can use the primary vertex index to seamlessly refer to standard primary vertices and those found from forward tracks.


jdbrice:
I am separating out all changes needed for the tracking updates that are part of other packages (not StFwdTrackMaker)
This is the update that is needed for the vertex handling in pr #571

plexoos (approved):

jdbrice:
@klendathu2k - please take a Quick Look at this, very minor update to StVertex to add an id for Fwd

klendathu2k (approved):

PR #728, opened 2025-01-25
fix: remove ROOT5 paths from ROOT6 environment

### LINKED ISSUE #729: Problem with mixed STAR + spack environments
From Jason:

> The problem with the mixed STAR + spack environment is illustrated with two
> scripts located here:
> /star/simu/simu/jwebb/2025/environment-issue
> 
> demo.csh demonstrates the problem.  It sets up the 64b SL24x environment with the
> root 6.24 environment overlaying it.
> 
> The LD_SEARCH_PATH contains ROOT 5 paths which interfere with loading shared libraries.
> The result is that a simple bfc.C\(0\) run tries to load in libVMC.so from the ROOT5 area.
> With predictable results.
> 
> The workaround.csh shows the hack I am currently using to overcome this.

plexoos:
[skip ci]

Resolves #729

### DIFF
--- a/mgr/config/v0.3.0-rhel7-root6.24.06.config
+++ b/mgr/config/v0.3.0-rhel7-root6.24.06.config
@@ -14,3 +14,6 @@ else
 endif
 
 module load star-env-0.3.0-root-6.24.06
+
+# Remove paths containing the substring “ROOT/5” from the environment variable.
+setenv LD_LIBRARY_PATH `echo $LD_LIBRARY_PATH | awk -v RS=: -v ORS=: '/ROOT\/5/ {next} {print}'`


klendathu2k (approved):
May need to clean up the binary search PATH as well.  But this may solve the issue.

iraklic (approved):
I think instead of the `next` and `print` your can do `awk -v RS=: -v ORS=: '!/ROOT\/5/'` but this works too of course

plexoos:
> May need to clean up the binary search PATH as well. But this may solve the issue.

I don’t mind adjusting the PATH if needed, but your test passes without it.

PR #730, opened 2025-01-28
Modify genDst.C to allow use of FXT DB flavors

genevb:
Very simple change.


plexoos (approved):

genevb:
@fgeurts and @plexoos :

It would be helpful to have this merged ASAP, and then tag SL24c (SL24c_0) which is _identical_ to SL24b except for the addition of this single changed macro.

We can consider freezing an SL25a at some point soon with the GMT and FWD tracking changes, but I want to complete the recently started conversion production effort with what is essentially SL24b plus this macro change.

fgeurts (approved):

plexoos:
I see what you’re asking for… but do you think it would make sense to just patch the SL24b branch with this commit and tag it as SL24b_1?

genevb:
> I see what you’re asking for… but do you think it would make sense to just patch the SL24b branch with this commit and tag it as SL24b_1?

The reason against this is that I used this macro for the P24y.SL24b conversion production that did not get btofFXT correct. I will need to reproduce that data, and for bookkeeping, it's best if we just make that P24y.SL24c.

plexoos:
Okay, fine. In the case of the SL24b branch, I see that it didn’t really branch from the main branch (still at SL24b_0), and for this reason, SL24c would also branch directly from the main branch. I guess my concern is that, in general, we should avoid creating official branches off anything other than the main branch to prevent confusion with tags.

genevb:
I'm fine with it not being an SL24c branch, just an SL24c tag....is that good?

plexoos:
Creating a tag without a branch is possible but will likely be just as confusing as, for example, having an SL24c_0 tag on the SL24b branch :) could be puzzling for someone trying to figure this out in the future.

Anyway, in this particular case, I don’t see any issue with creating both SL24c and SL24c_0, which **has now been done**.

PR #734, opened 2025-02-18
Kf particleno vc2

fisyak:
Th second attempt to create KFParticle without Vc library


genevb (dismissed):
I am only looking at 2 files under StRoot/StBFChain out of the 142 files changed by this PR. I see changes that seem un-related to KFParticle (e.g "LdEdx"), and codes specific to TFG builds of their software. Please exclude these from this PR.

fisyak:
142 modifications required by removing  +10 year old KFParticle version and adding new (Dec 2024) version.
Modification to StBFChain is required by the example to run KFParticle analysis in  StRoot/StKFParticleAnalysis/kfpAnalysis.C

genevb:
> 142 modifications required by removing +10 year old KFParticle version and adding new (Dec 2024) version. Modification to StBFChain is required by the example to run KFParticle analysis in StRoot/StKFParticleAnalysis/kfpAnalysis.C

Thank you for removing the TFG-specific `ifdef` codes and justifying the additional chain options.

We have generally been trying to avoid `ProcessLine()` usage recently, but doing so either requires modifying the MuDst and PicoDst makers to use more of the functionality inherited from StMaker (e.g. attributes), or to introduce a compile-time dependence on these makers as this PR already implements for ROOT6. Unless there's another suggestion for how to do this, or someone knows of an issue with running bfc.C in ROOT5 if we do this, I think you may as well just omit the `ProcessLine()` usage for ROOT5 and go ahead with having the dependencies for both ROOT5 and ROOT6, not just for ROOT6. That is my requested change.

I'm generally OK with the remaining StRoot/StBFChain changes, though the "quiet" functionality could/should very easily have been in its own little PR.

Excluding the 24 deleted files still leaves 38 modified-in-place files, and 82 new-and-potentially-modified files, which is 120 files for review (I only reviewed 2 of those). A lot are in new libraries, StKFParticle(Performance,Test,AnalysisMaker), that would in the past have triggered a software peer review for compliance with STAR coding standards.

fisyak:
ok. I addeddependence of StBFChain oon StMuDstMaker & StPicoDstMaker for both ROOT5 & ROOT6 instead only for ROOT6

iraklic:
@fisyak Hi Yuri. It took me a while but I went through this to some degree. Just one thing I don't know is what are `StRoot/StiMaker/StMuDstVtxT.cxx`, `StRoot/StiMaker/StVertexP.h`, `StRoot/StiMaker/StVertexT.h` some test classes that are not needed anymore?

fisyak:
Hi Irakli, thee codes are based on (+15 years) old KFParticle, which was located in StarRoot package. New KFParticle makes thee codes obsolete. The codes are relative to adaptive vertex finding and fitting  based on KFParticle. These codes are moved in new package (StKFVertexMaker) which is not a part of this pull request.

iraklic (approved):

iraklic:
Thanks for the reply @fisyak. From what I reviewed there is a KFParticle part (new and update) and the part where the rest of the STAR code needs to be modifies (this part is minimal and looks good to me), so I approved the PR. Of course the caveat that @genevb mentioned about "A lot are in new libraries, StKFParticle (Performance, Test, AnalysisMaker), that would in the past have triggered a software peer review for compliance with STAR coding standards" is there, but since this is besides the codebase merging and actual code I will leave it to the STAR members that are still closely involved in this.

nigmatkulov (approved):
Looks good to me

genevb:
The merger of this PR caused all nightly builds of DEV to fail. See for example:
https://www.star.bnl.gov/public/comp/prod/Sanity/AutoBuild.html

DEV updates will be disabled until this is resolved.

fisyak:
This contradict to check provided by github.
The example shows that build has failed to code non related to the pull request. 
Fresh check out from star-sw and build with 
time cons EXTRA_CXXFLAGS=-Werror
does not reproduce the problem. 
But the problem exists.  it comes from '-fdiagnostics-color=always-DEVTGEN_EXTERNAL=1' in StarGenerator due to missing blank before "-DEVTGEN_EXTERNAL=1'", and it did not caught because StarGenerator is not part of pull request check.
I have added the blank  mgr/Conscript-standard in Pr#736 (https://github.com/star-bnl/star-sw/pull/736)

PR #736, opened 2025-02-25
Kf particleno vc2

fisyak:


iraklic:
Hi @fisyak . I assume this is to address the lightly test crushes? Can you point to what is different? Also, if 734 is merged, shouldn't it be a little change on top of that? Or it will be reverted?

fisyak:
Hi Irakli, yes, it is response to nightly test fail.
The problem   comes from '-fdiagnostics-color=always-DEVTGEN_EXTERNAL=1' in StarGenerator due to missing blank before "-DEVTGEN_EXTERNAL=1'", and it did not caught because StarGenerator is not part of pull request check.
I have added the blank in  mgr/Conscript-standard (rhttps://github.com/star-bnl/star-sw/pull/736 (#736))

iraklic:
Thank you Yuri. And thanks for cleaning up this PR. Now I only see the necessary single file change.

iraklic (approved):

nigmatkulov (approved):
Looks good

PR #685, opened 2024-05-16
Updated timebin handling by StFstRawHitMaker

jdbrice:
This is an important update that effects the ongoing Fast offline production of Run24 data.
This fixes the raw hit maker producing incorrect hits when the default number of time bins  is not correct.

PR #737, opened 2025-03-09
Revert incorrect changes to fst consts merged in fst-tracking pr.

jdbrice:
recent merge of fast-tracking mode incorrectly modified the values in StFstConst.h
This reverts those changes, since they are no longer needed with the more general fix to FST (PR #685 )


plexoos (approved):

genevb (approved):
I remember well the headaches this caused in the past. Let's get this in and into the imminent SL25a library.

PR #739, opened 2025-03-14
Initial 2025 chain, similar to 2024

genevb:
No plans for pp running, so this has solely an AA chain.
Will need a production "a" geometry later.


iraklic:
I am probably forgetting how chains are constructed, but in the comment for the P2025 you  say "-CorrY" but it is based on the B2025 which has "CorrY" and I don't see it removed. What am I missing? (Also iTpcIT is duplicated)

genevb:
> I am probably forgetting how chains are constructed, but in the comment for the P2025 you say "-CorrY" but it is based on the B2025 which has "CorrY" and I don't see it removed. 

Sorry for the confusion: that isn't "-CorrY with "-" as a prefix to "CorrY. It's " - " as a spacer, indicating that it is a CorrY-based chain. I'll clarify. And goot catch on iTpcIT.

Thanks,
-Gene

klendathu2k (approved):
Looks good to me.

iraklic (approved):

Issue #742, opened 2025-03-29
64-bit crash for nightly tests for 2008 and earlier


genevb:
I tried DEV, SL23f, and SL21d 64bit, all of which crash seg fault after the following line is printed in the logs:

    StChain:INFO  - St_geant_Maker::InitRun -- Do mag.field initialization

In 32-bit, which doesn't crash, the following line is printed next:

     looping on agvolume

Unfortunately, it does not appear that we have functioning `gdb` for 64-bit, so I cannot easily find the line where the code is crashing (perhaps `gdb` should be a separate issue?):

    gdb: symbol lookup error: gdb: undefined symbol: PyUnicodeUCS4_FromEncodedObject

How to reproduce:

    setup 64b
    stardev
    root4star -b -q -l 'bfc.C(1,"pp2003,VFPPV,v0,xi,l3onl,beamLine,logger,CMuDst,-svtIT,NossdIT,Corr2","/star/rcf/test/daq/2003/095/st_physics_4095050_raw_0010002.daq")'

(it doesn't matter whether you run optimized or not, container or not)

-Gene

genevb:
Here is a list of 64-bit nightly test logs that demonstrate the failure in 64-bit
`-rw-r--r-- 1 starreco rhstar    155558 Mar 29 12:15 daq_sl302.ittf_opt/Sat/year_2003/ppMinBias_64bit/st_physics_4095050_raw_0010002.log
-rw-r--r-- 1 starreco rhstar    169796 Mar 29 12:16 daq_sl302.ittf_opt/Sat/year_2004/AuAuMinBias_64bit/st_physics_5028066_raw_1010003.log
-rw-r--r-- 1 starreco rhstar    169945 Mar 29 12:17 daq_sl302.ittf_opt/Sat/year_2004/AuAu_prodHigh_64bit/st_physics_5044102_raw_1010003.log
-rw-r--r-- 1 starreco rhstar    169928 Mar 29 12:17 daq_sl302.ittf_opt/Sat/year_2004/AuAu_prodLow_64bit/st_physics_5044116_raw_3010002.log
-rw-r--r-- 1 starreco rhstar    167412 Mar 29 12:16 daq_sl302.ittf_opt/Sat/year_2004/prodPP_64bit/st_physics_5134013_raw_2010010.log
[...]

PR #745, opened 2025-04-03
GtHash should use unsigned int rather than long, as its representation does not change between 32bit and 64bit.

klendathu2k:


klendathu2k:
Should resolve issue #742

StRoot/St_geant_Maker/GtHash.cxx
```diff
@@ -19,26 +19,26 @@ GtCradle(int n);
...
 }
 ULong_t GtCradle::Hash() const
 { 
-  ULong_t *me = (ULong_t *)fArray;
-  ULong_t ret = fNWords;
+  UInt_t *me = (UInt_t *)fArray;
+  UInt_t ret = fNWords;
 //VP  for (int i=0; i<fNWords; i++) ret ^= me[i];
-  ret = TMath::Hash(me,fNWords*sizeof(ULong_t));
+  ret = TMath::Hash(me,fNWords*sizeof(UInt_t));
```
genevb: The documentation for `TMath::Hash()` indicates that this has been moved to `TString::Hash()` and the function in TMath is only retained for backward compatibility. I checked ROOT 6.32, and `TMath::Hash()` is [still there](https://root.cern.ch/doc/v632/namespaceTMath.html#a567ae063dbdd2095689366bfb06f533b), with the same message. Nevertheless, maybe we want to switch to TString just in case (for future-proofing, as well as any updates/improvements that are made to the Hash() function itself by ROOT contributors)?
```suggestion
  ret = TString::Hash(me,fNWords*sizeof(UInt_t));
```

genevb (approved):
For some reason, github only lets me make change suggestions for lines of code near ones that were modified in this PR, so I cannot easily suggest the change to line 1 to replace TMath.h with TString.h.
```suggestion
#include "TString.h"
```

The other comment I wanted to make is that this file has been unchanged in 24 years, when `TMath::Hash()` was added. This at least makes it straightforward to include this PR's update for 64-bit compatibility in _any_ STAR libraries from the past umpteen years for which we want to run simulations that use the GtHash class. As you said elsewhere, @klendathu2k , this may be relevant for other volumes than the TOFr ones...

So the question is, other than for simulations of 2003-2008 data (which led to us finding the issue #742 in the first place), should we plan to update this file for all existing libraries, or update one-by-one with some understanding of which simulations need this, and consequently which libraries are used for those simulations.

plexoos (approved):

klendathu2k:
The issue is not confined to simulations.  The TVolume geometry is
used in TOF and BTOF match making, for instance.  (Though I believe
we have refactored to TGeo geometry in later libraries.)

I am unaware of other usage, which is just another way of saying that
I am not aware of the extent of its usage in our production chains at
all relevant points in time.  Thus, would argue to back port this into
all 64 bit libraries that we plan to use.

I am wondering if git/github can easily move this commit backwards 24 
years...?

genevb:
@klendathu2k : did you test that this resolves issue #742 ? I just tried compiling this (modified StRoot/St_geant_Maker) locally in 64-bit, unoptimized DEV, and I get the same crash as before:
```
root4star -b -q -l 'bfc.C(1,"pp2003,VFPPV,v0,xi,l3onl,beamLine,logger,CMuDst,-svtIT,NossdIT,Corr2","/star/rcf/test/daq/2003/095/st_physics_4095050_raw_001000
2.daq")'

...

StChain:INFO  - St_geant_Maker::InitRun -- Do mag.field initialization
 looping on agvolume 

Program received signal SIGSEGV, Segmentation fault.
StTofrGeometry::InitFromStar (this=this@entry=0xcfc36a0, starHall=starHall@entry=0xcff9010)
    at .sl73_x8664_gcc485/OBJ/StRoot/StTofUtil/StTofrGeometry.cxx:1290
warning: Source file is more recent than executable.
1290	    if ( trayVolume->GetListSize() ) {
(gdb) 
```

klendathu2k:
Short answer is yes.

Long answer is that I was able to reproduce the crash w/out using
the full chain (just instantiated the geometry via St_geant_Maker...)

Longer answer...  I just tried running the full chain and it worked.

Your build might have failed?

> "warning: Source file is more recent than executable."

genevb:
@plexoos or @veprbl , if one of you is around to quickly try this fix for the shown test, that would be appreciated. Jason and I have been unable to see why we get different outcomes, so it would be good to see if someone else's result can demonstrate which one of us is reproducible. Thanks in advance for any help!

plexoos:
Here’s a quick check. I ran the test command from #742 in two 64-bit containers created from SL25_0 and this PR branch (using the tarballed [artifact from the Actions summary page](https://github.com/star-bnl/star-sw/actions/runs/14256308357)). The command segfaults in the first case but runs to completion with the fixed code.

```
$ singularity run -B /star/rcf docker://ghcr.io/star-bnl/star-sw:SL25a_0-root5-gcc485 bash -l
INFO:    Using cached SIF image
Singularity> root4star -b -q -l 'bfc.C(1,"pp2003,VFPPV,v0,xi,l3onl,beamLine,logger,CMuDst,-svtIT,NossdIT,Corr2","/star/rcf/test/daq/2003/095/st_physics_4095050_raw_0010002.daq")'
```

```
$ singularity run -B /star/rcf docker://ghcr.io/star-bnl/star-sw:PR-745-root5-gcc485 bash -l
INFO:    Using cached SIF image
Singularity> root4star -b -q -l 'bfc.C(1,"pp2003,VFPPV,v0,xi,l3onl,beamLine,logger,CMuDst,-svtIT,NossdIT,Corr2","/star/rcf/test/daq/2003/095/st_physics_4095050_raw_0010002.daq")'
```

PR #747, opened 2025-05-08
Fix crash when running StMuDst2StEventMaker with partial MuDst

jdbrice:
When running afterburner MuDst->PicoDst conversion for FWD we run into this error due to Fms and RHICf Utils crashing when mInfo exists but has no entries. This fixes that issue


akioogawa (approved):

jdbrice:
@klendathu2k @plexoos 
This should be a quick PR to check over - fixes a long time bug that only shows up in MuDst->PicoDst conversion workflows

StRoot/StMuDSTMaker/COMMON/StMuFmsCollection.cxx
```diff
@@ -141,14 +141,18 @@ StMuFmsPoint* StMuFmsCollection::getPoint(int index) {
 }
 
 StMuFmsInfo* StMuFmsCollection::getInfo() {
-  if (!mInfo) return NULL;
-  return static_cast<StMuFmsInfo*>(mInfo->At(0));
+  if (!mInfo) return NULL;  
+  if (mInfo->GetEntriesFast()>0) return static_cast<StMuFmsInfo*>(mInfo->At(0));
+  return NULL;
```
plexoos:
```suggestion
  return nullptr;
```

plexoos (approved):

StRoot/StMuDSTMaker/COMMON/StMuFmsCollection.cxx
```diff
@@ -141,14 +141,18 @@ StMuFmsPoint* StMuFmsCollection::getPoint(int index) {
 }
 
 StMuFmsInfo* StMuFmsCollection::getInfo() {
-  if (!mInfo) return NULL;
-  return static_cast<StMuFmsInfo*>(mInfo->At(0));
+  if (!mInfo) return NULL;
```
genevb: @akioogawa , why use `NULL` here and `nullptr` below?

akioogawa: I thought @plexoos suggested change for both so took it but indeed one remained NULL.

PR #754, opened 2025-06-25
Remove redundant StiCA option from BigFullChain.h

### LINKED ISSUE #629: StiCA option defined twice in StRoot/StBFChain/BigFullChain.h
https://github.com/star-bnl/star-sw/blob/05ba0397d303ddec75efe26a101859d558b29475/StRoot/StBFChain/BigFullChain.h#L1717

https://github.com/star-bnl/star-sw/blob/05ba0397d303ddec75efe26a101859d558b29475/StRoot/StBFChain/BigFullChain.h#L1719-L1720

plexoos:
Resolves #629

### DIFF
--- a/StRoot/StBFChain/BigFullChain.h
+++ b/StRoot/StBFChain/BigFullChain.h
@@ -1784,8 +1784,6 @@ Bfc_st BFC[] = { // standard chains
    ,                      "StEventUtilities,StiUtilities,StiMaker","Sti tracker",kFALSE},
   {"StiCA"      ,"Sti","","Sti","StiMaker","TPCCATracker"                          ,"StiCA tracker",kFALSE},
   {"StiCAPerf","","","","",                             "TPCCATrackerPerformance", "CA Performance",kFALSE},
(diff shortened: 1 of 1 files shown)


iraklic:
@plexoos I see that if you remove this definition of `StiCA` then `libEG` will not be loaded with the other.
Is that intended result? Is `libEG` not needed for `StiCA` or am I missing it being included somewhere else?

plexoos:
I’m not sure where this libEG is coming from or what it provides. It doesn’t look like a STAR library — but please correct me if I’m wrong. I just ran a test job and didn’t see anything relevant in the output:

```
root4star -l -b 'bfc.C(1, "pp2022,StiCA,BEmcChkStat,ftt,-picowrite,-hitfilt,-evout", "/star/rcf/test/daq/2022/010/23010027/st_physics_23010027_raw_1000015.daq")'
```

iraklic:
I don't know if it is needed or not. Perhaps @fisyak can chime in.
It would just be good idea to make sure we are not going to discover where it was needed later :)

klendathu2k:
libEG is ROOT's generic event generator interface.  Unused, as far as I 
know.

genevb (approved):
I believe we should move ahead with this particular change, and see if anything breaks in the DEV nightly tests. We can back-step if we find that we need to.

PR #760, opened 2025-07-30
StBTofCalibMaker bug fix: missing setVertexIndex for MuDst when selecting FXT vertices

starsdong:
In processMuDst() function, MuDst::setVertexIndex is missing when picking a different primary vertex in FXT events. This leads to dcaGlobal values (gDcaMag) to be all default (-999..). The cut on the variable gDcaMag seems to be only used in mPPPAMode. Not sure how much it may affect the afterburner production in this mode.

The issue seems to be NOT relevant for processStEvent() (normal daq production) as the calculations are using different accessing methods.

### DIFF
--- a/StRoot/StBTofCalibMaker/StBTofCalibMaker.cxx
+++ b/StRoot/StBTofCalibMaker/StBTofCalibMaker.cxx
@@ -1401,7 +1401,11 @@ void StBTofCalibMaker::processMuDst()
           int nPrimaryVertices = mMuDst->numberOfPrimaryVertices();
           for(int iVtx=0; iVtx<nPrimaryVertices; iVtx++){
             pVtx = mMuDst->primaryVertex(iVtx);
-            if(pVtx->position().z() > 198. && pVtx->position().z() < 202.) break;
+            if(pVtx->position().z() > 198. && pVtx->position().z() < 202.){
+              mMuDst->setVertexIndex(iVtx);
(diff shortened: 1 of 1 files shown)


genevb:
Good find, @starsdong ...

However, I thought @YannickSoehngen was using this same code for other FXT datasets (processed from MuDst), if I understand correctly, so I would have expected him to run into this same issue earlier. Maybe I do not understand correctly?

starsdong:
This issue appears (from what I found so far) in the dcaGlobal calculation. Now this calculation will be using the default mCurrVertexIndex (-2). This still won't crash the code most of the time since it will check the vertex at this index. If it is null, the code will skip the calculation and returns default values (-999....). But it seems in Yannick's test case, the calculation when checking Vertex at (-2) index returns some memory address (not sure why), then when it performs the dca calculation, it fails and crashes the code. 

More interestingly, when I simply copy the StBTofCalibMaker from lib and compile, the program won't crash although all dcaGlobal values are not correct. 

I still don't fully understand why it crashes specifically at that particular event, but not on most others. Anyway, this PR should fix dcaGlobal calculation issue which is related to the crash of the code. I tested with the specific file Yannick provided, it runs OK in both 32-bit and 64-bit cases. I would suggest Yannick to go ahead and test in massive jobs.

starsdong:
Also just to add, the cut on dca values from what I see in the BTofCalibMaker is only used in the mPPPAMode. Would be interesting whether anyone else saw the issue with dca calculation before in this mode.

fgeurts (approved):

PR #761, opened 2025-07-31
Clang format checking

DanielWielanek:
This is another attempt to set up automatic format checking of modified files.
- files are compared with the main branch, only changed files are checked (repo contains too many files to check all of them)
- files are not modified, to avoid a situation wherethe  local repo is different the remote one

### DIFF
--- a/.clang_format
+++ b/.clang_format
@@ -1,0 +1,11 @@
+---
+BasedOnStyle: Google
+ColumnLimit: '120'
+IndentPPDirectives: None
+IndentWidth: '3'
+MaxEmptyLinesToKeep: '1'
+SpacesInParentheses: 'false'
+TabWidth: '3'
+UseTab: Never
+
+...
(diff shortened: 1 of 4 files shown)


veprbl:
Looks like a cool proposal. Perhaps, this should only apply to select whitelisted files and the new ones, because I don't think reformatting every touched file is what we want.

DanielWielanek:
Hi, but then we should discuss what to put into this whitelist .

JohnTheBlindMilkman:
I think it should be compliant with the STAR C++ Coding Guidelines (https://www.star.bnl.gov/public/comp/sofi/guidelines/codingguide.xml) and the naming standards (https://drupal.star.bnl.gov/STAR/comp/sofi/soft-n-libs/standards).

On that note, looking at e.g. the StHbtMaker, 90% of the code is indented with two spaces and is 80 columns wide. 
If that is the case for other classes and makers, then if we set the clang-format to those guidelines, it shouldn't affect much of the codebase (correct me if I'm wrong).

PR #767, opened 2025-09-08
Embedding pp510 setup 

akanugan:
- This PR is from Xianglei to the tpc calibration for the pp510 embedding setup.
- Has to be back ported to SL20c


iraklic:
Is there a QA presentation for this calibration set? I think we have practiced attaching slides to the corresponding PRs.

akanugan:
> Is there a QA presentation for this calibration set? I think we have practiced attaching slides to the corresponding PRs.

@zhux97 said that he will find the data/MC comparison for pp510 dataset done with this file, which would take some time (~ 2 weeks)

iraklic (approved):
Since this is not really a change but a new calib file with a unique timestamp I approve this not to stop anybody from using this calib, but I hope the promised QA presentation will be atteched.

PR #773, opened 2025-11-14
Make use of already designated topology bits

genevb:
Several already-designated bits in StTrackTopologyMap have remained unused (zeros). However, there are StTrack flags set in StiStEventFiller for exactly these that can be used to set the bits in the topology map. This does require swapping the order of setting the flags to precede filling the topology maps in StiStEventFiller. This PR does nothing to alter the intent of bits in the topology map, i.e. no change to anything in StEvent itself.

The advantage of filling the track topology map over the track flags is that it is propagated all the way to the PicoDst, while the track flags are not, thereby giving PicoDst users a handle on this information about the tracks without changing anything about the PicoDst structure.

I plan to follow this PR with another that makes use of currently undesignated bits in the track topology map to include additional track flags that may be useful to PicoDst users as well.

### DIFF
--- a/StRoot/StEventUtilities/StuFixTopoMap.cxx
+++ b/StRoot/StEventUtilities/StuFixTopoMap.cxx
@@ -35,7 +35,7 @@
  *    map[1]   Bit number  Quantity                                  
  *    ------   ----------  --------                                  
  *                0-20     TPC, remaining 21 padrows                 
- *                21       Track extrapolates to MWC (no=0, yes=1)   
+ *                21       Track extrapolates to MWC (no=0, yes=1) (or has prompt hits)
  *                22       Track extrapolates to CTB (no=0, yes=1)   
  *                23       Track extrapolates to TOF (no=0, yes=1)   
  *                24       Track extrapolates to RCH (no=0, yes=1)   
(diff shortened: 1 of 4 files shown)


StRoot/StEventUtilities/StuFixTopoMap.cxx
```diff
@@ -35,7 +35,7 @@
  *    map[1]   Bit number  Quantity                                  
  *    ------   ----------  --------                                  
  *                0-20     TPC, remaining 21 padrows                 
- *                21       Track extrapolates to MWC (no=0, yes=1)   
+ *                21       Track extrapolates to MWC (no=0, yes=1) (or has prompt hits)
  *                22       Track extrapolates to CTB (no=0, yes=1)   
  *                23       Track extrapolates to TOF (no=0, yes=1)
```
iraklic, line 40: Should this say TOF + BTOF?

iraklic: otherwise I think it this description matches the added code

iraklic (approved):

StRoot/StEventUtilities/StuFixTopoMap.cxx
genevb: I'm wondering if at some point it can also include whether there's an ETOF hit, since a track cannot have hits in both ETOF and BTOF at the same time (same is true for EEMC and BEMC, but oh well). So I'd rather just leave it generically as TOF.

genevb:
Well, I said no change to StEvent itself in my original post, but I realized it might be helpful to users to have accessor functions, so I've added those. So I can now say "no change to StEvent structures", but there are additional functions.

StRoot/StEventUtilities/StuFixTopoMap.cxx
iraklic: Sounds like a good reason to me :)

klendathu2k (approved):
Looks good to me.

PR #774, opened 2025-11-17
Additional flagging of possible pile-up

genevb:
Following quickly on the heels of PR #773 , I have more modifications to StTrackTopologyMap to propose...

A few bits were labeled as "reserved for future use"....well, there isn't much future left to add new detectors to STAR here in the last couple months of running. So I propose to target these for some of the more useful binary track flags instead of leaving them as useless zeros for eternity:

1. post-crossing tracks (definitely contain pile-up hits on the tracks)
2. tracks crossing the TPC central membrane (unlikely to be pile-up)

Both of these are rather TPC-centric, but I think this is worthwhile utilization of the topology map without requiring any structural changes to the PicoDsts in particular.

### DIFF
--- a/StRoot/StEvent/StTrackTopologyMap.cxx
+++ b/StRoot/StEvent/StTrackTopologyMap.cxx
@@ -88,8 +88,8 @@
  *  56   24   RICH
  *  57   25   Barrel EMC/SMD
  *  58   26   Endcap EMC/SMD
- *  59   27
- *  60   28
+ *  59   27   Post-crossing track
+ *  60   28   Central-membrane crossing track
  *  61   29   HFT Format (case 3) - TPC tracks
  *  62   30   turn around flag  (flags that track spirals back)
  *  63   31   FTPC Format (flags TOC or FTPC)
@@ -375,6 +375,26 @@ StTrackTopologyMap::hasHitInEemc() const
         return false;
     else {
         return bit(58);
+    }
+}
+
+bool
+StTrackTopologyMap::postXTrack() const
+{
+    if(ftpcFormat())
+        return false;
+    else {
+        return bit(59);
+    }
+}
+
+bool
+StTrackTopologyMap::membraneCrossingTrack() const
+{
+    if(ftpcFormat())
+        return false;
+    else {
+        return bit(60);
     }
 }
 
--- a/StRoot/StEvent/StTrackTopologyMap.h
+++ b/StRoot/StEvent/StTrackTopologyMap.h
@@ -94,6 +94,8 @@ public:
     bool           hasHitInRich() const;
     bool           hasHitInBemc() const;
     bool           hasHitInEemc() const;
+    bool           postXTrack() const;
+    bool           membraneCrossingTrack() const;
     
     bool           trackTpcOnly() const; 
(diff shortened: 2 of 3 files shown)


klendathu2k:
We should make maximal use of the space available.   Agree that additional TPC pileup flags would be useful here.

Isn't there a second 64bit word (the "addendum")  partially filled with  the iTPC inner padrows?   There would be a few more bits which could be leveraged.

genevb:
There are 23 additional unused bits in the 64-bit iTPC word(s)

Other binary (bool) flags already available in StEvent/StTrack:

    bool       isDecayTrack()      const {return testBit(kDecayTrack);}
    bool       isShortTrack2EMC()        const {return testBit(kShortTrack2EMC);}
    bool       isRejected()              const {return testBit(kRejectedTrack);}
    bool       isWestTpcOnly()           const {return testBit(kWestTpcOnlyTrack);}
    bool       isEastTpcOnly()           const {return testBit(kEastTpcOnlyTrack);}

I believe the last four of those are filled in `StiMaker/StiStEventFiller::fillFlags()`, but not the decay tracks. Further, I think rejected tracks are not propagated to the MuDst nor PicoDst (and a quick test seems to confirm this).

There is an additional boolean flag in StTrack for shortTrack2ETOF, and there's set function for it that is appropriately called, but there is (oddly) no boolean accessor function (could certainly be added).

So by my reckoning, that's 4 more possibly useful bits: two short track flags, two TPC-half flags. Any other suggestions?

I have not (yet) tested whether there are any issues with using the 64-bit iTPC word(s), e.g. for pre-iTPC era data.

klendathu2k:
When you posted the previous PR I almost commented on the BEMC and EEMC.  They are reserved only one bit of information each, and any hit in the towers, pre/postshower or smd layers is sufficient.  (And I'm not sure that  there is any matching done to the smd layers to set the bits in the first place). 

It may be worth asking the PWGs if there are any good use cases for the remaining bits.

genevb:
One could have implement the EMC bit with degeneracy between EEMC and BEMC: a track cannot be in both (similarly for TOF between ETOF and BTOF), so other information about the track would make clear which EMC (or TOF).

But indeed there are the EMC subcomponents you mention, @klendathu2k . The matching to these detectors is performed under StRoot/StiMaker in `StiStEventFiller::fillFlags()` calling `StTrack2FastDetectorMatcher::matchTrack2FastDetectors()`. The latter class uses a list of EEMC hits built in `StEemcHitList::build()` (located in the file StPhiEtaHitList.cxx). The same is true for the other currently-implemented "fast" detectors. And a similar set of codes could be copied to build the hit lists for pre-shower and SMD, though admittedly that's more work than I was aiming for with this PR.

Sidenote: the code for matching to BTOF in this location is one of three such codes, all slightly different (the other locations being  in the vertex-finding, and in the BTOF subsystem code).

klendathu2k:
Think the scope of this PR is appropriate.   Just wanted to start a discussion for additional information to store on the track.  MTD hits come to mind as an appropriate bit to set.

klendathu2k (approved):
LGTM

plexoos (approved):

PR #775, opened 2025-11-18
Update RootCint.pl

jlauret:
ROOT6 dictionary seem to require full path, added logic for this but there is a caveat in relation to path expansion (see comment inside).

### DIFF
--- a/mgr/RootCint.pl
+++ b/mgr/RootCint.pl
@@ -223,6 +223,7 @@ for my $h  (split /\s/,$sources) {
 	    if ( index($IncDirName,$h_dir) == -1 && $h_dir ne $DirName){
 		# print "=> Adding $h_dir to [$IncDirname]\n";
 		$IncDirName .= "-I$h_dir ";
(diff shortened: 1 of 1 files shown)


klendathu2k (approved):
Looks good to me.

plexoos (approved):
I understand this patch is meant to handle cases where $STAR does not point to the directory being compiled. In principle, I would expect the relative paths passed to `rootcint` to be resolved with respect to the current working directory, but if explicitly expanding them helps in practice, that’s fine with me.
I don’t have a test to verify this behavior, but I’m okay with the change.

klendathu2k:
Gene deployed this change in SL24x.  The particleGun simulation runs from an empty test directory without any issues now.

klendathu2k:
> Gene deployed this change in SL24x. The particleGun simulation runs from an empty test directory without any issues now.

... jumped the gun on the test.  Code may not have been recompiled as yet.  Will repeat test once the word is given.

PR #780, opened 2026-01-28
Initialize libs field to empty string.

klendathu2k:
Fixes minor issue.  Field was missing, resulting in the "comments" line being interpreted as the list of libraries to load and non-fatal warnings issued in the log file.


genevb (approved):
This wasn't problematic before???

plexoos (approved):
Looks like a valid fix to me

iraklic (approved):

klendathu2k:
> This wasn't problematic before???

This didn't cause any problems.  No libraries are actually needed for the etofa option, which simply triggers the calib, util and database options.    The only indication of misbehavior is that warnings are issued when the comments are interpreted as libraries which could not be loaded.

klendathu2k:
squashing and merging

PR #785, opened 2026-02-11
Simply add "b" geometries for 2023/4

genevb:
These were missing. Not critical for reconstruction (only the FXT is different), but appropriate for simulations, and helpful to have this in place for reconstruction so that embedding chains pick up the "b" geometries without confusion.


plexoos (approved):

klendathu2k (approved):
Looks good to me.

PR #790, opened 2026-03-05
Coerce to string

klendathu2k:
Rumors are that a new version of the support cone geometry, implemented in the deprecated agstar style, that will be incoming.  The ageParser should streamline this.  Requires (at least) this small fix.


plexoos (approved):

plexoos:
Looks like GitHub Actions is having issues right now. Since this is a low-risk change, we can skip CI for this one.

klendathu2k:
Quite low risk... we run this maybe once every decade and a half ;)

PR #793, opened 2026-03-27
Retire OnlTools

genevb:
This PR is a removal of the entire OnlTools directory (`git rm -r OnlTools`). I'm unaware of a need for this with RHIC operations completed, but I could be wrong. Definitely needs some feedback from @jml985 whether there is any value in maintaining it and ensuring it compiles going forward in time. I don't think there were any updates in the past couple years, so the code is already archived under tags we made in 2025. If I understand correctly, its history isn't lost from git if we remove it (@plexoos might confirm this?).


plexoos:
Yes, removing this directory won’t permanently delete the files. They’ll remain available in the Git history and can be restored later if needed.

plexoos (approved):

jml985:
I don't know of any reason it needs to be available - so long as the code remains possible to retrieve.   It can be used to regenerate JEVP plots from DAQ files but the existing plots exist and the offline QA presumably can handle further analysis.

genevb:
Thanks for the comments, @jml985 . I believe we are on the same page. It would help get this through if you "approve" it.

PR #796, opened 2026-04-07
BFC Chain options for FTT sim and FCS Track Matching

jdbrice:
Add chain options for FTT sim "fttSim"
and FCs track matching "FcsTrackMatch"


StRoot/StBFChain/BigFullChain.h
```diff
@@ -1741,6 +1723,9 @@ Bfc_st BFC[] = { // standard chains
   {"FwdTrack","","","fcsDb","StFwdTrackMaker",
    "XMLIO,genfit2,KiTrack,StarGeneratorUtil,libMathMore,StEventUtilities,StEpdUtil,StFwdTrackMaker",
                                                                              "Forward Track Maker", kFALSE},
+  {"FcsTrackMatch","","","FwdTrack","StFcsTrackMatchMaker",
+  "StFcsTrackMatchMaker",
```
genevb, line 1727: Why spread this out over 3 lines?

genevb (changes requested):
Seems the version of BigFullChain.h used here was a couple commits behind (see PRs #780  and #785  ).

I don't know about Co-Pilot's suggestion to add the StEvent dependence, as the actual StFttFastSimulatorMaker code isn't included here, but it's worth keeping in mind.

PR #801, opened 2026-05-11
Adding the generated event number for StRHICfSimGenerator

ggfdsa10:
This was added to the generated event number for the RHICf event generator
I would appreciate it if you could confirm it.

Seunghwan


plexoos (approved):

plexoos:
@klendathu2k Can we merge this?

klendathu2k (approved):

PR #802, opened 2026-05-11
RHICf CRMC event generator

ggfdsa10:
This package is the external event generator for RHICf MC based on CRMC v.2.2.1
(https://gitlab.iap.kit.edu/AirShowerPhysics/crmc)

I appreciate it if you could approve this PR.

Seunghwan


ggfdsa10:
This package is not compatible with StRoot. 
I canceled all checks.

Seunghwan

plexoos:
Thanks for putting this together. I am not sure this belongs directly in `star-sw` as submitted.

CRMC already has its own upstream repository: https://gitlab.iap.kit.edu/AirShowerPhysics/crmc. This PR appears to import a large external event-generator package, including its own build system, license, external dependencies, model sources, and data files.

What is the benefit of vendoring all of that code into `star-sw` instead of keeping CRMC as an external dependency pinned to a specific release or commit? From a review and maintenance standpoint, it would be much clearer if this PR contained only the STAR/RHICf integration layer needed to use CRMC.

ggfdsa10:
Hi Dmitri,

Thank you for your suggestions.
As I sent you an email about that, we want to produce the official STAR MC with these event generators.
I will close this PR and will make a new PR to reduce the code, which includes the essentially modified parts for RHICf MC, and link it to the original repository for some maintenance as your suggestions.

Best regards,
Seunghwan

PR #803, opened 2026-05-19
build: allow cons packages to opt out with marker files

plexoos:
Add support for excluding directories from the cons build by placing a `.skip_cons_build` marker
file in the directory.

This gives packages a directory-local way to opt out of the default cons traversal instead of
requiring every exclusion to live in `mgr/Construct` or `SKIP_DIRS`. For example, a package
currently disabled directly in `Construct` can instead carry:

```sh
touch StRoot/StSomePackage/.skip_cons_build
```

Explicit `+...` cons includes still override the marker, matching the existing `SKIP_DIRS` behavior:

```sh
cons +StSomePackage
```

The marker is checked during package discovery for top-level packages, nested Pool/Generator-style
subpackages, and final package filtering.

### DIFF
--- a/mgr/Construct
+++ b/mgr/Construct
@@ -262,6 +262,30 @@ if ($STAR_SYS =~ /x86_darwin/) {
 #$items .= "StBTofPool|StFgtPool|StPeCMaker";
 # closing exclusion list
 my $items = "";# print "items $#items  = @items\n";
+# A directory can opt out of the cons build by dropping this file in it.
+my $skipConsBuildMarker = ".skip_cons_build";
+
+sub cons_skip_marker_overridden {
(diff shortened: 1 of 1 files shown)


iraklic (approved):

plexoos:
@jlauret, could you take a look when you have a chance?

plexoos:
@codex review

jlauret:
The change is fine. I would note that we could use name convention to skip compilation but it is true that there is a need to keep code but skip compilation (if no maintainer exists anymore and the Maker is not fundamental) so, all good.

jlauret (approved):
Left a comment that all is fine.

PR #806, opened 2026-05-21
StFwdTrackMaker log4cxx gating

jdbrice:


akioogawa:
This feels like an issue with STAR logger itself. It looks like it has settable limits based on string, yet it only disable printout and still keeps entire messages? Daniel said he tried. This maybe a simple and easy solution for now.

akioogawa (approved):

genevb:
This method gates _all_ LOG_INFO/WARN/DEBUG messages, such that there are no messages whatsover from this maker when running outside of debug mode. Is that desirable? I guess LOG_INFO from inherited classes still persists, so perhaps this is sufficient, but we need to be sure that's acceptable here, and in PRs #808 and #814 .

jdbrice:
@genevb - I will change the flag from bool->int to distinguish DEBUG / INFO / WARN - so that more important (WARN) are not hidden from production, standby for similar commits on the other two PRs

genevb:
Still `bool` for the message flag?

jdbrice:
@genevb - with [e8ca003](https://github.com/star-bnl/star-sw/pull/806/commits/e8ca0030b891c8473365ee2b3bcca2e2e56d2f01) we are ready to go. I still use a bool here (I dont want any ambiguity about what is silenced, hence why I decided not to go with a sliding int-based gate.)

I removed gating of LOG_WARNING and changed the only potential track level call of LOG_WARNING to DEBUG. we have other mechanisms (namely QA) to ensure that we do not have an inappropriately large number of failed track fits, so the warning message is not needed.

genevb (approved):

I approve with reservations.

plexoos (approved):
This is a reasonable short-term workaround for the confirmed Log4cxx memory-growth issue. However, it relies on locally overriding `LOG_INFO` and `LOG_DEBUG`, and it also downgrades the failed-track-fit message from `LOG_WARN` to `LOG_DEBUG`.  

For future reference, related changes are in #808 and #814. We should verify that STAR uses the latest Log4cxx version and revisit these workarounds when the underlying issue is addressed.

PR #807, opened 2026-05-21
StFwdTrackCollection ownership

jdbrice:
Fix a double-free in ~StFwdTrackCollection and add a Clear() method. StSPtrVec already deletes its owned pointers on destruction; the explicit delete-loop ran on top of that.

  - StFwdTrackCollection.cxx/.h: drop the manual loop; add Clear(Option_t*) that calls mTracks.clear() so the SPtrVec runs its delete-on-clear path.
  - StFwdTrackMaker.cxx::FillEvent(): call ftc->Clear() when the collection already exists, so a new event starts empty instead of accumulating across events.


akioogawa (approved):
LGTM

klendathu2k (approved):
Responsibility for clearing the track collection is now moved into a Clear method.  Clear is called as the first item of business in the track maker's FillEvent() method.  As implemented this should work.  However, please note the following.

This leaves StEvent's forward track collection in an awkward state between clearing out the current event and processing the next event.  In the unlikely event that an external maker tries to access the forward track collection _before_ StFwdTrackMaker runs, it could be presented with a non-empty track collection.  

It would be preferable to (1) create the track collection during StFwdTrackMaker::Make and (2) clear the collection during StFwdTrackMaker::Clear.

That said, let's move this along and address later.

plexoos (approved):
I share @klendathu2k’s concern that clearing the collection only at the start of FillEvent() can leave stale forward tracks visible between events. This is safe once StFwdTrackMaker runs, but an earlier maker could read data from the prior event. Let’s track moving collection creation to Make() and clearing to Clear() as a follow-up.

PR #808, opened 2026-05-21
StFwdHitLoader log4cxx gating

jdbrice:


akioogawa (approved):
same comment as #806

jdbrice:
@genevb - this is ready to go. Removed the gating of LOG_WARNING and doubled checked the INFO & DEBUG calls

genevb (approved):

I approve with reservations.

plexoos (approved):
The reported Log4cxx memory leak is concerning. Since it has been confirmed, we should verify that we are using the latest available version of Log4cxx.

PR #811, opened 2026-05-21
introduce fwd afterburner macro - re-run FwdTracking on MuDst and produce PicoDst

jdbrice:
Add fwd_afterburner macro, which is a standard way to (re)create PicoDsts from MuDst, allowing re-running of tracking and any Fwd reco chain.

Introduce fwd_afterburner.C: a MuDst-driven afterburner macro that re-runs FTT clustering / point-making, FCS clustering, the FwdTracker, and PicoDst writing on top of an existing MuDst. Includes per-subchain toggles useful for memory profiling and debugging.

  - New file StRoot/StFwdTrackMaker/macro/mudst/fwd_afterburner.C (~375 lines).
  - Optional nSkip arg to advance the input MuDst without running downstream makers.
  - Per-event RSS / memory-delta reporting tagged for easy grep to monitor/debug memory usage .


akioogawa:
Please take a look at 
  https://github.com/jdbrice/star-sw-1/blob/dev/StRoot/StFwdTrackMaker/macro/mudst/fwd_afterburner_db.C 
and merge if it looks good. Or add it along with this. That takes 1st event's time info to set DB date/time in case we need DB (runDb = true).

akioogawa (approved):

PR #812, opened 2026-05-21
FwdTracker leaks const

jdbrice:
Fix per-event leaks in ForwardTrackMaker and tighten const correctness on the hit-search interface so missing disks no longer silently insert empty entries into the data-source hit map.

  - FwdTracker.h::ForwardTrackMaker::Clear(): delete each entry in mFwdVertices before clearing the vector (was leaking on every event).
  - findFst/Ftt/EpdHitsNearProjectedState and addFst/Ftt/EpdHits: take HitMap_t / Seed_t inputs by const & and switch from map[key] to map.at(key) / find(). Note: the findFttStripsNearProjectedState default thresholdPhi is also adjusted (0.004*3 → 0.004*30) 
  - GenfitTrackResult.h::~GenfitTrackResult(): now actually runs Clear().
  - TrackFitter.h: small parallel tweaks for the same purpose.


akioogawa (approved):

genevb:
Are we expecting a review from @nigmatkulov as a code owner for StFwdTrackMaker? If not, @plexoos , anything else to hold this up?

plexoos:
> anything else to hold this up?

Nothing else from my side. Let me request a review from Copilot.

plexoos:
No comments from copilot — good.

One additional thought: since this PR is intended to fix a memory leak, it would be helpful to include a minimal reproducer showing the before-and-after behavior. I wouldn’t make this a blocker, but a note in the PR description indicating whether the issue can be demonstrated with an existing test or with a new test command that could be added to CI would be helpful.

plexoos (approved):

nigmatkulov (approved):
Looks good

PR #813, opened 2026-05-21
fwdtracker default verbosity = none for production

jdbrice:
uiet the fwd-tracker default verbosity and route stray cerr/cout through the STAR logger.
  
  - FwdTracker.h: flip ForwardTrackMaker::verbose default 1 → 0.
  - TrackFitter.h: flip kVerbose default 1 → 0.
  - GenfitTrackResult.h: drop the per-track LOG_INFO/LOG_DEBUG lines (mIdTruth=..., "Setting momentum from track", "set Track successful").
  - FitterUtils.h: collinear-points warnings in GenericFitSeeder switched from std::cerr/std::cout to LOG_DEBUG.


plexoos (approved):
LGTM

jdbrice:
@plexoos - most of the recent jobs are failing the same test due to a bad_alloc -> doesnt seem like all these changes could cause the same error. 

Is there a known bad commit or something that has that test failing?

plexoos:
> @plexoos - most of the recent jobs are failing the same test due to a bad_alloc -> doesnt seem like all these changes could cause the same error.
> 
> Is there a known bad commit or something that has that test failing?

Yes, I saw that too. There isn’t any known new commit that would explain this error. The test appears to be somewhat random and unstable; in some cases, rerunning it does not reproduce the failure. For example, that happened in https://github.com/star-bnl/star-sw/pull/815.

plexoos:
Yes, I just reran this specific test by itself, and that fixed it.

akioogawa (approved):

PR #814, opened 2026-05-21
StFttClusterMaker log4cxx gating

jdbrice:


akioogawa (approved):
same comment as #806

jdbrice:
@genevb this one is ready from my standpoint. I removed gating of LOG_WARNING and made more changes from INFO->DEBUG.
We've run with this in private productions and found it still manages to take care of memory issues by setting mDebug=false

genevb (approved):

I approve with reservations.

plexoos (approved):
This workaround addresses the immediate memory-growth issue, but we should verify that STAR is using the latest Log4cxx version and revisit these local macro overrides once the upstream issue is resolved.

Also see #808

PR #815, opened 2026-05-21
StFttClusterMaker correctness fixes

jdbrice:
Correctness fixes in StFttClusterMaker 
  - Make(): call mFttCollection->clusters().clear() at entry so clusters do not accumulate across events when the collection survives event boundaries.
  - Per-ROB hit lookup: introduce a getHits(map, key) helper backed by a static empty vector, so std::map::operator[] no longer silently inserts empty entries for missing ROB keys (parallel to the const-correctness theme in pr/fwdtracker-leaks-const).
  - FindMaxAdc / SearchClusterEdges: take input vectors by const & instead of by value (header + implementation).
  - CalculateClusterInfo: guard the m0Sum division with if (m0Sum > 0); assign -999 sentinels for x / sigma when there's no accumulated charge instead of producing NaN from divide-by-zero.
  - FindClusters: clusters.push_back(clu) is now called before CalculateClusterInfo(clu), so the cluster pointer is owned by the vector before any exception can throw inside the calculation.


akioogawa (approved):
LGTM

genevb (commented):
I was assigned as a reviewer, but I'm not familiar with this code. I have no reason to object to the changes. Is this ready to go through, @plexoos ?

plexoos:
@codex review

plexoos:
Some of the comments from Codex and Copilot may be worth taking a look at. @jdbrice?

PR #816, opened 2026-05-29
Do not rely on specific path for perl

genevb:
Use env to find a usable perl.

### DIFF
--- a/mgr/BFCOpt2Html.pl
+++ b/mgr/BFCOpt2Html.pl
@@ -1,4 +1,4 @@
-#!/usr/local/bin/perl -w
+#!/usr/bin/env -S perl -w
 
 #
 # Written J.Lauret sometime in January 2002
--- a/mgr/BFCOpt2Html.pl
+++ b/mgr/BFCOpt2Html.pl
@@ -1,4 +1,4 @@
-#!/usr/bin/env -S perl -w
+#!/usr/bin/env perl
 
 #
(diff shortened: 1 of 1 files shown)


akioogawa (approved):

iraklic:
Hi Gene. Does SL7 support -S? Because if it does not (and I thought it did not) will this brake some backwards compatibility?

genevb:
> Hi Gene. Does SL7 support -S? Because if it does not (and I thought it did not) will this brake some backwards compatibility?

Ugh. Looks like you're correct, @iraklic , that version 8.30 of `env` introduced `-S`, and we have version 8.22 on SL7 and 8.32 on A9. That leaves us with these options:

1. Use `-S` and say "bye-bye" to SL7 compatibility.
2. Drop both env's `-S` and perl's `-w` and forego the warnings from perl.
3. Just use `/usr/bin/perl -w`, which seems ever-present anyhow.

@jlauret ?

jlauret:
The use of -w is not required. This was added at first in case of issues (could also use strict or use warnings but at this stage, skip). Which means you can over-simplify and use

#!/usr/bin/env perl

and that's all.
And I would approve. This is minor.

genevb:
Fixed as Jerome approved by Jerome in his comment. @plexoos , can you please go ahead and approve and we can merge?

plexoos (approved):

PR #817, opened 2026-06-02
Alma9 compilation take 1

klendathu2k:
Changes to enable building agetof and stic under Alma 9.  

starsim compiles under Alma 9, but does not run due to ABI incompatibility with external dependencies.

### DIFF
--- a/asps/staf/sdd/Conscript
+++ b/asps/staf/sdd/Conscript
@@ -14,11 +14,11 @@ my $YACC = "yacc";
 my $YACCLIB = "-ly";
 my $LEX  = "lex";
 my $LEXLIB = "-ll";
-if ($env->{ENV}->{STAR_SYS} =~ /i386/) {
+if ($env->{ENV}->{STAR_SYS} =~ /i386/ or $env->{ENV}->{STAR_SYS} =~ /^al/) {
   $YACC = "/usr/bin/bison -y";
   $YACCLIB = "";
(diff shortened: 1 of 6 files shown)


asps/staf/sdd/src/idl.l
```diff
@@ -1,3 +1,4 @@
+%option noyywrap
```
klendathu2k, line 1: Does this break SL7 builds?

asps/staf/sdd/Conscript
```diff
@@ -14,11 +14,11 @@ my $YACC = "yacc";
 my $YACCLIB = "-ly";
 my $LEX  = "lex";
 my $LEXLIB = "-ll";
-if ($env->{ENV}->{STAR_SYS} =~ /i386/) {
+if ($env->{ENV}->{STAR_SYS} =~ /i386/ or $env->{ENV}->{STAR_SYS} =~ /^al/) {
   $YACC = "/usr/bin/bison -y";
   $YACCLIB = "";
   $LEX = "flex";
-  $LEXLIB = "-lfl";
+  $LEXLIB = "";
```
klendathu2k, line 21: Is this backwards compatible?

klendathu2k: ... CI tests succeeded... so strongly suggests that the changes are backwards compatible to SL7 ...

PR #820, opened 2026-06-09
Transfering setDbAccess to StFcsDb just in case user forget to set

akioogawa:
This is a minor change for StFcsDbMaker::setDbAccess to also set StFcsDb::setDbAccess for users (like me) forgetting to set the same for StFcsDb as well while debugging. This should not affect anything while using this with DB access on (default).

### DIFF
--- a/StRoot/StFcsDbMaker/StFcsDbMaker.cxx
+++ b/StRoot/StFcsDbMaker/StFcsDbMaker.cxx
@@ -117,7 +117,6 @@
  **************************************************************************/
 
 #include "StFcsDbMaker.h"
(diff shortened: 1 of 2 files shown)


dkapukchyan:
Good catch! Since you opened this minor fix and looking over this code it might be good to just change mDbAccess to a boolean. Also, we can change the mRun19 to a boolean too. This would mean changing a few if statement too. I would do it but I'm not sure if I can modify files in your pull request.

dkapukchyan (approved):

plexoos (approved):

PR #822, opened 2026-07-08
Small Fix to StFcsWaveformFitMaker and new experimental shower shape

dkapukchyan:
Fix to StFcsWaveformFitMaker to correctly pedestal subtract Highest3 method. New shower shape is still in testing so default is the same.

### DIFF
--- a/StRoot/StFcsPointMaker/StFcsPointMaker.cxx
+++ b/StRoot/StFcsPointMaker/StFcsPointMaker.cxx
@@ -70,7 +70,7 @@ void StFcsPointMaker::setShowerShapeParameters(int det){
 				 unused, unused, unused, unused, unused, unused, unused, unused, unused, unused,
 				 unused, unused, unused, unused, unused, unused, unused, unused, unused, unused};
     }else if(mShowerShape==1){ 
-	//Yuxi's 6 slices * 3 gaus                                                                                                                   
+	//Yuxi's 6 slices * 3 gaus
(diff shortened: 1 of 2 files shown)


akioogawa (approved):
Looks ok to me.

Trying new shower shape is good but need some motivation or theory or idea to backup for saying this may do better job. We have not yet really tried and tuned already existing shower shapes imported from FMS. I believe Yuxi's 6 z slice was used for late FMS analysis and it should be adjusted for FCS.

dkapukchyan:
> Looks ok to me.
> 
> Trying new shower shape is good but need some motivation or theory or idea to backup for saying this may do better job. We have not yet really tried and tuned already existing shower shapes imported from FMS. I believe Yuxi's 6 z slice was used for late FMS analysis and it should be adjusted for FCS.

That's right. This was a leftover from my studies on the FCS shower shapes from a few years ago that never really finished. I was playing with various parameters (hence the "new" shower shape) and found that our point reconstruction was too centered on the tower cell but I never did find a good set of values that have a more even distribution. That's why I called it "experimental" since I was using it for testing.

akioogawa:
I recently did Yuxi's 6 slide shower shape tuning for EPIC forward Ecal : 

https://www.star.bnl.gov/public/spin/akio/epic/photonFit/index.html

Yes indeed, fitting (and log E weighted mean as well) can pull position towards center or edge of the cell, depending mostly on shower shape transverse size. We need to do something like that for FCS.

PR #830, opened 2026-09-15
Geant4star sl26x branch

klendathu2k:
Attempt #2...


klendathu2k:
mistakenly targets main.  also likely to fail at the same spot / missing dependencies.

